PR #35, opened 2024-10-13
Migrate to exwlshelleventloop and iced 13

MalpenZibo:
Checklist before merging

- [x] fix general performance regression (higher CPU usage)
- [x] fix update scrollbar lag (maybe related to the same performance issue?)
- [x] fix slider change value reset issue (maybe related to the same performance issue?)
- [ ] find a way to handle multi monitor features (move the layer to an external attached monitor destroying the fallback layer on the main one)
- [ ] fix idle performance regression (compared to the current version or to pop_os branch 0.0/0.1% vs 0.3/0.4%)
- [ ] minor, blurry interface during application startup (first 5/6 seconds)


Decodetalkers:
Emm, seems the fourth feature is impossible.. the iced compositer is always binded with a wl_surface, if distory the surface, iced will panic.. Emmm

MalpenZibo:
Superseded by #42 

In the future, I may consider moving to this library to support the official iced version

PR #55, opened 2024-10-31
add ALT Linux install instruction

fiersik:
I would really like to help with the code, but so far I'm just learning RUST.

I'll start with this.


MalpenZibo:
Thanks you so much! Just one request, could you move the ALT Linux section under the arch one?

fiersik:
Of course, will that do?

PR #58, opened 2024-11-02
add Clipboard button

fiersik:
A simple button to open the clipboard may not seem very necessary at first, but it will be really useful for some users.

I also often use this button myself)

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-### Changed
+### Added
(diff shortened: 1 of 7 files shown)


MalpenZibo:
Very nice!! Just give me some time to test the feature in the next few days!!

fiersik:
> Just give me some time to test the feature in the next few days!!

Of course! I use it with the value `clipboardCmd: "cliphist-wofi-img | wl-copy"` and everything works fine!

MalpenZibo:
In the meantime if you want you can add a section in the changelog that describe this feature

fiersik:
It seems like it should look something like this

PR #63, opened 2024-11-05
Fix special workspaces

leftas:
Vector was constantly surprised by the massive request for capacity (in 44th/45th line of workspaces.rs). I chose to hide special workspaces so they wouldn't appear in the status bar.

In addition to this, I have added a backtrace. Right now, it simply just spit out the error message without any backtrace, whatsoever - it was hard to debug. In a perfect world, this should have been handled differently, but for now, it will be sufficient, I hope.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -10,9 +10,9 @@ use iced_sctk::{
     multi_window::{settings::Settings, Application},
     settings::InitialSurface,
 };
-use log::error;
-use std::{borrow::Cow, panic};
+use log::{error, LevelFilter};
 use std::backtrace::Backtrace;
+use std::{borrow::Cow, panic};
 
 mod app;
 mod centerbox;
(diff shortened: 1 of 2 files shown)


MalpenZibo:
Thank you so much for the PR :smile:  
In the end, I decided to handle the special workspace differently. I show the special workspaces using their names instead of the ids just before every other workspaces.
I made these changes in PR #64.

If you think that could be useful to completely hide any special workspace your changes could be integrated using a configuration flag (something like `hideSpecialWorkspaces`).

leftas:
Fair enough! I don't see a need for hiding if you want to show!

Thank you! At first, I thought I missed your PR, however, it seems that you created it after mine, haha.

MalpenZibo:
Yes, I notice your PR just after opening mine

leftas:
Now I am scared to add my keyboard's language module, as you already covered my two PRs in one shot, lol.

PR #65, opened 2024-11-05
Fix title panic

leftas:
Sadly, I can make this diff a little bit nicer (just merging into my previous branch, instead of the main one) as I cannot put this PR in your repo when targeting my other PR, so you will have to ignore that noise, haha. 

There was this special case when the substring was made in the middle of the symbol's codepoint(s) - rust just panics and returns an error: "byte index X is not a char boundary". As I understand that's the easiest way to solve this (numbering by codepoints and then taking the symbol's byte index). 
Pardon me, if that's not the right way - Rust is not my main language.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -10,9 +10,9 @@ use iced_sctk::{
     multi_window::{settings::Settings, Application},
     settings::InitialSurface,
 };
-use log::error;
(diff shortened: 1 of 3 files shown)


leftas:
Fixed in https://github.com/MalpenZibo/ashell/pull/64

PR #72, opened 2024-11-17
add keyboard-layout and keyboard-submap modules

fiersik:
## Description 
A simple module for displaying the current keyboard layout

image

## Functionality 
- Display of the current layout
- Switching the layout by clicking on the module 

The module is displayed only if more than one keyboard layout is installed, but if necessary, it can be forcibly hidden 
```
language:
    disabled: true
```

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -4,8 +4,9 @@ use crate::{
     get_log_spec,
     menu::{menu_wrapper, Menu, MenuPosition, MenuType},
     modules::{
-        self, clipboard, clock::Clock, launcher, privacy::PrivacyMessage, settings::Settings,
-        system_info::SystemInfo, title::Title, updates::Updates, workspaces::Workspaces,
+        self, clipboard, clock::Clock, keyboard_layout::KeyboardLayout, launcher,
+        privacy::PrivacyMessage, settings::Settings, system_info::SystemInfo, title::Title,
+        updates::Updates, workspaces::Workspaces,
     },
     services::{privacy::PrivacyService, ReadOnlyService, ServiceEvent},
     style::ashell_theme,
@@ -30,6 +31,7 @@ pub struct App {
     workspaces: Workspaces,
     window_title: Title,
     system_info: SystemInfo,
+    keyboard_layout: KeyboardLayout,
     clock: Clock,
     privacy: Option<PrivacyService>,
     pub settings: Settings,
@@ -46,6 +48,7 @@ pub enum Message {
     Workspaces(modules::workspaces::Message),
     Title(modules::title::Message),
     SystemInfo(modules::system_info::Message),
+    KeyboardLayout(modules::keyboard_layout::Message),
     Clock(modules::clock::Message),
     Privacy(modules::privacy::PrivacyMessage),
     Settings(modules::settings::Message),
@@ -67,6 +70,7 @@ impl Application for App {
                 workspaces: Workspaces::default(),
                 window_title: Title::default(),
                 system_info: SystemInfo::default(),
+                keyboard_layout: KeyboardLayout::default(),
                 clock: Clock::default(),
                 privacy: None,
                 settings: Settings::default(),
@@ -136,6 +140,10 @@ impl Application for App {
             }
             Message::SystemInfo(message) => {
                 self.system_info.update(message);
+                Command::none()
+            }
+            Message::KeyboardLayout(message) => {
+                self.keyboard_layout.update(message);
                 Command::none()
             }
             Message::Clock(message) => {
@@ -227,6 +235,13 @@ impl Application for App {
                         .map(|c| c.map(Message::SystemInfo)),
                 )
                 .push(
+                    Row::new().push_maybe(
+                        self.keyboard_layout
+                            .view(&self.config.keyboard.layout)
+                            .map(|l| l.map(Message::KeyboardLayout)),
+                    ),
+                )
+                .push(
                     Row::new()
                         .push(
                             self.clock
@@ -264,6 +279,11 @@ impl Application for App {
                 Some(self.workspaces.subscription().map(Message::Workspaces)),
                 Some(self.window_title.subscription().map(Message::Title)),
                 Some(self.system_info.subscription().map(Message::SystemInfo)),
+                Some(
+                    self.keyboard_layout
+                        .subscription()
+                        .map(Message::KeyboardLayout),
+                ),
                 Some(self.clock.subscription().map(Message::Clock)),
                 Some(
                     PrivacyService::subscribe().map(|e| Message::Privacy(PrivacyMessage::Event(e))),
--- a/src/app.rs
+++ b/src/app.rs
@@ -4,9 +4,9 @@ use crate::{
     get_log_spec,
     menu::{menu_wrapper, Menu, MenuPosition, MenuType},
     modules::{
-        self, clipboard, clock::Clock, keyboard_layout::KeyboardLayout, launcher,
-        privacy::PrivacyMessage, settings::Settings, system_info::SystemInfo, title::Title,
-        updates::Updates, workspaces::Workspaces,
+        self, clipboard, clock::Clock, keyboard_layout::KeyboardLayout,
+        keyboard_submap::KeyboardSubmap, launcher, privacy::PrivacyMessage, settings::Settings,
+        system_info::SystemInfo, title::Title, updates::Updates, workspaces::Workspaces,
     },
     services::{privacy::PrivacyService, ReadOnlyService, ServiceEvent},
     style::ashell_theme,
@@ -32,6 +32,7 @@ pub struct App {
     window_title: Title,
     system_info: SystemInfo,
     keyboard_layout: KeyboardLayout,
+    keyboard_submap: KeyboardSubmap,
     clock: Clock,
     privacy: Option<PrivacyService>,
     pub settings: Settings,
@@ -49,6 +50,7 @@ pub enum Message {
     Title(modules::title::Message),
     SystemInfo(modules::system_info::Message),
     KeyboardLayout(modules::keyboard_layout::Message),
+    KeyboardSubmap(modules::keyboard_submap::Message),
     Clock(modules::clock::Message),
     Privacy(modules::privacy::PrivacyMessage),
     Settings(modules::settings::Message),
@@ -71,6 +73,7 @@ impl Application for App {
                 window_title: Title::default(),
                 system_info: SystemInfo::default(),
                 keyboard_layout: KeyboardLayout::default(),
+                keyboard_submap: KeyboardSubmap::default(),
                 clock: Clock::default(),
                 privacy: None,
                 settings: Settings::default(),
@@ -144,6 +147,10 @@ impl Application for App {
             }
             Message::KeyboardLayout(message) => {
                 self.keyboard_layout.update(message);
+                Command::none()
+            }
+            Message::KeyboardSubmap(message) => {
+                self.keyboard_submap.update(message);
                 Command::none()
             }
             Message::Clock(message) => {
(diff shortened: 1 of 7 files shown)


fiersik:
Checking the code is my main enemy (

fiersik:
Hooray, I passed the tests

MalpenZibo:
Very nice! Thanks for submitting new features :) Recently I don't have too much time but I will try to review you code as soon as I can.

fiersik:
> Thanks for submitting new features :)

It's more fun together!

> but I will try to review you code as soon as I can.

okay, I'll think about how to improve this module for now.) 
Alas, this PR conflicts with the tray branch

fiersik:
I have another idea!

fiersik:
ouch...

src/app.rs
```diff
@@ -30,6 +31,7 @@ pub struct App {
     workspaces: Workspaces,
     window_title: Title,
     system_info: SystemInfo,
+    language: Language,
```
MalpenZibo: here I prefer something like `KeyboardLayout` for the name of the module and the struct

src/modules/language.rs
```diff
@@ -0,0 +1,110 @@
...
+
+fn get_active_keymap() -> String {
+    let keyboards = hyprland::data::Devices::get().unwrap().keyboards;
+    let main_keyboard = keyboards.iter().find(|k| k.main);
+    main_keyboard.unwrap().active_keymap.to_string()
+}
+
+pub struct Language {
+    language: String,
+}
```
MalpenZibo:
```suggestion
pub struct KeyboardLayout {
    layouts: Vec<(String, bool)>, // the bool indicate if it's active otherwise we could create a new struct { layout: String, active: bool }
}
```

src/modules/language.rs
```diff
@@ -0,0 +1,110 @@
...
+        Self {
+            language: get_active_keymap(),
+        }
+    }
+}
+
+#[derive(Debug, Clone)]
+pub enum Message {
+    LanguageChanged(String),
+    ChangeLanguage(),
```
MalpenZibo:
```suggestion
    ChangeLanguage,
```

src/modules/language.rs
```diff
@@ -0,0 +1,110 @@
...
+            }
+        }
+    }
+
+    pub fn view(&self, config: &LanguageModuleConfig) -> Option<Element<Message>> {
+        let layouts = hyprland::keyword::Keyword::get("input:kb_layout")
+            .unwrap()
+            .value
+            .to_string();
+        let layout_list: Vec<&str> = layouts.split(',').collect();
```
MalpenZibo: For performance reason I prefer to move this logic to the subscription, maybe we could change the `get_active_keymap` to return a `Vec<Layout>` where Layout is something like
```
struct Layout {
  layout: String,
  active: bool
}
```

src/modules/language.rs
```diff
@@ -0,0 +1,110 @@
...
+                button(text(self.language.clone()))
+                    .padding([2, 7])
+                    .on_press(Message::ChangeLanguage())
+                    .style(Button::custom(HeaderButtonStyle::Full))
+                    .into(),
+            )
+        }
+    }
+
+    pub fn subscription(&self) -> Subscription<Message> {
```
MalpenZibo: I will move all the possible logic here. The `LanguageChanged` message should contain the list of all possible kb_layouts and the active one.

src/modules/language.rs
```diff
@@ -0,0 +1,110 @@
...
+    Element, Subscription,
+};
+use log::{debug, error};
+use std::{
+    any::TypeId,
+    sync::{Arc, RwLock},
+};
+
+fn get_active_keymap() -> String {
+    let keyboards = hyprland::data::Devices::get().unwrap().keyboards;
```
MalpenZibo: Let's try to avoid `unwrap.` We should choose between returning a default value and an Option. In this case, I cannot think of a sane default value, so maybe it's better to return an Option.

MalpenZibo (changes requested):
Very nice! I left some comments about the name of the module, I prefer something that better indicates that this is a switch to change the Keyboard layout and not the system language, and some performance stuff

fiersik:
Thank you very much! I will try to take into account all the suggestions.

fiersik:
I didn't use all the suggestions, however, I definitely improved the code.

src/app.rs
```diff
@@ -226,6 +234,13 @@ impl Application for App {
                         .view(&self.config.system)
                         .map(|c| c.map(Message::SystemInfo)),
                 )
+                .push(
+                    Row::new().push_maybe(
+                        self.keyboard_layout
+                            .view(&self.config.keyboard.layout)
+                            .map(|l| l.map(Message::KeyboardLayout)),
+                    ),
+                )
```
```suggestion
                .push_maybe(
                        self.keyboard_layout
                            .view(&self.config.keyboard.layout)
                            .map(|l| l.map(Message::KeyboardLayout)),
                )
```

src/config.rs
```diff
@@ -78,6 +78,32 @@ impl Default for SystemModuleConfig {
...
+}
+
+impl Default for KeyboardModuleConfig {
+    fn default() -> Self {
+        Self {
+            layout: default_keyboard_layout(),
+        }
+    }
+}
+
```
MalpenZibo: Why have you created a nested struct? Something like
```
#[derive(Deserialize, Clone, Default, Debug)]
#[serde(rename_all = "camelCase")]
pub struct KeyboardModuleConfig {
    #[serde(default)]
    pub disabled: bool,
}
```
is not enough?

src/modules/keyboard_layout.rs
```diff
@@ -0,0 +1,130 @@
...
+
+    pub fn subscription(&self) -> Subscription<Message> {
+        let id = TypeId::of::<Self>();
+
+        channel(id, 10, |output| async move {
+            let output = Arc::new(RwLock::new(output));
+            loop {
+                let mut event_listener = AsyncEventListener::new();
+
+                event_listener.add_layout_changed_handler({
```
MalpenZibo, line 100: Here we should also listen to hyprland config changes
```
                event_listener.add_config_reloaded_handler({
                    let output = output.clone();
                    move || {
                        let output = output.clone();
                        Box::pin(async move {
                            if let Ok(mut output) = output.write() {
                                output
                                    .try_send(Message::LayoutConfigChanged)
                                    .expect("error sending message: layout config changed event");
                            }
                        })
                    }
                });

```

src/modules/keyboard_layout.rs
```diff
@@ -0,0 +1,130 @@
...
+            }
+        }
+    }
+
+    pub fn view(&self, config: &KeyboardLayoutModule) -> Option<Element<Message>> {
+        if config.disabled || self.layout.num_layouts < 2 {
+            None
+        } else {
+            Some(
+                button(text(self.layout.active.clone()))
```
MalpenZibo:
```suggestion
                button(text(&self.layout.active))
```

src/modules/keyboard_layout.rs
```diff
@@ -0,0 +1,130 @@
...
+            let layout_list: Vec<&str> = layout_str.split(',').collect();
+            layout_list.len()
+        }
+        Err(_) => 0,
+    };
+
+    Layout {
+        num_layouts,
+        active,
+    }
```
MalpenZibo: Now I understand better the problem.

In the end, maybe we should split this function into two: one to get the active keyboard layout and one to get a flag indicating if there are more than two possible keyboard layouts. 

The first one should be used during module initialization and when we receive a hyprland config changes event, and the second during the module init and when we receive a hyprland layout changed event.

Here's an example
```
fn get_multiple_layout_flag() -> bool {
    match hyprland::keyword::Keyword::get("input:kb_layout") {
        Ok(layouts) => layouts.value.to_string().split(",").count() > 1, //In this way we skip the collect to avoid useless string allocation
        Err(_) => false,
    }
}

fn get_active_layout() -> String {
    hyprland::data::Devices::get()
        .ok()
        .and_then(|devices| {
            devices
                .keyboards
                .iter()
                .find(|k| k.main)
                .map(|keyboard| keyboard.active_keymap.to_string())
        })
        .unwrap_or_else(|| "unknown".to_string())
}

```

src/modules/keyboard_layout.rs
```diff
@@ -0,0 +1,130 @@
...
+    layout: Layout,
+}
+
+impl Default for KeyboardLayout {
+    fn default() -> Self {
+        Self {
+            layout: get_layout(),
+        }
+    }
+}
```
MalpenZibo: This could be simplified into (feel free to change names)
```suggestion
#[derive(Debug, Clone)]
pub struct KeyboardLayout {
    multiple_layout: bool,
    active: String,
}

impl Default for KeyboardLayout {
    fn default() -> Self {
        Self {
            multiple_layout: get_multiple_layout_flag(),
            active: get_active_layout(),
        }
    }
}
```

src/modules/keyboard_layout.rs
```diff
@@ -0,0 +1,130 @@
...
+
+#[derive(Debug, Clone)]
+pub enum Message {
+    LanguageChanged(Layout),
+    ChangeLanguage,
+}
```
MalpenZibo:
```suggestion
pub enum Message {
    LayoutConfigChanged,
    ActiveLayoutChanged(String),
    ChangeLayout,
}
```

src/modules/keyboard_layout.rs
```diff
@@ -0,0 +1,130 @@
...
+
+                if let Err(e) = res {
+                    error!("failed to keymap change: {:?}", e);
+                }
+            }
+        }
+    }
+
+    pub fn view(&self, config: &KeyboardLayoutModule) -> Option<Element<Message>> {
+        if config.disabled || self.layout.num_layouts < 2 {
```
MalpenZibo: with the other change this could become
```suggestion
  if config.disabled || !self.multiple_layout 
```

MalpenZibo (changes requested):
Nice! we're almost there :rocket:

fiersik:
regarding the first few amendments...

The nested config is made specifically and will be used for another module `keyboard_submap`
The nested Row::new() is made for the same purpose

They should not be cleaned

MalpenZibo:
> regarding the first few amendments...
> 
> The nested config is made specifically and will be used for another module `keyboard_submap` The nested Row::new() is made for the same purpose
> 
> They should not be cleaned

Ah okok, so you plan to add additional features in this PR? Ok!

src/modules/keyboard_layout.rs
fiersik: > and when we receive a hyprland config changes event,

When changing the `input` configuration, the keyboard layout update event comes, so it doesn't make much sense.

Both functions will still be called.

fiersik:
> so you plan to add additional features in this PR?

Yes! But here the task is more interesting, we need to output data that cannot be requested in the usual way, we will have to take it directly from the event

fiersik:
I corrected a few sentences, but left the code that may be useful to me in the future.

fiersik:
Great, I'm waiting for the latest checks and it will be possible to update the README and changelog

Now it looks like this:
image
Modules are displayed only if they are needed, hidden if necessary

MalpenZibo:
@fiersik ok, if you have finished with the PR make it ready to review and I will do the final review and test

fiersik:
I think I've done everything I wanted to do.

src/modules/keyboard_submap.rs
```diff
@@ -0,0 +1,84 @@
...
+
+    pub fn subscription(&self) -> Subscription<Message> {
+        let id = TypeId::of::<Self>();
+
+        channel(id, 10, |output| async move {
+            let output = Arc::new(RwLock::new(output));
+            loop {
+                let mut event_listener = AsyncEventListener::new();
+
+                event_listener.add_sub_map_changed_handler({
```
MalpenZibo, line 61: I don't know if we want to listen to Hyland config changes here. (Maybe it's too much; I don't really know.)

MalpenZibo (approved):
:rocket: Thanks for the effort.

Just a minor comment but it's more than ok as it is!

fiersik:
submap is configuration independent, so I don't see the point in tracking configuration changes.

fiersik:
well...Are there any amendments?
I think everything is ready here

fiersik:
Thanks :)

PR #80, opened 2024-12-30
Add module position configuration

ReshetnikovPavel:
Added a way to configure which modules should be on the left, center or right
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,7 +121,16 @@ outputs: # optional, default empty list (the bar will be displayed on the active
   - DP-1
 # Bar position, possible values Top | Bottom.
 position: Top # optional, default Top
-# App lancher commanda, it will be used to open the launcher,
+# Lists of modules on left, center and right
+# possible values: launcher | clipboard | updates | workspaces | title | systemInfo | keyboardSubmap | keyboardLayout | clock | privacy | settings
+left: # optional, this list is default
+    - workspaces
+center: # optional, this list is default
+    - title
+right: # optional, this list is default
+    - clock
+    - settings
+# App lancher command, it will be used to open the launcher,
 # without a value the related button will not appear
 appLauncherCmd: "~/.config/rofi/launcher.sh" # optional, default None
 # Clipboard command, it will be used to open the clipboard menu,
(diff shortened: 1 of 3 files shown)


fiersik:
That's a great idea! However, the screenshot you provided shows the main problem with this implementation.

image

image

(sorry for my bad English)

fiersik:
also, changing the position of the module will not change the position of the menu that is opened by the module, for example, Updates

MalpenZibo:
I thought about this kind of configuration recently.

I would like to have a configuration field like `modules` that's something like

``` rust 
enum ModuleName {
  Workspace,
  Title,
  Settings,
  etc ...
}

enum ModuleDef {
  Single(ModuleName),
  Group(Vec<ModuleName>)
}

struct Modules {
  left: Vec<ModuleDef>,
  center: Vec<ModuleDef>,
  right: Vec<ModuleDef>
}

...
modules: Modules,
...
``` 

A Single ModuleDef should have the `HeaderButtonStyle::Full` or the `header_pills` style (I should probably refactor these styles), and a Group ModuleDef should have:
-  the `HeaderButtonStyle::Full` or the `header_pills` if there's only one element in the vector
-  the `HeaderButtonStyle::Left` or the `header_pills_left` if it's the first element in the vector
-  the `HeaderButtonStyle::Right` or the `header_pills_right` if it's the last element in the vector
- what I currently use for the privacy module otherwise (it could be called HeaderButtonStyle::None or header_pills_none I don't know yet)

Also with this new configuration, we could remove the `enabled` flag in each module config structure and a module is enabled only if it's present inside this new `modules` field.

Regarding the menu position, I'm currently working on a solution that decides the position of a menu based on the position of the relative button. This solution should be implemented to make the tray module work correctly, so I plan to use this new system for the other modules as well.

It's not a trivial change. So If you want, you could begin working, for example, on the configuration stuff then as soon as I can I could make the other changes to complete this feature

MalpenZibo:
@ReshetnikovPavel thanks for the input! I moved the implementation in this PR https://github.com/MalpenZibo/ashell/pull/84

PR #94, opened 2025-01-28
Added MediaPlayer module

### LINKED ISSUE #93: Music player module
Love the work being done here. If its within scope, I would like a module to control music players. I've made a fork with one specifically for mpd, but it could make use of `playerctl` instead for better integration with more than just mpd. I could submit a PR for it.

mazei513:
Closes #93. A configuration feature that can be added is the metadata format which currently is hardcoded to `"{{ artist }} - {{ title }}"` and a `config.playerctl.max_length`, similar to `truncate_title_after_length` for window titles, some titles get very long.


mazei513:
Have tested this with mpd, Youtube and Spotify web.

MalpenZibo:
Very nice, I will review and test everything as soon as possible.
In the future, I would like to integrate directly with the MPRIS D-Bus interface but for now, it's ok to use playerctl.

mazei513:
If integrating directly with MPRIS D-Bus, would relying on a crate like [mpris](https://crates.io/crates/mpris) be preferred or are you looking to write your own implementation?

mazei513:
Just realized there's already other D-Bus integrations, will reference those in the future.

MalpenZibo:
I think that the best approach is to complete this version and then proceed with a refactor of the service adding a dbus integration

src/modules/playerctl.rs
```diff
@@ -0,0 +1,163 @@
...
+pub struct Playerctl {
+    song: String,
+    volume: f64,
+}
+
+impl Default for Playerctl {
+    fn default() -> Self {
+        Self {
+            song: String::default(),
+            volume: get_volume(),
```
MalpenZibo: This module is not a default one, so because right now I don't want to add `playerctl` as a "hard" dependency we should take into consideration that a user could not have it installed.
For this reason even if in the config this module is not present this default call tries to get thee volume causing an error. 

Probably it's better to set this value to `None` (both the song and the volume)

src/modules/playerctl.rs
```diff
@@ -0,0 +1,163 @@
...
+        .arg("playerctl volume")
+        .stdout(Stdio::piped())
+        .output();
+
+    match get_current_song_cmd {
+        Ok(check_update_cmd) => {
+            String::from_utf8_lossy(&check_update_cmd.stdout)
+                .trim()
+                .parse::<f64>()
+                .unwrap()
```
MalpenZibo: This unwrap causes problems if it's not possible to get the volume.
Right now I'm testing playing a video on youtube using Zen browser and I get `No player could handle this command`.

I'm not an expert but I think that the `prev`, `next`, and `get/set volume` could fail. Right now the play/pause seems to work fine in every circumstance.

Probably we could use a default or maybe an Option to disable the feature

src/modules/playerctl.rs
```diff
@@ -0,0 +1,163 @@
...
+async fn get_current_song() -> String {
+    let get_current_song_cmd = process::Command::new("bash")
+        .arg("-c")
+        .arg("playerctl metadata --format \"{{ artist }} - {{ title }}\"")
+        .stdout(Stdio::piped())
+        .output()
+        .await;
+
+    match get_current_song_cmd {
+        Ok(get_current_song_cmd) => String::from_utf8_lossy(&get_current_song_cmd.stdout)
```
MalpenZibo: We should handle situations in which no songs are played. In this case, we shouldn't display the module. Right now there's an empty string.

We could set the song to `None` and in that case doesn't show the module

src/modules/playerctl.rs
```diff
@@ -0,0 +1,163 @@
...
+
+    pub fn menu_view(&self) -> Element<Message> {
+        column![
+            slider(0.0..=100.0, self.volume, |new_v| {
+                Message::SetVolume(new_v)
+            }),
+            row![
+                button(icon(Icons::SkipPrevious)).on_press(Message::Prev),
+                button(icon(Icons::PlayPause)).on_press(Message::Play),
+                button(icon(Icons::SkipNext)).on_press(Message::Next)
```
MalpenZibo:
```suggestion
                button(icon(Icons::SkipPrevious))
                    .on_press(Message::Prev)
                    .padding([5, 12])
                    .style(SettingsButtonStyle.into_style()),
                button(icon(Icons::PlayPause))
                    .on_press(Message::Play)
                    .style(SettingsButtonStyle.into_style()),
                button(icon(Icons::SkipNext))
                    .on_press(Message::Next)
                    .padding([5, 12])
                    .style(SettingsButtonStyle.into_style())
```

May I suggest using this style to align with the other button in the settings menu?

src/app.rs
```diff
@@ -266,6 +270,13 @@ impl App {
                     *button_ui_ref,
                     self.config.position,
                 ),
+                Some((MenuType::Playerctl, button_ui_ref)) => menu_wrapper(
+                    id,
+                    self.playerctl.menu_view().map(Message::Playerctl),
+                    MenuSize::Large,
```
```suggestion
                    MenuSize::Normal,
```

src/config.rs
```diff
@@ -283,6 +283,7 @@ pub enum ModuleName {
     Clock,
     Privacy,
     Settings,
+    Playerctl,
```
MalpenZibo: Could we use a name not strictly related to the underlying tool that we use to get the data? Something like `Media` or something similar?

src/modules/playerctl.rs
```diff
@@ -0,0 +1,163 @@
...
+    fn subscription(&self, (): Self::SubscriptionData<'_>) -> Option<Subscription<app::Message>> {
+        let id = TypeId::of::<Self>();
+
+        Some(Subscription::batch(vec![Subscription::run_with_id(
+            id,
+            channel(10, |mut output| async move {
+                loop {
+                    let song = get_current_song().await;
+                    let _ = output.try_send(Message::SetSong(song));
+                    sleep(Duration::from_secs(1)).await;
```
MalpenZibo: here should be better to avoid polling but without a dbus integration I think that's no other way so right now it ok!

src/modules/playerctl.rs
```diff
@@ -0,0 +1,163 @@
...
+
+impl Module for Playerctl {
+    type ViewData<'a> = ();
+    type SubscriptionData<'a> = ();
+
+    fn view(
+        &self,
+        (): Self::ViewData<'_>,
+    ) -> Option<(Element<app::Message>, Option<OnModulePress>)> {
+        Some((
```
MalpenZibo: Here we should return Some only if we have a song title

MalpenZibo:
> If integrating directly with MPRIS D-Bus, would relying on a crate like [mpris](https://crates.io/crates/mpris) be preferred or are you looking to write your own implementation?

I'm open to use external crate but I want to reduce at minimum dependency on small crate or things that could become unmaintained. I have to understand the effort to write this stuff from scratch (for the tray service at the end I wrote my implementation) but generally speaking I prefer to write something by my own if it's not too complex.

mazei513:
> > If integrating directly with MPRIS D-Bus, would relying on a crate like [mpris](https://crates.io/crates/mpris) be preferred or are you looking to write your own implementation?
> 
> I'm open to use external crate but I want to reduce at minimum dependency on small crate or things that could become unmaintained. I have to understand the effort to write this stuff from scratch (for the tray service at the end I wrote my implementation) but generally speaking I prefer to write something by my own if it's not too complex.

Totally understand that. Looking at the services under `src/services`, I can figure out what's going on generally and I think something similar could be done for for this.

> I think that the best approach is to complete this version and then proceed with a refactor of the service adding a dbus integration

Agreed on that. I tried to write up an MPRIS service with reference to the other services, but am facing issues with borrows and lifetimes. Quite new to Rust so having to figure out the async/await and `Stream` stuff on top of that has been a little confusing but I'll continue figuring that out on the side. For now I'll work on your comments to get this merged in.

src/modules/playerctl.rs
mazei513: Is 1 second maybe too quick? Though with future plans, maybe discussing this is not useful and better to focus on D-Bus integration.

src/config.rs
mazei513: Would `MediaPlayer` make the most sense? Would match the naming scheme in `MPRIS` as well, keeping it consistent.

mazei513:
Have renamed everything to `MediaPlayer`/`media_player`.

Have changed the initialization of the `MediaPlayer` module. The volume fetching is now also a part of the `MediaPlayer` subscriptions just like the song fetching. This also should fix any sync issues if the volume was changed outside of the module. Both song and volume fetching should handle errors better and will log the error and return `None` if there are issues.

Fixed the button and menu styling.

mazei513:
> As you suggest could be useful to have a setting to specify the max title length, it's ok to leave the format hardcoded for now.

Should the max title length config be included in this PR or done separately?

src/modules/media_player.rs
```diff
@@ -0,0 +1,190 @@
...
+        .arg("playerctl metadata --format \"{{ artist }} - {{ title }}\"")
+        .stdout(Stdio::piped())
+        .output()
+        .await;
+
+    match get_current_song_cmd {
+        Ok(get_current_song_cmd) => {
+            if !get_current_song_cmd.status.success() {
+                return None;
+            }
```
mazei513, line 33: Should this log the `stderr`? `playerctl` seems to output to `stderr` in non-zero exit cases.

mazei513: The same question goes for `get_volume_cmd` later on.

src/modules/playerctl.rs
MalpenZibo: Yes, leve it as It Is. We could merge this PR and then work on the dbus integration

MalpenZibo:
> > As you suggest could be useful to have a setting to specify the max title length, it's ok to leave the format hardcoded for now.
> 
> Should the max title length config be included in this PR or done separately?

If it comes easy I would like the feature in this PR.

mazei513:
Have added the config.

Also added a SyncVolume message for use with the volume subscription. The issue I was facing was that if the current player that playerctl would control doesn't allow setting volume, it would constantly log that error. SyncVolume is for simply syncing the internal state with what playerctl reports, while the slider will still use SetVolume which will call playerctl to set the volume.

src/modules/media_player.rs
```diff
@@ -0,0 +1,214 @@
...
+
+    fn view(
+        &self,
+        (): Self::ViewData<'_>,
+    ) -> Option<(Element<app::Message>, Option<OnModulePress>)> {
+        self.song.clone().and_then(|s| {
+            Some((
+                text(s).size(12).into(),
+                Some(OnModulePress::ToggleMenu(MenuType::MediaPlayer)),
+            ))
```
MalpenZibo:
```suggestion
        self.song.clone().map(|s| {
            (
                text(s).size(12).into(),
                Some(OnModulePress::ToggleMenu(MenuType::MediaPlayer)),
            )
```

src/modules/media_player.rs
MalpenZibo: No, it's ok as it is. No need to add a log.

MalpenZibo (approved):
Well just a minor thing and then it's ok for me

MalpenZibo:
I almost forgot. Could you update the README  with reference to this module and with the instructions and the CHANGELOG to mention this change in the Unreleased section?

mazei513:
Fixed the `and_then` to `map` and added the info to the README and CHANGELOG.

MalpenZibo:
Perfect!

PR #98, opened 2025-02-02
MPRIS DBus integration & multiple media players

mazei513:
Directly integrate with MPRIS DBus interface instead of using `playerctl`. With this, CPU time is significantly lower (simple check through `top`).

This PR also implements support for displaying more than 1 player at a time in the Media Player module menu.

### DIFF
--- a/src/modules/media_player.rs
+++ b/src/modules/media_player.rs
@@ -7,7 +7,7 @@ use crate::{
     config::MediaPlayerModuleConfig,
     menu::MenuType,
     services::{
-        mpris::{MprisPlayerCommand, MprisPlayerData, MprisPlayerService, PlayerCommand},
+        mpris::{MprisPlayerCommand, MprisPlayerService, PlayerCommand},
         ReadOnlyService, Service, ServiceEvent,
     },
     style::SettingsButtonStyle,
@@ -21,14 +21,7 @@ use iced::{
 
 #[derive(Default)]
 pub struct MediaPlayer {
-    data: Vec<PlayerData>,
     service: Option<MprisPlayerService>,
-}
-
-struct PlayerData {
-    name: String,
-    song: Option<String>,
-    volume: Option<f64>,
 }
 
 #[derive(Debug, Clone)]
@@ -41,11 +34,7 @@ pub enum Message {
 }
 
 impl MediaPlayer {
-    pub fn update(
-        &mut self,
-        message: Message,
-        config: &MediaPlayerModuleConfig,
-    ) -> Task<crate::app::Message> {
+    pub fn update(&mut self, message: Message) -> Task<crate::app::Message> {
         match message {
             Message::Prev(s) => self.handle_command(s, PlayerCommand::Prev),
             Message::PlayPause(s) => self.handle_command(s, PlayerCommand::PlayPause),
@@ -53,13 +42,10 @@ impl MediaPlayer {
             Message::SetVolume(s, v) => self.handle_command(s, PlayerCommand::Volume(v)),
             Message::Event(event) => match event {
                 ServiceEvent::Init(s) => {
-                    self.data = Self::map_service_to_module_data(s.deref(), config);
                     self.service = Some(s);
                     Task::none()
                 }
                 ServiceEvent::Update(d) => {
-                    self.data = Self::map_service_to_module_data(&d, config);
-
                     if let Some(service) = self.service.as_mut() {
                         service.update(d);
                     }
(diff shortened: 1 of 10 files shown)


src/modules/media_player.rs
```diff
@@ -1,94 +1,43 @@
...
-        Err(e) => {
-            error!("Error: {:?}", e);
-            None
-        }
-    }
-}
 
 #[derive(Default)]
 pub struct MediaPlayer {
+    data: Vec<PlayerData>,
```
MalpenZibo: no need to duplicate the data. In theory, you could use the one directly inside the `MprisPlayerService`. But maybe it's easier for you in this way. I notice that the `PlayerData` and the `MprisPlayerData` have different structure

src/modules/media_player.rs
```diff
@@ -184,31 +156,25 @@ impl Module for MediaPlayer {
...
-                text(s).size(12).into(),
+        match self.data.len() {
+            0 => None,
+            1 => self.data[0].song.clone().map(|s| {
+                (
+                    text(s).into(),
+                    Some(OnModulePress::ToggleMenu(MenuType::MediaPlayer)),
+                )
+            }),
+            _ => Some((
```
MalpenZibo, line 140: Here maybe it's better to show only the first one instead of only the MusicNote?

src/modules/media_player.rs
```diff
@@ -184,31 +156,25 @@ impl Module for MediaPlayer {
         &self,
         (): Self::ViewData<'_>,
     ) -> Option<(Element<app::Message>, Option<OnModulePress>)> {
-        self.song.clone().map(|s| {
-            (
-                text(s).size(12).into(),
+        match self.data.len() {
+            0 => None,
+            1 => self.data[0].song.clone().map(|s| {
```
MalpenZibo: Maybe could be better to always show the `MusicNote` as a prefix to better distinguish from the windows title?

src/modules/media_player.rs
```diff
@@ -98,81 +47,104 @@ impl MediaPlayer {
...
+                            .padding([5, 12])
+                            .style(SettingsButtonStyle.into_style())
+                    ]
+                    .spacing(8);
+
+                    [
+                        iced::widget::horizontal_rule(2).into(),
+                        container(
+                            column![]
+                                .push_maybe(d.song.clone().map(text))
```
MalpenZibo: Would be nice to have a fallback value where we cannot find the song title. Something like `NoTitle` or boh I don't know some common default value in this case :sweat_smile:

src/services/mpris/mod.rs
```diff
@@ -0,0 +1,340 @@
...
+    async fn initialize_data(
+        conn: &zbus::Connection,
+    ) -> anyhow::Result<(Vec<String>, Vec<MprisPlayerData>)> {
+        let dbus = DBusProxy::new(conn).await?;
+        let names: Vec<String> = dbus
+            .list_names()
+            .await?
+            .iter()
+            .filter(|&a| a.starts_with(MPRIS_PLAYER_SERVICE_PREFIX))
+            .map(|a| a.to_string())
```
MalpenZibo:
```suggestion
            .filter_map(|a| {
                if a.starts_with(MPRIS_PLAYER_SERVICE_PREFIX) {
                    Some(a.to_string())
                } else {
                    None
                }
            })

```

src/services/mpris/mod.rs
```diff
@@ -0,0 +1,340 @@
...
+                .await?
+                .filter_map(|s| {
+                    iced::futures::future::ready(match s.args() {
+                        Ok(a) => a
+                            .name
+                            .starts_with(MPRIS_PLAYER_SERVICE_PREFIX)
+                            .then_some(Event::NameOwner),
+                        Err(_) => None,
+                    })
+                })
```
MalpenZibo:
```suggestion
                .filter_map(|s| async move {
                    match s.args() {
                        Ok(a) => a
                            .name
                            .starts_with(MPRIS_PLAYER_SERVICE_PREFIX)
                            .then_some(Event::NameOwner),
                        Err(_) => None,
                    }
                })
```

src/services/mpris/mod.rs
```diff
@@ -0,0 +1,340 @@
...
+                }
+            },
+            State::Active(conn, names) => match Self::events(&conn).await {
+                Ok(mut events) => {
+                    let mut names = names;
+                    while let Some(event) = events.next().await {
+                        match event {
+                            Event::NameOwner => match Self::initialize_data(&conn).await {
+                                Ok(data) => {
+                                    info!("MPRIS player service new data");
```
MalpenZibo:
```suggestion
                                    debug!("MPRIS player service new data");
```

src/services/mpris/mod.rs
```diff
@@ -0,0 +1,340 @@
...
+                                    names = data.0;
+                                    let _ = output.send(ServiceEvent::Update(data.1)).await;
+                                }
+                                Err(err) => {
+                                    error!("Failed to fetch MPRIS player data: {}", err);
+                                }
+                            },
+                            _ => {
```
MalpenZibo:
```suggestion
                            Event::Volume | Event::Metadata => {
```
I prefer to be more explicit :sweat_smile:

MalpenZibo (changes requested):
First of all: Great Job!!

I left some comments. 

As an addition could be nice to check the `PlaybackStatus` from dbus to show the correct `play` or `pause` icon instead of a generic `PlayPause` icon

src/modules/media_player.rs
mazei513: Was thinking it gets confusing/annoying because I think DBus returns an alphabetical list of services. I was hoping to eventually have it as showing the most recently changed, but that also seemed annoying in other ways.

Will change it to display the first one for now.

mazei513:
> An addition that could be nice is to check the `PlaybackStatus` from dbus to show the correct `play` or `pause` icon instead of a generic `PlayPause` icon

Was trying to keep the changes to a minimum with at least feature parity with #94 and the multiple players support. Will look into it along with checking  [`CanControl`](https://specifications.freedesktop.org/mpris-spec/latest/Player_Interface.html#Property:CanControl) and etc. in a separate PR.

src/modules/media_player.rs
mazei513: Went with "No Title" for now.

I think in the future it can rely on [Track_Id](https://specifications.freedesktop.org/mpris-spec/latest/Player_Interface.html#Simple-Type:Track_Id) as a fallback, though if it has custom formatting support its back to square one. I have not looked at other status bars/shells to see what they do, but would think that's probably the easiest way to figure out what's sensible.

src/services/mpris/mod.rs
mazei513: Fair, I do too 🙂

src/modules/media_player.rs
mazei513: Makes way more sense to refer to the data within the service, at the very least, having a single source for the data.

mazei513:
Have made the changes with the suggested fixes.

src/modules/media_player.rs
```diff
@@ -1,214 +1,164 @@
...
-                .spacing(8),
+    pub fn menu_view(&self, config: &MediaPlayerModuleConfig) -> Element<Message> {
+        match &self.service {
+            None => text("Not connected to MPRIS service").into(),
+            Some(s) => column(
+                s.deref()
+                    .iter()
+                    .flat_map(|d| {
+                        let d = d.clone();
+                        let title = text(Self::get_title(&d, config));
```
MalpenZibo:
```suggestion
                s.iter()
                    .flat_map(|d| {
                        let title = text(Self::get_title(d, config));
```

src/modules/media_player.rs
```diff
@@ -1,214 +1,164 @@
...
-                Some(OnModulePress::ToggleMenu(MenuType::MediaPlayer)),
-            )
+        self.service.as_ref().and_then(|s| {
+            let data = s.deref();
+            match data.len() {
+                0 => None,
+                _ => Some((
+                    row![
+                        icon(Icons::MusicNote),
+                        text(Self::get_title(&data[0], config))
```
MalpenZibo:
```suggestion
        self.service.as_ref().and_then(|s| {
            match s.len() {
                0 => None,
                _ => Some((
                    row![
                        icon(Icons::MusicNote),
                        text(Self::get_title(&s[0], config))
```

MalpenZibo (approved):
:rocket: 
Just two minor things. We also forgot to update the README config section with the new module name.

mazei513:
> We also forgot to update the README config section with the new module name.

It seems correct on my end and when looking at master though. Is my mind fried 😵‍💫???

```markdown
# Media player module configuration
mediaPlayer:
  maxTitleLength: 100 # optional, default 100
```

Also pushed the changes.

MalpenZibo:
> > We also forgot to update the README config section with the new module name.
> 
> It seems correct on my end and when looking at master though. Is my mind fried 😵‍💫???
> 
> ```
> # Media player module configuration
> mediaPlayer:
>   maxTitleLength: 100 # optional, default 100
> ```
> 
> Also pushed the changes.

Ahah no that section it's ok. Last time I forgot to check this other section
```
# Declare which modules should be used and in which position in the status bar.
# This is the list of all possible modules
#  - AppLauncher
#  - Updates
#  - Clipboard
#  - Workspaces
#  - WindowTitle
#  - SystemInfo
#  - KeyboardLayout
#  - KeyboardSubmap
#  - Tray
#  - Clock
#  - Privacy
#  - Settings
# optional, the following is the default configuration
modules:
```

Also, the CI shows an unused `use std::ops::Deref;`

mazei513:
Fixed the use declaration and updated the README.

MalpenZibo:
Great job!

PR #101, opened 2025-02-16
Fixed media player not updating

mazei513:
Simply mutating state doesn't cause a re-render, causing background changes to not get re-rendered.

### DIFF
--- a/src/modules/media_player.rs
+++ b/src/modules/media_player.rs
@@ -46,6 +46,7 @@ impl MediaPlayer {
                 ServiceEvent::Update(d) => {
                     if let Some(service) = self.service.as_mut() {
                         service.update(d);
+                        self.service = Some(service.clone());
                     }
                     Task::none()
                 }


MalpenZibo:
Hum... It seems strange. In theory a message returned from a view widget or a subscription should trigger a re-render and mutating or not mutating the state shouldn't affect this behavior.

I will try to understand better what's the problem here

mazei513:
Yeah seems like the fix in this PR is still quite hit or miss. It works more often but still not 100% which is weird.

MalpenZibo:
I was able to make some progress. It seems that `receive_metadata_changed` produces a lot of events. I'm working on a solution that filters the events if the new events metadata is the same as the old one.

With this change plus another change on how we handle the NameOwner change, the module seems to be working better without skipping any changes.

I will open a PR in the next few days.

mazei513:
That's great, thanks for looking into it. I'll close this PR then?

MalpenZibo:
let's wait until my PR

MalpenZibo:
Ok, I think that we could close this PR in favor of #106

PR #106, opened 2025-03-05
Mediaplayer event filter

MalpenZibo:


MalpenZibo:
@mazei513 I got good results with these changes. Please, if you can take a look and test it :smile:

mazei513:
From a quick test, its working really well. Thank you so much for looking into it.

mazei513:
A new issue popped up that the list of media players don't seem to update when a player goes away, but I'll raise that as a separate issue.

MalpenZibo:
@mazei513 I made other changes that may fix the issue that popped up.

This is an interesting bug, and maybe with these changes I can improve events handling also on other modules

PR #129, opened 2025-04-27
Iwd as an alternative network backend

clotodex:
This very work in progress...


clotodex:
@MalpenZibo I abstracted away the network backend. I would love to get your feedback if that is fine or how you would approach abstracting iwd and nm together. For now I will just continue to create feature-parity with NM and can sort the architecture later.

What I did: I created an async trait called NetworkBackend that iwd and nm implement.
I wanted to pass a boxed backed around in the active state initially, but I had some trouble with async traits and where to pin them etc. 
So instead: NetworkService has a new field BackendChoice which is essentially a router to NM or IWD. Technically we could skip the trait now, but it is nice to know what needs to be implemented if more backends should be supported in the future. I pulled all NM functions into its dbus.rs and created an iwd_dbus folder containing all its proxies. So shared implementation is in mod.rs and otherwise in dedicated files.

clotodex:
Alright, this is now usable.
We are still missing password support - this requires registering agents, which I still need to understand how.
Also device state is not implemented yet - these are also all agents in iwd.
Agent = Listener getting updates live.
This implementation can also be done in a new pull request after, so I will mark this for review.

Let me know what you think @MalpenZibo

MalpenZibo:
OK, I still need to make the code review. Right now, I saw the new trait, and it seems like a good solution to me.

I notice that right now the backend choice is done, trying to connect to NetworkManager, and in case of error, fallback to iwd. I don't know if maybe it's better to let the user select the correct backend in the configuration...

Right now, we only have these two choices, so I think that it could be fine!

clotodex:
I am getting more used to the codebase so a configuration to choose from an enum is pretty easy

However i love this bars "it just works" mentality, so personally i prefer the fallback
Will follow your lead

MalpenZibo:
Ok, let's leave the fallback system for now. In the future we could add a configuration.

I will try to review the code in the next few days! In the meantime you could add an entry in the changelog file (also for the other PRs)

src/services/network/mod.rs
```diff
@@ -273,31 +427,63 @@ impl NetworkService {
...
-                        while let Some(event) = events.next().await {
-                            let mut exit_loop = false;
-                            if let NetworkEvent::WirelessDevice { .. } = event {
-                                exit_loop = true;
+                // TODO: i dont know how to combine the opaque types.. rust streams
+                match choice {
+                    BackendChoice::NetworkManager => {
+                        let nm = NetworkDbus::new(&conn).await.unwrap();
+
+                        match nm.subscribe_events().await {
```
MalpenZibo:
```suggestion
                        match NetworkDbus::new(&conn).await {
                          Ok(nm) => match nm.subscribe_events().await {
                          ...
                                                      Err(err) => {
                                error!("Failed to connect to NetworkManager: {}", err);
                                State::Error
                            }
```

Same thing for the IWD case

src/services/network/mod.rs
```diff
@@ -273,31 +427,63 @@ impl NetworkService {
...
+                            Ok(mut events) => {
+                                while let Some(event) = events.next().await {
+                                    let mut exit_loop = false;
+                                    // TODO: why do we do this?
```
MalpenZibo, line 448: Complete answer: because I should really refactoring this service :sweat_smile: 
I found a better solution working on the `mpris` service that I should adapt to the other services, but I don't have enough time and courage

Long story short: because I had to re-subscribe to the events.

src/services/network/mod.rs
```diff
@@ -310,303 +496,23 @@ impl NetworkService {
...
-    }
 }
 
 impl Service for NetworkService {
     type Command = NetworkCommand;
 
     fn command(&mut self, command: Self::Command) -> Task<ServiceEvent<Self>> {
         debug!("Command: {:?}", command);
+        let conn = self.conn.clone();
+        let bc = self.backend_choice.clone().with_connection(conn);
```
MalpenZibo:
```suggestion
        let bc = self.backend_choice.with_connection(conn);
```

src/services/network/mod.rs
```diff
@@ -203,66 +240,183 @@ impl ReadOnlyService for NetworkService {
     }
 }
 
-impl NetworkService {
-    async fn initialize_data(conn: &zbus::Connection) -> anyhow::Result<NetworkData> {
-        let nm = NetworkDbus::new(conn).await?;
+#[derive(Debug, Clone)]
```
MalpenZibo:
```suggestion
#[derive(Debug, Copy, Clone)]
```

MalpenZibo (changes requested):
Great work! Really!

I didn't carefully check the `iwd` related code, I just made a generic check, and I didn't test it because right now I don't have a PC with `iwd`.

I left only a few comments on the common network part.

Please remember to add an entry in the CHANGELOG file. Something like
```
### Added

- Experimental IWD service
- Add font name configuration
...
```

should work!

src/services/network/mod.rs
clotodex: this does not work

clotodex:
From my side (pending code review) this is now mergable.
This is feature complete to where I wanted it.

There is still one bug with the signal strength not updating live - in theory everything in the implementation should be correct though, so I would rather immediately re-open an issue after merge.

MalpenZibo (approved):
:rocket:

MalpenZibo:
There are just some minor formatting errors to be fixed, and then we're ready to go!

clotodex:
should be ready to merge now

MalpenZibo (approved):
There's some nix issue but I will check it later! Let's merge :partying_face:

PR #130, opened 2025-04-28
fix: network-manager was not listening to device changes

clotodex:
Just noticed this while implementing the iwd branch. I guess this was not on purpose.

### DIFF
--- a/src/services/network/mod.rs
+++ b/src/services/network/mod.rs
@@ -393,11 +393,18 @@ impl NetworkService {
             })
             .boxed();
 
-        // When devices list change I need to update the wireless device state changes
         let wireless_ac = nm.wireless_access_points().await?;
 
+        // When devices list change I need to update the wireless device state changes
         let mut device_state_changes = Vec::with_capacity(wireless_ac.len());
-        for ac in wireless_ac.clone() {
+        // When devices list change I need to update the access points changes
+        let mut ac_changes = Vec::with_capacity(wireless_ac.len());
+        // When devices list change I need to update the wireless strength changes
+        let mut strength_changes = Vec::with_capacity(wireless_ac.len());
+
+        for ac in wireless_ac.iter() {
+            let ssid = ac.ssid.clone();
+
             let dp = DeviceProxy::builder(conn)
                 .path(ac.device_path.clone())?
                 .build()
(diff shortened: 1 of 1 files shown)


src/services/network/mod.rs
MalpenZibo, line 1: I think that we could do everything in a single for loop. I don't know why I wrote everything in multiple loop O.o
```rust
        let wireless_ac = nm.wireless_access_points().await?;

        // When devices list change I need to update the wireless device state changes
        let mut device_state_changes = Vec::with_capacity(wireless_ac.len());
        // When devices list change I need to update the access points changes
        let mut ac_changes = Vec::with_capacity(wireless_ac.len());
        // When devices list change I need to update the wireless strength changes
        let mut strength_changes = Vec::with_capacity(wireless_ac.len());

        for ac in wireless_ac.iter() {
            let ssid = ac.ssid.clone();

            let dp = DeviceProxy::builder(conn)
                .path(ac.device_path.clone())?
                .build()
                .await?;

            device_state_changes.push(
                dp.receive_state_changed()
                    .await
                    .filter_map({
                        let ssid = ssid.clone();
                        move |val| {
                            let ssid = ssid.clone();
                            async move {
                                let val = val.get().await;
                                let val = val.map(DeviceState::from).unwrap_or_default();

                                if val == DeviceState::NeedAuth {
                                    debug!("Request password for ssid {}", ssid);
                                    Some(NetworkEvent::RequestPasswordForSSID(ssid))
                                } else {
                                    None
                                }
                            }
                        }
                    })
                    .boxed(),
            );

            let dp = WirelessDeviceProxy::builder(conn)
                .path(ac.device_path.clone())?
                .build()
                .await?;

            ac_changes.push(
                dp.receive_access_points_changed()
                    .await
                    .then({
                        let conn = conn.clone();
                        move |_| {
                            let conn = conn.clone();
                            async move {
                                let nm = NetworkDbus::new(&conn).await.unwrap();
                                let wireless_access_point =
                                    nm.wireless_access_points().await.unwrap_or_default();
                                debug!("access_points_changed {:?}", wireless_access_point);

                                NetworkEvent::WirelessAccessPoint(wireless_access_point)
                            }
                        }
                    })
                    .boxed(),
            );

            let app = AccessPointProxy::builder(conn)
                .path(ac.path.clone())?
                .build()
                .await?;

            strength_changes.push(
                app.receive_strength_changed()
                    .await
                    .then(move |val| {
                        let ssid = ssid.clone();
                        async move {
                            let value = val.get().await.unwrap_or_default();
                            debug!("Strength changed value: {}, {}", &ssid, value);
                            NetworkEvent::Strength((ssid.clone(), value))
                        }
                    })
                    .boxed(),
            );
        }
```

MalpenZibo (changes requested):
Ups :sweat_smile:

src/services/network/mod.rs
clotodex: Will check it out tomorrow but lgtm

clotodex: implemented your changes. I dont have nm - so cant test it. Otherwise ready to merge from my side.

MalpenZibo:
I make some tests. There's something strange. With this fix, something doesn't work well when I switch to a new Wi-Fi, and the system asks for the password. This event is triggered multiple times and asks for the password of the current wifi and for the new one.

I need to investigate more on this O.o sadly the network service is a mess :smiling_face_with_tear:

clotodex:
Without checking the code - I ran into a similar thing in IWD.
What happend for me: trying to connect WITH a password can trigger the underlying network to still ask for the password - which would immediately be resolved but STILL trigger a request message in the service code.
In your case specifically this might trigger a device state change but keep it in "needs auth" which triggers the message.
Probably some (print) debugging needed.
And then it obviously gets triggered again when it realizes it needs a password for the new network

This is surprising thought that that was NOT the case before.

clotodex:
I dont understand the failing CI. Running the same clippy command locally shows no issues.

MalpenZibo:
> I dont understand the failing CI. Running the same clippy command locally shows no issues.

It could be a different clippy version? Have you already tried with a rustup update?

clotodex:
> It could be a different clippy version? Have you already tried with a rustup update?

Actually, given iwd merged, the whole fix needs to be ported to there anyways, let me give that a go to.
Did you get around testing further if this fix did not break anything else?

MalpenZibo:
The only thing is the wifi password is requested twice.

clotodex:
That was (is) not the case with the version before?

MalpenZibo:
The problem doesn't happen in the main version. With this branch, when I switch wifi networks, ashell requests the wifi password twice. I will do more tests to understand why

MalpenZibo:
I will close this PR. I plan a refactor of the network service in the future, so it doesn't make sense to find the root cause of this issue

Issue #128, opened 2025-04-26
Missing features to switch from waybar


clotodex:
I find this project super interesting and have long been looking for an opinionated out-of-the-box bar. Love pretty much everything about it!

There are however a few things missing so I would switch from waybar and are creating this issue as a tracking issue for myself as well as for feedback what is compatible with this repo, and what would not get merged.

A live comparison - as you can see, not far apart.

![Image](https://github.com/user-attachments/assets/8963a200-7fde-418e-bd86-30653beab87b)

The missing features for me:

Windows/Workspaces:
- Multimonitor: would like the active workspace on the other monitor to be a different color
- Alerted windows support (as you can see with the "!" and red color)

Broken modules:
- Network - I use raw iwd (could be implemented with iwdrs)
- Airplane mode

New Modules:
- Standalone: Connected bluetooth device
- Standalone: Wifi (strength / status)
- SwayNC for notification management

Custom Modules:
- IconModule - Icon that launches a specified command (e.g. qr code scanner, ..) - similar to launcher
- IconDynamicModule - IconModule, but Icon can switch based on command output that could be polled - super flexible to just add any type of extra info

Style (preference)
- different backgrounds of modules (especially the workspaces I would like to be way less intrusive)
- collapse behind icon (e.g. system info) - could be a collapsed icon, if i wanna see it i can hover on the icon
- MediaPlayer - spotify is shown after browser - I would like to specify preference or switch it with click (could be separate issue)
- Volume indication or percent directly in module
- I am using "hover for more info" a lot (e.g. calendar for date, wifi network for wifi, ...)

MalpenZibo:
Thanks for the feedback. In the next few days I should return from my holiday.

I will check what could be done!

clotodex:
working on iwd in #129 (the state of iwd dbus in rust was virtually non-existant, and async traits with impl Stream are a b***). Feel free to have a look, will continue in the upcoming week.

clotodex:
Also implemented minimal custom module support in #131

MalpenZibo:
Ok, I'm back!

Let's start:
- IconModule and IconDynamicModule: I have already thought about it. It should be an easy change and could replace the AppLauncher and the Clipboard module.
- File system usage (and other info like network download/upload speed): I opened a PR that improves the SystemInfo module to support these use cases
[...]

PR #131, opened 2025-04-28
Feature: custom modules (name, icon, command)

clotodex:
- config parser to allow multiple custom modules that are parsed based on their name
- custom module that will render configurable nerdfont icon
- custom command that gets executed on click

```toml
[modules]
# The modules that will be displayed on the left side of the status bar
left = [
	"appLauncher",
    [ "qr-scan", "updates"],
	"windowTitle",
]
center = [ "workspaces" ]
right = [ "qr-scan", "clipboard", "systemInfo", "tray", [ "clock", "privacy", "settings" ] ]

[[CustomModule]]
name = "qr-scan"
icon = "󰐲"
command = "screenshot-area-scan-qr"
```

To be discussed:
To parse the config I ran into serdes limitation of not being able to parse nested enums with different taggings.
So instead I am now parsing the ModuleNames directly from string - this also has benefits, like e.g. allowing any spelling to map to the module.


MalpenZibo:
Ok, I saw the code, and it seems very promising.

The problem with serde seems related to the `toml` crate.
I tried with this enum
```
#[derive(Deserialize, Serialize, Clone, Debug, PartialEq, Eq)]
#[serde(rename_all = "camelCase")]
pub struct CustomModule {
    pub icon: String,
    pub command: String,
}

#[derive(Deserialize, Serialize, Clone, Debug, PartialEq, Eq)]
#[serde(rename_all = "camelCase")]
pub enum ModuleName {
    AppLauncher,
    Updates,
    Clipboard,
    Workspaces,
    WindowTitle,
    SystemInfo,
    KeyboardLayout,
    KeyboardSubmap,
    Tray,
    Clock,
    Privacy,
    Settings,
    MediaPlayer,
    CustomModule(CustomModule),
}
```

With `json_serde`, it seems to be working fine. I need to understand if I like your solution more or this one.

Your solution has some advantages. Starting with the enum naming, `ModuleName` indicates only the module names and not the module config data. For this reason, maybe splitting the two things could be better. It would be possible to place the same module in multiple locations without redefining the config (I don't think it's something useful, but right now it could be done with every module).

The point is:
I like your solution, but I don't like this change (and everything related) too much
```
pub enum ModuleDef {
    Single(String),
    Group(Vec<String>),
}
```

I prefer using the ModuleName. I know that's a limitation of toml_serde, and for this reason, I'm thinking of switching to json_serde. I haven't released a version with the toml config, so it could be THE moment to make this change. I know everyone using the `dev` version will hate me sooo much :sweat_smile: 

What do you think?

clotodex:
Yeah this was my going back and forth when setting it up

In the grand scheme of things i am using nixos so i will likely anyways build my own nix based module in the end so please dont follow my preferences here

Why i like the toml method is:
- toml overall has a really nice modular design, especially the custom modules
- json just works but is in my eyes a quite clunky config format

Why i think toml is maintainable:

I think (as i did) i would throw out the auto-parsing here and parse ALL modules as strings 
Then it is as "simple" as before, no big code duplication and stays "single & group" - simply would not auto-parse. However we can easily give then also custom error messages like "module not defined".
If you dont want the implementation to "leak" into the other files, we could do some From/Into magic as well.

These are my two cents - hope it helps

clotodex:
ALSO I already implemented dynamic custom modules including an alert system (just not pr ready yet)
This allows you to specify a command that yields updates and can change text and icon and alert status (red dot) dynamically

https://github.com/clotodex/ashell-iwd/tree/feature%2Fdynamic-custom-components

clotodex:
The dynamic modules work like this:

- listen command: a command that when executed yields jsons containing two fields:
  - alt: what alternative it is (e.g. swaync: notification, dnd-notification, none, ...) => based on this we will adjust icons etc
  - text: text the module wants to display (e.g. number of notifications with swaync)
  - example swaync output: `{"text": "16", "alt": "notification", "tooltip": "16 Notifications", "class": "notification"}`

And can be configured like this:
The icons config follows a regex, and if that regex is true, will display the other icon.
Alert is also a regex, so if there is a match, the red dot will be shown at the right top
```
[[CustomModule]]
name = "swaync"
icon = "" # default icon
icons.'dnd.*' = "" # icon if regex of "alt" is true, multiple such options can be given
alert = ".*notification" # to show a red dot next to it
listenCmd = "swaync-client -swb"
command = "swaync-client -t -sw"
```
image

For simplicity I followed waybars json format here, which solves my #128 

I think this is ready for a first review now as well

src/config.rs
```diff
@@ -348,14 +348,44 @@ pub enum ModuleName {
...
+            "mediaPlayer" => ModuleName::MediaPlayer,
+            _ => ModuleName::Custom(value),
+        }
+    }
+}
+impl<'a> From<&'a str> for ModuleName {
+    fn from(value: &'a str) -> ModuleName {
+        String::into(value.to_owned())
+    }
 }
```
MalpenZibo: Let's try this. It's a custom deserializer. Tries to deserialize every entry in a ModuleName, and in case there isn't a match it uses the Custom ModuleName

```suggestion
impl<'de> Deserialize<'de> for ModuleName {
    fn deserialize<D>(deserializer: D) -> Result<ModuleName, D::Error>
    where
        D: Deserializer<'de>,
    {
        struct ModuleNameVisitor;

        impl Visitor<'_> for ModuleNameVisitor {
            type Value = ModuleName;

            fn expecting(&self, formatter: &mut std::fmt::Formatter) -> std::fmt::Result {
                formatter.write_str("a camelCase string representing a ModuleName")
            }

            fn visit_str<E>(self, value: &str) -> Result<ModuleName, E>
            where
                E: Error,
            {
                Ok(match value {
                    "appLauncher" => ModuleName::AppLauncher,
                    "updates" => ModuleName::Updates,
                    "clipboard" => ModuleName::Clipboard,
                    "workspaces" => ModuleName::Workspaces,
                    "windowTitle" => ModuleName::WindowTitle,
                    "systemInfo" => ModuleName::SystemInfo,
                    "keyboardLayout" => ModuleName::KeyboardLayout,
                    "keyboardSubmap" => ModuleName::KeyboardSubmap,
                    "tray" => ModuleName::Tray,
                    "clock" => ModuleName::Clock,
                    "privacy" => ModuleName::Privacy,
                    "settings" => ModuleName::Settings,
                    "mediaPlayer" => ModuleName::MediaPlayer,
                    other => ModuleName::Custom(other.to_string()),
                })
            }
        }

        deserializer.deserialize_str(ModuleNameVisitor)
    }
}
```

src/config.rs
```diff
@@ -348,14 +348,44 @@ pub enum ModuleName {
...
 
 #[derive(Deserialize, Clone, Debug)]
 #[serde(rename_all = "camelCase")]
 #[serde(untagged)]
 pub enum ModuleDef {
-    Single(ModuleName),
-    Group(Vec<ModuleName>),
+    Single(String),
+    Group(Vec<String>),
```
MalpenZibo: This could return to be a ModuleName

src/modules/mod.rs
```diff
@@ -59,8 +62,18 @@ impl App {
 
         for module_def in modules_def {
             row = row.push_maybe(match module_def {
-                ModuleDef::Single(module) => self.single_module_wrapper(*module, id, opacity),
-                ModuleDef::Group(group) => self.group_module_wrapper(group, id, opacity),
+                // life parsing of string to module
+                ModuleDef::Single(module) => {
+                    self.single_module_wrapper(module.to_owned().into(), id, opacity)
```
MalpenZibo:
```suggestion
                    self.single_module_wrapper(module, id, opacity)
```

src/modules/mod.rs
```diff
@@ -59,8 +62,18 @@ impl App {
...
+                }
+                ModuleDef::Group(group) => self.group_module_wrapper(
+                    &group
+                        .iter()
+                        .map(|m| m.to_owned().into())
+                        .collect::<Vec<_>>(),
+                    id,
+                    opacity,
+                ),
```
MalpenZibo:
```suggestion
                ModuleDef::Group(group) => self.group_module_wrapper(
                    &group,
                    id,
                    opacity,
                ),
```

src/modules/mod.rs
```diff
@@ -71,10 +84,12 @@ impl App {
         modules_def
             .iter()
             .flat_map(|module_def| match module_def {
-                ModuleDef::Single(module) => vec![self.get_module_subscription(*module)],
+                ModuleDef::Single(module) => {
+                    vec![self.get_module_subscription(module.to_owned().into())]
```
MalpenZibo: Well... there are some other places where we should fix this kind of stuff. But in this way we can avoid some `clone`

```suggestion
                    vec![self.get_module_subscription(module)]
```

MalpenZibo (commented):
Maybe I found a solution to make me happy about the change to how we handle the `ModuleName` deserialization :smile:

MalpenZibo:
> The dynamic modules work like this:
> 
>     * listen command: a command that when executed yields jsons containing two fields:
>       
>       * alt: what alternative it is (e.g. swaync: notification, dnd-notification, none, ...) => based on this we will adjust icons etc
>       * text: text the module wants to display (e.g. number of notifications with swaync)
>       * example swaync output: `{"text": "16", "alt": "notification", "tooltip": "16 Notifications", "class": "notification"}`
> 
> 
> And can be configured like this: The icons config follows a regex, and if that regex is true, will display the other icon. Alert is also a regex, so if there is a match, the red dot will be shown at the right top
> 
> ```
> [[CustomModule]]
> name = "swaync"
> icon = "" # default icon
> icons.'dnd.*' = "" # icon if regex of "alt" is true, multiple such options can be given
> alert = ".*notification" # to show a red dot next to it
> listenCmd = "swaync-client -swb"
> command = "swaync-client -t -sw"
> ```
> 
> image
> 
> For simplicity I followed waybars json format here, which solves my #128
> 
> I think this is ready for a first review now as well

Ok, I had to check this part but again, great work, very appreciated :100:

clotodex:
bit swamped rn, but custom deserializer is a good shout
Will do the changes as soon as I can get to it

src/app.rs
```diff
@@ -205,6 +219,14 @@ impl App {
                 }
                 Task::none()
             }
+            Message::LaunchCommand(command) => {
+                utils::launcher::execute_command(command);
+                Task::none()
+            }
+            Message::CustomUpdate(name, message) => {
+                self.custom.get_mut(&name).unwrap().update(message);
```
MalpenZibo: We should handle the unwrap

src/modules/custom_module.rs
```diff
@@ -0,0 +1,223 @@
...
+        _cursor: Cursor,
+    ) -> Vec<Geometry> {
+        let cache = Cache::new(); // Use a local cache for simplicity here
+
+        vec![cache.draw(renderer, bounds.size(), |frame| {
+            let center = frame.center();
+            // Use a smaller radius so the circle doesn't touch the canvas edges
+            let radius = 2.0; // Creates a 4px diameter circle
+            let circle = Path::circle(center, radius);
+            frame.fill(&circle, Color::from_rgb8(243, 139, 168)); // Red color
```
MalpenZibo: Here we should use the theme danger color instead of a fixed one

src/modules/custom_module.rs
```diff
@@ -0,0 +1,223 @@
...
+
+        if let Some(icons_map) = &config.icons {
+            for (pattern_str, icon_str) in icons_map {
+                if let Ok(re) = regex::Regex::new(pattern_str) {
+                    if re.is_match(&self.data.alt) {
+                        icon_element = icon_raw(icon_str.clone());
+                        break; // Use the first match
+                    }
+                } else {
+                    error!("Invalid regex pattern in config: {}", pattern_str);
```
MalpenZibo: It would be great to move this part during the configuration evaluation and store the Regex in the configuration.

src/modules/custom_module.rs
```diff
@@ -0,0 +1,223 @@
...
+        let padded_icon_container = container(icon_element).padding([0, 1]);
+
+        let mut show_alert = false;
+        if let Some(alert_pattern) = &config.alert {
+            if let Ok(re) = regex::Regex::new(alert_pattern) {
+                if re.is_match(&self.data.alt) {
+                    show_alert = true;
+                }
+            } else {
+                error!("Invalid alert regex pattern in config: {}", alert_pattern);
```
MalpenZibo: same thing

MalpenZibo:
I merged this PR https://github.com/MalpenZibo/ashell/pull/135, so probably there are some code alignments to do

clotodex:
> I merged this PR #135, so probably there are some code alignments to do

Rebased with main and fixed all related config errors etc

Something I noticed is that the naming of the toml modules is not uniform
system vs mediaPlayer vs CustomModules

Is there something you prefer? should all be camelCase or snake_case? Or since it is a module upper CamleCase? Just let me know your preference and I will change it accordingly

clotodex:
all your suggestions implemented - thanks a lot for already writing the deserializer code!

This now looks ready to me.

src/app.rs
```diff
@@ -81,18 +83,29 @@ pub enum Message {
...
 impl App {
     pub fn new((logger, config): (LoggerHandle, Config)) -> impl FnOnce() -> (Self, Task<Message>) {
         || {
             let (outputs, task) = Outputs::new(config.appearance.style, config.position);
 
+            let names = config
+                .custom_modules
+                .iter()
+                .map(|o| o.name.clone())
+                .collect::<Vec<_>>();
```
```suggestion
            let custom: HashMap<String, Custom> = config
                .custom_modules
                .iter()
                .map(|o| (o.name.clone(), Custom::default()))
                .collect();

```

src/app.rs
```diff
@@ -81,18 +83,29 @@ pub enum Message {
...
             (
                 App {
                     logger,
                     outputs,
                     app_launcher: AppLauncher,
+                    custom: names
+                        .into_iter()
+                        .map(|n| (n.clone(), Custom::default()))
+                        .collect(),
```
```suggestion
                    custom,
```

src/modules/custom_module.rs
```diff
@@ -0,0 +1,223 @@
...
+                channel(10, async move |mut output| {
+                    let command = Command::new("bash")
+                        .arg("-c")
+                        .arg(&check_cmd)
+                        .stdout(Stdio::piped())
+                        .spawn();
+
+                    match command {
+                        Ok(mut child) => {
+                            let mut reader = BufReader::new(child.stdout.take().unwrap()).lines();
```
MalpenZibo:
```suggestion
                            if let Some(stdout) = child.stdout.take() {
                                let mut reader = BufReader::new(stdout).lines();

```

src/modules/custom_module.rs
```diff
@@ -0,0 +1,223 @@
...
+                            tokio::spawn(async move {
+                                let status = child
+                                    .wait()
+                                    .await
+                                    .expect("child process encountered an error");
+
+                                println!("child status was: {}", status);
+                            });
+
+                            while let Some(line) = reader.next_line().await.unwrap() {
```
MalpenZibo:
```suggestion
                                while let Some(line) = reader.next_line().await.ok().flatten() {

```

src/modules/mod.rs
```diff
@@ -225,12 +231,21 @@ impl App {
...
             ModuleName::AppLauncher => self.app_launcher.view(&self.config.app_launcher_cmd),
+            ModuleName::Custom(name) => {
+                let c = self.config.custom_modules.iter().find(|m| &m.name == name);
+                if let Some(mc) = c {
+                    self.custom.get(name).unwrap().view(mc)
+                } else {
+                    error!("Custom module `{}` not found", name);
+                    None
+                }
+            }
```
MalpenZibo:
```suggestion
            ModuleName::Custom(name) => {
                if let Some(v) = self
                    .config
                    .custom_modules
                    .iter()
                    .find(|m| &m.name == name)
                    .and_then(|mc| self.custom.get(name).map(|cm| cm.view(mc)))
                {
                    v
                } else {
                    error!("Custom module `{}` not found", name);
                    None
                }
            }

```

src/modules/mod.rs
```diff
@@ -252,9 +267,18 @@ impl App {
...
             ModuleName::AppLauncher => self.app_launcher.subscription(()),
+            ModuleName::Custom(name) => {
+                let c = self.config.custom_modules.iter().find(|m| &m.name == name);
+                if let Some(mc) = c {
+                    self.custom.get(name).unwrap().subscription(mc)
+                } else {
+                    error!("Custom module `{}` not found", name);
+                    None
+                }
+            }
```
MalpenZibo:
```suggestion
            ModuleName::Custom(name) => {
                if let Some(sub) = self
                    .config
                    .custom_modules
                    .iter()
                    .find(|m| &m.name == name)
                    .and_then(|mc| self.custom.get(name).map(|cm| cm.subscription(mc)))
                {
                    sub
                } else {
                    error!("Custom module `{}` not found", name);
                    None
                }
            }

```

MalpenZibo (changes requested):
Great, we're almost there.

I left you some comments, mainly related to some unwrapping.

Sorry if I'm a bit pedantic, I really appreciate your effort

clotodex:
Haha thank you for the review! I see a lot of "lets get a quick thing working" stuff still stuck around - on it :D

src/modules/custom_module.rs
clotodex: needed to write a newtype so that I can use the regex in the hashmap and deserialize it, but I like it (with deref the rest of the application can use it as a regex directly)

clotodex:
ready to go i think - sorry for making a git mess - i think resolving conflicts from github (just cargo-toml imports) pulls in main again :D

MalpenZibo (approved):
Perfect! :tada: :rocket: :+1: :100: 

If you have time, please add an entry in the changelog and update the readme with some documentation about these custom modules. 
Otherwise, in the next few days, I will proceed with the merge and try to write something about it in the release PR.

clotodex:
what do you think of this?

MalpenZibo:
> what do you think of this?

:tada:

PR #138, opened 2025-05-10
Fix division by zero in system_info/network speed

ineu:
If there were no traffic since the previous check, the code panicked with 
```
ERROR [ashell] Panic: panicked at src/modules/system_info.rs:107:17:
attempt to divide by zero 
```

This PR fixes it


MalpenZibo (approved):
Wups! 😅

PR #139, opened 2025-05-10
Log warning when failing to read config file

rahatarmanahmed:
I didn't realize that the config path was moved, and there was no indication why. This warning message would have helped :)


MalpenZibo (approved):
:rocket:

PR #140, opened 2025-05-11
Simplify style of "expand" button on wifi/bt buttons

rahatarmanahmed:
Okay I totally understand that this is a personal preference thing, but I do think it looks better this way for me. Also, it seems more like how the ChromeOS control panel styles there (which I assume is an inspiration since ashell looks very similar).

Now I don't expect to merge this as-is (unless you like it too!), but would you be open to adding an appearance config option for the vertical bar and replacing the 3-dot icon? 

image


MalpenZibo:
I like it! More than the actual implementation. I think that we could simply change the current implementation with this one. I only need to make a bunch of test :+1: 

You could add an entry in the CHANGELOG file

src/modules/settings/mod.rs
```diff
@@ -741,13 +739,12 @@ fn quick_setting_button<'a, Msg: Clone + 'static>(
     button(
         Row::new()
             .push(main_content)
-            .push_maybe(with_submenu.as_ref().map(|_| vertical_rule(1)))
             .push_maybe(with_submenu.map(|(menu_type, submenu, msg)| {
                 button(
                     container(icon(if Some(menu_type) == submenu {
                         Icons::Close
                     } else {
-                        Icons::VerticalDots
```
MalpenZibo, line 750: Could you remove the VerticalDots from the enum? I don't plan to let the user change this setting, and in the end, I prefer this style, so the VerticalDots shouldn't be used in other parts of the codebase

MalpenZibo (changes requested):

src/modules/settings/mod.rs
rahatarmanahmed: Done!

rahatarmanahmed:
> I like it! More than the actual implementation. I think that we could simply change the current implementation with this one. I only need to make a bunch of test 👍
> 
> You could add an entry in the CHANGELOG file

Awesome! I've added the CHANGELOG entry.

rahatarmanahmed (commented):
Made requested changes (lmk if I did something wrong here, haven't used github in a long while. not used to this review changes flow)

MalpenZibo (approved):
:rocket:

PR #148, opened 2025-05-17
Remove backround from battery info text

rahatarmanahmed:
Prior to this commit it had a very similar style as the other buttons in the settings panel, which misleadingly makes it look like a button itself even though it has no action. IMO it also looks cleaner w/o the background

image


src/modules/settings/upower.rs
```diff
@@ -55,34 +55,27 @@ impl BatteryData {
                 }),
                 ..Default::default()
             });
+
+            let secondary_font = font::Font {
+                weight: font::Weight::Light,
+                ..font::Font::DEFAULT
+            };
```
MalpenZibo: Could you remove this font style? I mean, I like it, but on my system, without specifying a font in the config, for some reason, I get a different font. 

Until I discover what causes this behavior, I prefer to avoid other users experiencing this problem.

MalpenZibo (changes requested):
Very nice! I like it

You could add a Changelog entry if you like even if it's a small changes

src/modules/settings/upower.rs
rahatarmanahmed: Hmm, I wonder what fonts it is matching against? I have Noto fonts installed so I guess it just uses those. What do you get for these two commands?:

```sh
❯ fc-match sans-serif
NotoSans-Regular.ttf: "Noto Sans" "Regular"
❯ fc-match sans-serif:weight=light
NotoSans-Light.ttf: "Noto Sans" "Light"
```

MalpenZibo:
```
fc-match sans-serif
DejaVuSans.ttf: "DejaVu Sans" "Book"
fc-match sans-serif:weight=light
DejaVuSans-ExtraLight.ttf: "DejaVu Sans" "ExtraLight"
```

And here a screenshot
image

image

MalpenZibo: ok, the problem seems related to the fact that "DejaVu" doesn't have a Light version but only an ExtraLight one. And for some reason uses a mono font instead.

If instead of `weight: font::Weight::Light,` I place `weight: font::Weight::ExtraLight,` everything works as expected

rahatarmanahmed: Haha, but then it started showing a different font for me :)

I'll just leave it out then. Doesn't look that bad w normal weight.

MalpenZibo (approved):
:+1:

PR #150, opened 2025-05-17
Fix opacity for Group modules

### LINKED ISSUE #149: Opacity overlaps when using grouped modules
Do something like this:

```toml
[modules]
right = ["Tray", ["Clock", "Settings"]]

```

and you get this:

Image

I would have expected the opacity to be the same

rahatarmanahmed:
Fixes #149


MalpenZibo (approved):
:+1:  Nice!

PR #151, opened 2025-05-18
Add an option to specify the number of workspaces

### LINKED ISSUE #136: Add an option to specify the number of workspaces
Hello and thank you for the great bar! Found it today, trying to replicate what I have in waybar but without all the unwieldy jsonc stuff. Here's one thing:

In hyprland workspaces are dynamic, they only exist once you create a window in a workspace or switch to the workspace. There's an option `enableWorkspaceFilling` which shows inactive workspaces, but only those that come *before* the last active one.
Obviously, because of the dynamic nature of hyprland workspaces ashell cannot know in advance how many workspaces a user needs, so the current behavior is the only possible. However, I find it helpful to see all workspaces that I use. There's the screenshot of waybar and ashell running together:

Image

In waybar I achieve this by declaring the workspaces in the config:
```
    "hyprland/workspaces": {
        "persistent-workspaces": {
            "1": [],
            "2": [],
            "3": [],
            "4": [],
            "5": [],
            "6": [],
            "7": [],
            "8": [],
            "9": [],
        },
    },
```
This, however, is a bit too verbose for such a simple thing. Assuming that workspaces normally go in order, I suggest the following change:
1. Add another option to the `workspaces` group, something like `maxWorkspaces` (or maybe `lastWorkspace`?), integer
2. In the `enableWorkspaceFilling` handler use this value as the last workspace number, fall back to the current behavior if not specified

ineu:
Resolves #136 

The new option `max_workspaces` only works if `enable_workspace_filling` is set to true.

There's also a bit of refactoring to simplify and speed up the code.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -87,17 +87,15 @@ impl App {
     pub fn new((logger, config): (LoggerHandle, Config)) -> impl FnOnce() -> (Self, Task<Message>) {
         || {
             let (outputs, task) = Outputs::new(config.appearance.style, config.position);
-            let workspaces_config = &config.workspaces.clone();
 
             (
                 App {
                     logger,
-                    config,
                     outputs,
                     app_launcher: AppLauncher,
                     updates: Updates::default(),
                     clipboard: Clipboard,
-                    workspaces: Workspaces::new(workspaces_config),
+                    workspaces: Workspaces::new(&config.workspaces),
                     window_title: WindowTitle::default(),
                     system_info: SystemInfo::default(),
                     keyboard_layout: KeyboardLayout::default(),
@@ -107,6 +105,7 @@ impl App {
                     privacy: Privacy::default(),
                     settings: Settings::default(),
                     media_player: MediaPlayer::default(),
+                    config,
                 },
                 task,
             )
--- a/src/app.rs
+++ b/src/app.rs
@@ -87,7 +87,8 @@ impl App {
     pub fn new((logger, config): (LoggerHandle, Config)) -> impl FnOnce() -> (Self, Task<Message>) {
         || {
             let (outputs, task) = Outputs::new(config.appearance.style, config.position);
-            let enable_workspace_filling = config.workspaces.enable_workspace_filling;
+            let workspaces_config = &config.workspaces.clone();
+
             (
                 App {
                     logger,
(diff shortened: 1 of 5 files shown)


src/app.rs
```diff
@@ -87,7 +87,8 @@ impl App {
     pub fn new((logger, config): (LoggerHandle, Config)) -> impl FnOnce() -> (Self, Task<Message>) {
         || {
             let (outputs, task) = Outputs::new(config.appearance.style, config.position);
-            let enable_workspace_filling = config.workspaces.enable_workspace_filling;
+            let workspaces_config = &config.workspaces.clone();
+
```
MalpenZibo: No need to clone
```suggestion
            let workspaces = Workspaces::new(&config.workspaces);
```

src/app.rs
```diff
@@ -96,7 +97,7 @@ impl App {
                     app_launcher: AppLauncher,
                     updates: Updates::default(),
                     clipboard: Clipboard,
-                    workspaces: Workspaces::new(enable_workspace_filling),
+                    workspaces: Workspaces::new(workspaces_config),
```
```suggestion
                    workspaces,
```

src/modules/workspaces.rs
```diff
@@ -32,92 +33,109 @@ pub struct Workspace {
...
-                });
+    if !config.enable_workspace_filling || normal.is_empty() {
+        // nothing more to do, early return
+        result.sort_by_key(|w| w.id);
+        return result;
+    };
+
+    // To show workspaces that don't exist in Hyprland we need to create fake ones
+    let existing_ids = normal.iter().map(|w| w.id).collect_vec();
+    let mut max_id = *existing_ids.iter().max().unwrap();
```
MalpenZibo: maybe it's better to use `unwrap_or_default()` just in case :sweat_smile:

MalpenZibo (changes requested):
Great! Only some minor suggestions!

src/modules/workspaces.rs
ineu: Well, there must be some element in the vector anyway, because we have the check above:
```rust
if !config.enable_workspace_filling || normal.is_empty() {
```
I'll try the default of 0 though. If hyprland always starts counting workspaces from 1 and if special workspaces are always negative, then 0 should be safe.

ineu:
Pushed the fixes. Moved `config` to the bottom of the App initializer because it moves the ownership so we can't use parts of the config for the modules. I haven't introduced a variable for `config.workspaces` because I can imagine more modules needing config in the future, and each of them would require a separate variable. So the approach from c4a1999 is kind of future-proof

ineu:
Actually I though it would be great to pass a module's config to a module's constructor, something like
```rust
pub struct Workspaces {
    workspaces: Vec<Workspace>,
   config: &WorkspacesModuleConfig,
}
```
But in order for the config reloading to work the config must be passed as a pointer, and it involves some lifetime management, that is a bit too much for my Rust skills :)

MalpenZibo (approved):
:rocket:

PR #166, opened 2025-06-07
Use 'alias' instead of 'name' for bluetooth

beeender:
'name' may not exist and errors will be reported like:

"""
ERROR [ashell::services::bluetooth] Failed to initialize bluetooth service: org.freedesktop.DBus.Error.InvalidArgs: No such property 'Name'
ERROR [ashell::services::bluetooth] Bluetooth service error
"""

'Alias' is recommended by bluez doc:
https://git.kernel.org/pub/scm/bluetooth/bluez.git/tree/doc/org.bluez.Device.rst#n238

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Bluetooth: use alias instead of name for device name
+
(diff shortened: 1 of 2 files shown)


MalpenZibo:
Nice! Could you add an entry in the changelog?

Something like:
```
...
## [Unreleased]

### Fixed

- Bluetooth: use alias instead of name for device name

...
```

beeender:
done

PR #169, opened 2025-06-11
Add window_title option for class instead of title

Pebor:
WindowTitle as the name implies only displays the title, but in for example waybar, I prefer to only show the class of the window, as that is less distracting to me.

This PR adds a new window_title configuration with modes [Title, Class] and I also moved `truncate_title_after_length` to this config, as it seemed to belong there (?)

Maybe the name "WindowTitle" doesn't now make that much sense, or maybe more Client info could be made accessible through a broader "WindowInfo" component with formatting in the future.

image
image


src/modules/window_title.rs
```diff
@@ -9,29 +9,38 @@ use std::{
 
 use super::{Module, OnModulePress};
 
+fn get_window(config: &WindowTitleConfig) -> Option<String> {
+    Client::get_active().ok().and_then(|w| {
+        w.map(|w| match config.mode {
+            crate::config::WindowTitleMode::Title => w.title,
```
MalpenZibo: Here we could avoid the entire path with an `use crate::config etc etc`

MalpenZibo (approved):
Nice! WindowTitle as module name now doesn't seem right. 

I think that in the future, I will probably change everything to something like `Client` module or `Window` module.

Anyway, could you add an entry in the changelog?

MalpenZibo:
@Pebor, any news on this? Otherwise, if you don't mind, I can merge this PR and open a new one to address the changelog entry and the other minor comments

Pebor:
@MalpenZibo Sorry! Life got in the way. Added an entry and the imports. Tell me if you need anything!

PR #97, opened 2025-01-31
96 Fix NixOs bluetooth service

MalpenZibo:
Closes #96

PR #173, opened 2025-06-17
158 Airplane button improvement

### LINKED ISSUE #158: Airplane mode button does nothing when clicked
The airplane mode button in the settings doesn't seem to do anything. It doesn't toggle its state, not affect my internet connection.

I tried running `ashell` from my terminal to see any logs, here's my full output, even after clicking the button:
```
❯ ashell
INFO [ashell::config] Reading config file
WARN [ashell::services::privacy] Failed to connect to webcam: No such file or directory (os error 2)
WARN [ashell::services::brightness] No backlight devices found
ERROR [ashell::services::brightness] Failed to access to brightness files: No backlight devices found
ERROR [ashell::services::brightness] Brightness service error
ERROR [ashell::services::upower] Failed to initialize upower service: org.freedesktop.DBus.Error.ServiceUnknown: The name org.freedesktop.UPower was not provided by any .service files
WARN [wgpu_hal::gles::egl] No config found!
WARN [wgpu_hal::gles::egl] EGL says it can present to the window but not natively
WARN [wgpu_hal::gles::adapter] Max vertex attribute stride unknown. Assuming it is 2048
WARN [wgpu_hal::gles::egl] No config found!
WARN [wgpu_hal::gles::egl] EGL says it can present to the window but not natively
```

Additionally, it would be nice if there was a configuration to disable the airplane mode button entirely.

MalpenZibo:
Closes #158

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,7 +127,7 @@ log_level = "warn"
 # Possible status bar outputs, values could be: All, Active, or a list of outputs
 # All: the status bar will be displayed on all the available outputs, example: outputs = "All"
 # active: the status bar will be displayed on the active output, example: outputs = "Active"
-# list of outputs: the status bar will be displayed on the outputs listed here, example: outputs = { targets = ["DP-1", "eDP-1"] }
+# list of outputs: the status bar will be displayed on the outputs listed here, example: outputs = { Targets = ["DP-1", "eDP-1"] }
 # if the outputs is not available the bar will be displayed in the active output
 outputs = "All"
 # Bar position, possible values Top | Bottom.


Libadoxon:
This fixed the issue of the airplane button doing nothing for me. One note tho, hardcoding the rfkill path leads to problems on NixOs do to it not having a FHS filesystem, it might be possible to just change it to `rfkill` or are there distros where `rfkill` is installed but not in PATH?

Libadoxon:
It might also be nice to throw an error or at least a warning instead of debug when the rfkill command fails, as at least for me, it is very hard to read the logs on debug level because of all the wgpu messages that are constantly being send

MalpenZibo:
> This fixed the issue of the airplane button doing nothing for me. One note tho, hardcoding the rfkill path leads to problems on NixOs do to it not having a FHS filesystem, it might be possible to just change it to `rfkill` or are there distros where `rfkill` is installed but not in PATH?

Hum, something similar has been done here https://github.com/MalpenZibo/ashell/pull/97 so could make sense to make the same changes here

MalpenZibo:
> It might also be nice to throw an error or at least a warning instead of debug when the rfkill command fails, as at least for me, it is very hard to read the logs on debug level because of all the wgpu messages that are constantly being send

Yes, the wgpu message could be removed with something like `log_level = "ashell=info"` that should set the level only for every ashell submodules

Libadoxon:
> Yes, the wgpu message could be removed with something like `log_level = "ashell=info"` that should set the level only for every ashell submodules

That makes the log a lot more readable, maybe consider adding a smalle note to the README

PR #183, opened 2025-06-27
Track GPU usage and temperature for nvidia gpus

CodedNil:
This PR adds tracking for nvidia gpus of usage and temps, it also renames the existing temperature to be CpuTemperature (I know it was mobo temp before but it was more or less the same as cpu, I also fixed that to properly read the temps for my system and presumably others with amd cpus as it didnt have "acpitz temp1".

It changes the icon for cpu usage and adds one for gpu usage, maybe not the best icon but couldn't find a more fitting one.

It uses nvidia-smi command line tool if available (afaik all nvidia gpu systems have this on linux by default with the drivers).
If you'd prefer it to use a proper rust crate I'm happy to get that working but all I could find added a lot of dependencies including C binding ones so needed gcc to build too etc, and I don't believe nvidia-smi has any performance worries at all, and it's a very stable api.
These seem like the best options for nvidia and amd that I could find. https://github.com/rust-nvml/nvml-wrapper https://github.com/Umio-Yasuno/libdrm-amdgpu-sys-rs


MalpenZibo:
Nice! I'm trying to find a way to have a "global" solution for every kind of GPU. Right now I'm watching how https://gitlab.com/mission-center-devs/mission-center retrieves this information, but it seems an "internal" crate not published yet.

PR #185, opened 2025-07-08
feat: config keys for custom shutdown, suspend, reboot and logout commands

GabMus:
As the title states. This was prompted by the command used for logout being incorrect when using `uwsm` for session management (it should be `uwsm stop` according to [hyprland's wiki](https://wiki.hypr.land/Configuring/Dispatchers/#list-of-dispatchers)).

This allows to set appropriate commands depending on the particular user's setup, and if left empty it would just use the defaults.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -206,9 +206,33 @@ impl Default for ClockModuleConfig {
     }
 }
 
+fn default_shutdown_cmd() -> String {
+    return "shutdown now".to_string();
+}
+
+fn default_suspend_cmd() -> String {
+    return "systemctl suspend".to_string();
+}
+
+fn default_reboot_cmd() -> String {
+    return "systemctl reboot".to_string();
+}
+
+fn default_logout_cmd() -> String {
+    return "loginctl kill-user $(whoami)".to_string();
+}
+
 #[derive(Deserialize, Default, Clone, Debug)]
 pub struct SettingsModuleConfig {
     pub lock_cmd: Option<String>,
+    #[serde(default = "default_shutdown_cmd")]
+    pub shutdown_cmd: String,
+    #[serde(default = "default_suspend_cmd")]
+    pub suspend_cmd: String,
+    #[serde(default = "default_reboot_cmd")]
+    pub reboot_cmd: String,
+    #[serde(default = "default_logout_cmd")]
+    pub logout_cmd: String,
     pub audio_sinks_more_cmd: Option<String>,
     pub audio_sources_more_cmd: Option<String>,
     pub wifi_more_cmd: Option<String>,
--- a/README.md
+++ b/README.md
@@ -295,6 +295,12 @@ alert = ".*notification"
 # without a value the related button will not appear
 # optional, default None
 lock_cmd = "hyprlock &"
+# commands used to respectively shutdown, suspend, reboot and logout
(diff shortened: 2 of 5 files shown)


MalpenZibo:
Very nice!

Is it possible to auto detect if ashell it's running in a uwsm env? 
I prefer to avoid the addition of that command in the configuration.

If it's not possible to understand if we're in a uwsm env then I prefer to have a generic 'uwsm' option (true/false) that change the behavior of that command.

I tried with chatgpt to understand if it's possible but I wasn't able to test it (I don't use uwsm).

Chatgpt says that there's a dbus way to check

```
use zbus::Connection;

async fn running_under_uwsm() -> zbus::Result<bool> {
    let connection = Connection::session().await?;
    let names = connection.list_names().await?;
    Ok(names.iter().any(|n| n.contains("uwsm")))
}
```

Maybe it could work 😅

GabMus:
The point is not to support uwsm specifically, but to support any compositor/DE without having to bake in every possible variant, so I would rather not bind to UWSM specifically.

Just for sake of argument, it appears UWSM sessions export these: env vars:

- `UWSM_FINALIZE_VARNAMES`
- `UWSM_WAIT_VARNAMES`

And "binds" this /run dir as well: `/run/user/$UID/uwsm/`.

MalpenZibo:
Ok, it make sense. 
I just want to make things as easier as possible.

In this case the only command that should be changed for uwsm is the logout command?

In other context the only problem could be with distro without systemd.

I will think about this. I would like to have something enough smart to understand the correct command for this cases. But if it's too complex than I think that we could go with these changes

GabMus:
> Ok, it make sense. I just want to make things as easier as possible.

Sure, but you might also want to make things easier for you, supporting uwsm explicitly is yet another integration you'd have to maintain. Plus, uwsm in particular is seeing use right now, but who knows if it's gonna stick around, or if some other thing will take its place. 

> In this case the only command that should be changed for uwsm is the logout command?

To my understanding, yes

> In other context the only problem could be with distro without systemd.

Good point. Also some user might want to spawn a custom confirmation window instead of shutting down/rebooting/logging out directly. This addition provides that option as well.

src/config.rs
```diff
@@ -209,6 +209,10 @@ impl Default for ClockModuleConfig {
 #[derive(Deserialize, Default, Clone, Debug)]
 pub struct SettingsModuleConfig {
     pub lock_cmd: Option<String>,
+    pub shutdown_cmd: Option<String>,
+    pub suspend_cmd: Option<String>,
+    pub reboot_cmd: Option<String>,
+    pub logout_cmd: Option<String>,
```
MalpenZibo: Here you could use a serde default value to avoid the `Option`
```
#[serde(default = "default_shutdown_cmd")]
```

With this change, you can avoid the `unwrap_or` later in the launcher utility

GabMus:
made your requested change, but the CI is failing. in particular clippy is failing but on a lot of code that's not mine. not sure if you usually ignore it or if there's something else broken

MalpenZibo:
I already know the clippy "issue". I will fix everything in a dedicated PR so no worries

GabMus:
Alright, feel free to merge when you prefer then

PR #207, opened 2025-08-05
Improve nixos experience

repomaa:
This streamlines the NixOS experience so that it will use the standard way of defining package outputs and defines a simple home manager module which will run ashell in a systemd user service and allow configuring it from within the home configuration.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 /target
 vendor
+/result
--- a/flake.nix
+++ b/flake.nix
@@ -4,75 +4,134 @@
   inputs = {
     crane.url = "github:ipetkov/crane";
     nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
-    flake-utils.url = "github:numtide/flake-utils";
+    flake-parts = {
+      url = "github:hercules-ci/flake-parts";
+      inputs.nixpkgs.follows = "nixpkgs";
+    };
     rust-overlay = {
       url = "github:oxalica/rust-overlay";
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -1,2 +1,3 @@
 /target
 vendor
+/result
```
repomaa, line 3: this is where nix flake build would put the result

flake.nix
```diff
@@ -4,75 +4,134 @@
...
+                Install.WantedBy = [ "graphical-session.target" ];
+                Unit = {
+                  After = [ "graphical-session.target" ];
+                  ConditionEnvironment = "WAYLAND_DISPLAY";
+                  Description = "Ready to go Wayland status bar for Hyprland";
+                  PartOf = [ "graphical-session.target" ];
+                };
+                Service = {
+                  ExecStart = lib.getExe cfg.package;
+                  X-Restart-Trigger = [ settings ];
```
repomaa, line 70: makes sure the service is restarted when the config changes.

MalpenZibo:
I'm not a nixOS expert, so I'm not able to do a proper review :D I will try a simple smoke test by installing ashell using Nix and this configuration on my Arch Linux environment. If it works, then we can proceed

repomaa:
@MalpenZibo never mind! home manager has a module by now :)

MalpenZibo:
Is this PR no longer needed, or is there still something useful in it?

repomaa:
No, not needed, let’s close it. 👍

Issue #176, opened 2025-06-19
Close popups with esc


Libadoxon:
Popups created from tray and settings menu currently only close when clicking on another window or on the button again. I think it would be a nice feature to close them using the esc button

MalpenZibo:
It sounds nice, but I have to find a way to listen to a global keypress. Probably, I can activate the keyboard interactivity when a menu is open to do that.

Right now, when a menu is open, you can continue to write in another application. For example, right now I'm writing here with the settings menu open. With this change, I think that this behavior will no longer be possible (but probably is not a big issue).

adamm-xyz:
Hey this project is really cool and this seems like an issue I think I can tackle. Learning iced right now so I'll give this a shot.

PR #208, opened 2025-08-07
Add support for closing all menus with Escape key

adamm-xyz:
- Introduced `CloseAllMenus` message to handle bulk menu closure
- Implemented `close_all_menus` in `Outputs`
- ESC key now triggers `CloseAllMenus` if any menu is open
- Changed keyboard interactivity to `OnDemand` for menus
- Implements #176

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -515,7 +515,7 @@ impl App {
             Subscription::batch(self.modules_subscriptions(&self.config.modules.center)),
             Subscription::batch(self.modules_subscriptions(&self.config.modules.right)),
             config::subscription(&self.config_path),
-            listen_with(|evt, _, _| match evt {
+            listen_with(move |evt, _, _| match evt {
                 iced::Event::PlatformSpecific(iced::event::PlatformSpecific::Wayland(
                     WaylandEvent::Output(event, wl_output),
                 )) => {
--- a/src/app.rs
+++ b/src/app.rs
@@ -527,7 +527,7 @@ impl App {
                     Some(Message::OutputEvent((event, wl_output)))
                 }
                 iced::Event::Keyboard(keyboard::Event::KeyPressed { key, .. }) => {
-                    debug!("Keyboard event received: {:?}", key);
+                    debug!("Keyboard event received: {key:?}");
                     if matches!(key, keyboard::Key::Named(keyboard::key::Named::Escape)) {
                         debug!("ESC key pressed, closing all menus");
                         Some(Message::CloseAllMenus)
--- a/src/app.rs
+++ b/src/app.rs
@@ -37,6 +37,7 @@ use iced::{
         wayland::{Event as WaylandEvent, OutputEvent},
     },
     gradient::Linear,
+    keyboard,
     widget::{Row, container},
     window::Id,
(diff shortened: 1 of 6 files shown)


MalpenZibo:
Because this feature require the menu to get the keyboard interactivity maybe it's better to add a config parameter to disable this feature in case someone prefer the current behavior.

The default behavior should be the one implemented in this PR and with the config flag to disable this feature the menu should not request the keyboard interactivity

adamm-xyz:
Okay. I'll go ahead and fix that clippy issue. Btw, regarding the config parameter to toggle keyboard interactivity: should I add that as part of this PR, or should we open a separate “feature request” issue to track and discuss it before implementation? I can make the change here if we want it included right away I suppose

MalpenZibo:
If you can add the config directly in this PR. 

If you can also update the website documentation and the changelog (no need to update the readme config documentation I will deprecate that section soon).

I will do a real review in the next few days!

src/outputs.rs
```diff
@@ -67,6 +68,7 @@ impl Outputs {
         style: AppearanceStyle,
         wl_output: Option<WlOutput>,
         position: Position,
+        config: &crate::config::Config,
```
MalpenZibo: Could we use the `menu_keyboard_focus` flag directly?

src/menu.rs
```diff
@@ -89,12 +98,20 @@ impl Menu {
         }
     }
 
-    pub fn request_keyboard<Message: 'static>(&self) -> Task<Message> {
-        set_keyboard_interactivity(self.id, KeyboardInteractivity::OnDemand)
+    pub fn request_keyboard<Message: 'static>(&self, config: &crate::config::Config) -> Task<Message> {
```
MalpenZibo: Same here, could we use the `menu_keyboard_focus` directly?

src/app.rs
```diff
@@ -505,6 +517,15 @@ impl App {
                     debug!("Wayland event: {event:?}");
                     Some(Message::OutputEvent((event, wl_output)))
                 }
+                iced::Event::Keyboard(keyboard::Event::KeyPressed { key, .. }) => {
+                    debug!("Keyboard event received: {key:?}");
+                    if matches!(key, keyboard::Key::Named(keyboard::key::Named::Escape)) {
+                        debug!("ESC key pressed, closing all menus");
+                        Some(Message::CloseAllMenus)
```
MalpenZibo, line 529: Here we could send the message only if some menus are open

MalpenZibo (changes requested):

src/app.rs
adamm-xyz: Doesn't close_all_menus in output.rs handle this already? Wouldn't it redundant to add it here again?

MalpenZibo: Yes, but if we avoid sending a Message in theory, we should be able to avoid an entire rerender

MalpenZibo:
I'm back. I will do the review as soon as I can

MalpenZibo (approved):
I will fix the remaining clippy, changelog and documentation stuff in another PR!

Issue #120, opened 2025-04-03
Change fontSize and/or bar height


mock1328:
In addition to specify `fontFamily`, I think it would be a good idea to change `fontSize`, as it is not seem to be possible atm. I tried tweaking `~/.config/gtk-4.0/settings.ini` and forcing settings with `dconf-editor`, but both seem to be ignored by `ashell`, and I still think it'd be better to change font size via `ashell`'s config.

Optionally, changing bar height sounds like a good idea too imo. I think these two are related.

MalpenZibo:
`ashell` doesn't use the GTK toolkit for this reason is useless to change the gtk settings.ini file.

Anyway, it's something that I would explore in the future, but right now, letting the user change the height of the bar and the font size could lead to many UI issues, and I also had to refactor some aspects of the codebase.

In the future will be some config to change the font dimension (right now I'm thinking of a `dimension = small | medium| large` option) but not in the next releases

Quackdoc:
is it possible to be possible to set `scale_factor`? Personally the bar would be almost perfect for me, but the size is just too small.

Quackdoc:
it seems like this will work. I made a quick POC by doing the below hack. HEIGHT needs to be increased or else the bar height will just remain the same and the stuff inside won't render at all. I don't have time to make a PR sorry.

```
diff --git a/src/app.rs b/src/app.rs
index 375faaa..214cd5e 100644
--- a/src/app.rs
+++ b/src/app.rs
@@ -128,6 +128,10 @@ impl App {
         }
     }
 
+    pub fn scale_factor(&self, _id: Id) -> f64 {
+        1.5
+    }
+
     pub fn update(&mut self, message: Message) -> Task<Message> {
         match message {
             Message::None => Task::none(),
diff --git a/src/main.rs b/src/main.rs
index 4a2f804..52bc7f6 100644
--- a/src/main.rs
+++ b/src/main.rs
@@ -22,7 +22,7 @@ mod style;
 mod utils;
 
 const ICON_FONT: &[u8] = include_bytes!("../assets/SymbolsNerdFont-Regular.ttf");
-const HEIGHT: u32 = 34;
+const HEIGHT: u32 = 51;
 
 fn get_log_spec(log_level: &str) -> LogSpecification {
     LogSpecification::env_or_parse(log_level).unwrap_or_else(|err| {
@@ -68,6 +68,7 @@ async fn main() -> iced::Result {
 
[...]

PR #209, opened 2025-08-09
fork scale

tutanra:
I have added a configuration variable **SCALE** in the appearance settings that uses floating-point values. The default value is 1.0, which corresponds to the original size. A value of 1.5 will represent a 50% increase, while a value of 2.0 will double the size. 

Additionally, I have enlarged the workspace selection sizes to improve accessibility.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -1,7 +1,7 @@
 use std::{collections::HashMap, f32::consts::PI, path::PathBuf};
 
 use crate::{
-    HEIGHT, centerbox,
+    get_height, centerbox,
     config::{self, AppearanceStyle, Config, Position},
     get_log_spec,
     menu::{MenuSize, MenuType, menu_wrapper},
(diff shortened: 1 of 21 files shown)


MalpenZibo:
Great work! It's something that I started with the Fe refactor here https://github.com/MalpenZibo/ashell/pull/175

I will try to include these changes in that PR as soon as I can!

MalpenZibo:
Maybe I found an easy solution to the problem, I need to patch the iced pop_os fork, but then everything works fine, see: https://github.com/MalpenZibo/ashell/issues/120#issuecomment-3175739340

MalpenZibo:
Could you try the latest main commit? I added a `scale_factor` that should resolve the problem

tutanra:
> Could you try the latest main commit? I added a `scale_factor` that should resolve the problem

 I'm on vacation away from home, I'll try it in a week. Anyway, anything will be fine as long as it's a little bit bigger.

MalpenZibo:
I will close this PR because the issue should be resolved with #211.

PR #214, opened 2025-08-17
Add support for virtual desktops

### LINKED ISSUE #210: Support for hyprland virtual desktops plugin
Love the bar and have been running it as my main system bar for a while now. Wanted to pitch a feature request to support the [hyprland-virtual-desktops plugin](https://github.com/levnikmyskin/hyprland-virtual-desktops) for the Workspaces module. The plugin combines workspaces for each monitor together to create a single "virtual desktop" so you can switch all monitors at once. I find it very useful.

It would be nice if one could show the grouped workspaces as one and swap to a new virtual desktop using the vdesk dispatcher on click. Could be a configuration option.

I will give it a stab during the weekend, it will probably not be too difficult even though I am not used to Rust.

emarforio:
Added support for the hyprland virtual desktops plugin. I am not a Rust programmer so it it should probably only be used as a reference since I have not put much thought into this but it works. You will probably have to consider the colors for workspaces better but this works well enough for me and I will try to keep an up to date fork available if you don't want this functionality in the main project.

Closes #210

### DIFF
--- a/website/docs/configuration/modules/workspaces.md
+++ b/website/docs/configuration/modules/workspaces.md
@@ -81,7 +81,21 @@ For example you can use [Nerd Fonts](https://www.nerdfonts.com/)
 ```toml
 [workspaces]
 workspace_names = ["一","二","三","四","五","六","七","八","九","十",]
+```
 
+## Virtual Desktop Plugin Support
+
+If you are using the Hyprland plugin [hyprland-virtual-desktops](https://github.com/levnikmyskin/hyprland-virtual-desktops)
+you probably want the workspaces to be grouped together instead
+of showing up individually. Use the configuration option
+`enable_virtual_desktops` to handle a virtual desktop as a single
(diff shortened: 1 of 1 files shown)


MalpenZibo:
Sorry, I haven’t had time to look at this PR until now.  
At the moment I don’t have a way to test the system (I don't have that hyprland plugin), but I saw that all the changes are behind a configuration flag, so nothing should break.  

If it works for you, I’ll just do a quick check to make sure there are no regressions without enabling this feature, and then I think it can be merged.  

I’d ask you, if you have time, to add a line in the changelog with this change, and maybe also in the site documentation.  

Thanks a lot!

MalpenZibo:
You no longer need to add the changelog file.

If you're able to resolve the conflict, we could proceed

emarforio:
Sorry for not responding earlier, life got in the way. I have merged the main branch now and moved the documentation to the website. Everything seems to work just fine still.

PR #232, opened 2025-09-13
feat: add option to remove idle inhibitor button in settings

sudo-Tiz:
Add a simple config option to disable idle inhibitor button in settings. The implementation follows the exact same pattern as the existing `remove_airplane_btn`


MalpenZibo (approved):
Great, I will fix clippy and add an entry to the changelog in a dedicated PR

PR #233, opened 2025-09-13
Feat: add custom button to settings panel

sudo-Tiz:
# Add Custom Buttons to Settings Panel

## Summary
Adds support for custom buttons in the settings panel, allowing users to execute commands or toggle applications directly from the status bar.

## Features
- **Launcher buttons**: Execute commands on click
- **Toggle buttons**: Track visual state using `status_command`
- Configurable via TOML with `[[settings.CustomButton]]` sections
- Support for custom tooltips

## Configuration Example
```toml
[[settings.CustomButton]]
name = "Virtual Keyboard"
icon = "⌨️"
command = "toggle-onscreen-keyboard.sh"
status_command = "pgrep -x onboard"
tooltip = "Toggle On-Screen Keyboard"
```

## Files Changed
- `src/config.rs`: Added `SettingsCustomButton` struct
- `src/modules/settings/mod.rs`: Implemented button rendering and command execution
- `README.md`: Added configuration examples
- `website/docs/configuration/modules/settings.md`: Updated documentation

## Current Limitations
- Icon support not implemented: The `icon` field is defined in the configuration structure but not yet used in the UI. All custom buttons currently display the same default icon (`Icons::AppLauncher`).
## Help Needed
Looking for assistance with implementing the icon feature. The configuration parsing is ready, but the UI rendering needs to be updated to use the configured icon instead of the hardcoded `Icons::AppLauncher`

## Breaking Changes
None - fully backward compatible.


MalpenZibo:
Generally, I like the idea of adding custom setting buttons.

There are two main things that I don't like:
- I'm not convinced about having a launcher button in the settings menu. I prefer to have only toggles.
- We should avoid calling the `check_toggle_status` inside the view function.

I will try to see if I find the time to fix the icon problem. In the meantime, I will convert the PR to a draft

sudo-Tiz:
Well, a launcher button is basically a toggle without state (state defined by status_command in my implementation).

MalpenZibo:
yep! that's the point. I prefer to have those buttons show some kind of state. For this reason, I'm not a big fan of the idea of having launcher buttons there

sudo-Tiz:
the "launcher mode" is just a fallback if the state command is not working/defined. Maybe we should not focus on this case

MalpenZibo:
Could make sense! Some conflicts should be resolved. In case of a problem, let me know

sudo-Tiz:
I fixed the conflict. i could use some help on the icon part, I am not sure about the best way to implement it. Also, please take a look at this [branch](https://github.com/sudo-Tiz/ashell/tree/feature/settings_CustomButton_async) where the status_command run asynchronously

MalpenZibo:
For the Icon, the problem is that currently the `Icon` enum implements Copy, so it's not easy to add an enum value like `Dynamic(String)` for a custom icon.

We have an `icon_raw` function that accepts a String, but the `quick_setting_button` wants an `Icons`.

We could rename the `Icons` enum to something like `StaticIcons` that implements `Copy` and is easy to use without placing `clone` everywhere. Then we could add an `Icons` enum like this:

```rust
enum Icons {
  Static(StaticIcons),
  Dynamic(String)
}
```

Then we could change the `quick_setting_button` to accept a `impl Into<Icons>` and implement `Into<Icona>` to the new `StaticIcons` enum.

Maybe we could also rename everything in the singular form eheh

```rust
pub enum Icon {
  Static(StaticIcon),
  Dynamic(String)
}

pub enum StaticIcon {
      None,
    AppLauncher,
    Clipboard,
    Refresh,
    NoUpdatesAvailable,
    UpdatesAvailable,
    ...
}
```

MalpenZibo:
I started watching the `async` version! It's better.

In the next few days, I will try to test it.

MalpenZibo:
[Here](https://github.com/MalpenZibo/ashell/pull/269) I made a refactor of the `Icon`.

Now the `Icon` is a `trait` that means that we have `StaticIcon`, the one defined in the codebase, and `DynamicIcon`. In your case, you could use the `DynamicIcon`. See [here](https://github.com/MalpenZibo/ashell/pull/269/files#diff-17846255c47b11c10a4f6e0a554d4373700d243fd2df0c54dd4a7998f46f9eccR587)

sudo-Tiz:
async version with icon support

src/modules/settings/mod.rs
```diff
@@ -260,15 +274,103 @@ impl Settings {
...
+                                (button.name, status)
+                            });
+                            handles.push(handle);
+                        }
+
+                        // Wait for all threads to complete
+                        handles.into_iter().filter_map(|h| h.join().ok()).collect()
+                    },
+                    Message::CustomButtonsStatus,
+                );
```
MalpenZibo: With this, the `wait-timeout` crate is not needed

```suggestion
                let buttons = self.custom_buttons.clone();

                let custom_buttons_task = if buttons.is_empty() {
                    Task::none()
                } else {
                    Task::perform(
                        async move {
                            let futures = buttons.into_iter().map(|button| async move {
                                if let Some(cmd) = button.status_command {
                                    let result = timeout(Duration::from_secs(1), async {
                                        let output = Command::new("bash")
                                            .arg("-c")
                                            .arg(cmd)
                                            .status()
                                            .await?;

                                        Ok::<_, std::io::Error>(output.success())
                                    })
                                    .await;

                                    match result {
                                        Ok(Ok(output)) => {
                                            debug!(
                                                "Custom button '{}' status_command executed with result: {}",
                                                button.name, output
                                            );
                                            (button.name, Some(output))
                                        },
                                        Ok(Err(e)) => {
                                            error!(
                                                "Failed to spawn status_command for custom button '{}': {}",
                                                button.name, e
                                            );
                                            (button.name, None)
                                        },
                                        Err(_) => {
                                            error!(
                                                "Custom button '{}' status_command timed out after 1000ms",
                                                button.name
                                            );
                                            (button.name, None)
                                        },
                                    }
                                } else {
                                    (button.name, Some(false))
                                }
                            });

                            join_all(futures).await
                        },
                        Message::CustomButtonsStatus,
                    )
                };

```

sudo-Tiz: done

sudo-Tiz:
Replaced thread-based status command execution with async/await pattern using tokio

PR #240, opened 2025-09-25
feat(workspaces): add support for custom workspace names

adamm-xyz:
This PR introduces support for user-defined workspace names in the Workspaces module. Users can now configure a list of names for their workspaces instead of relying solely on numeric IDs. Currently, workspaces are displayed using numeric IDs only, which can be less intuitive for users who want different numerals (eg roman numerals, chinese numbers). This change improves usability and customization without breaking existing configurations.

Example config addition below

`[workspaces]

workspace_names = ["一","二","三","四","五","六","七","八","九","十"]
`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,7 @@ jobs:
     runs-on: ubuntu-24.04
 
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
 
       - name: Install dependencies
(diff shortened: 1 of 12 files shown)


MalpenZibo:
Great!!! Can I ask you to add a changelog entry and some documentation on the website about this configuration? `website/docs/configuration/modules/workspaces.md`

MalpenZibo (approved):
🚀

PR #254, opened 2025-10-03
feat: add temperature sensor configuration option

sudo-Tiz:
This PR intend to add an option to configure the temperature sensor used by system info module

configure like so:
```toml
[system.temperature]
sensor = "coretemp Package id 0"
```


MalpenZibo (approved):
Perfect! 🚀  

Can you add an entry in the changelog?

PR #278, opened 2025-10-15
feat: add hibernate option to power settings

sudo-Tiz:
## Add Hibernate Option to Power Menu

Adds a hibernate button to the quick settings power menu, positioned between Suspend and Reboot.

### Changes
- New `hibernate_cmd` configuration option (defaults to `systemctl hibernate`)
- Hibernate button in power menu with moon icon
- Updated documentation

### Configuration
```toml
[settings]
hibernate_cmd = "systemctl hibernate"  # optional, this is the default
```

### Note
Currently using moon icon (`\u{f0f61}`) for hibernate - open to suggestions for a better icon choice.

### DIFF
--- a/.github/workflows/generate-icons.yml
+++ b/.github/workflows/generate-icons.yml
@@ -14,6 +14,8 @@ permissions:
 jobs:
   generate-icons:
     runs-on: ubuntu-latest
+    # Only run on push events to avoid permission issues on PRs from forks
+    if: github.event_name == 'push'
     steps:
       - uses: actions/checkout@v4
         with:


MalpenZibo (changes requested):
The code it's ok!

There are two things that I would like to change:
- The hibernation process is not possible on every system (for example, I'm using zram as a swap "partition"). For this reason, I'd like to treat hibernation differently. I prefer to have it as an opt-in option, so instead of having a default value, I prefer to have a default to `Option::None` and avoid displaying the button.
- For the Icon, I don't have too many ideas 😅 maybe the half moon is too similar to a semicircle. We could use a snowflake? maybe f0717? 

Also, remember that now I don't package by default the whole nerdfont icon, but only a subset. If you add a new icon, you have to run the generate script. If you have trouble because you had to install some programs to run the script, let me know. We could open an issue after merging this PR to remember to update the font. 

Maybe in the future, I could add a GH action to do this process automatically.

sudo-Tiz:
Icon updated. Generate script ran and a GitHub Action now automates this process when icons.rs is modified.

Regarding the opt-in approach: this could be generalized. For example, I don't need the logout button (single user with pam_autologin). A better solution would be letting users define an array of {icon, label, command} entries. This way users could define as many commands as needed (for exemple adding "hyprctl kill" or removing unwanted buttons entirely).

sudo-Tiz:
Since it's not possible for gh action to push on a fork for security reason, i edited the action so it run only when the maintainer (you) merge the code

sudo-Tiz:
If you’re okay with it, we could merge this PR now since having the Hibernate button isn’t that bothersome. Then, we can rework this section to allow the user to define their preferences in the config.

MalpenZibo (approved):
🚀

PR #279, opened 2025-10-16
Feature: Add Tempo module

### LINKED ISSUE #79: Add calendar
Add a calendar that will be displayed when you click on the clock

MalpenZibo:
From the Italian word `tempo`, which can mean both `weather` and `time`, say hello to the Tempo module, which combines a calendar and a weather widget.

Closes #79 

Some teaser screenshot
image

### DIFF
--- a/src/modules/tempo.rs
+++ b/src/modules/tempo.rs
@@ -95,11 +95,12 @@ impl Tempo {
         self.weather_data.as_ref().map(|data| {
             row!(
                 weather_icon(data.current.weather_code, data.current.is_day > 0)
-                    .width(Length::Fixed(15.)),
+                    .width(Length::Fixed(theme.font_size.sm as f32)),
                 text(format!("{}°C", data.current.temperature_2m))
                     .align_y(Vertical::Center)
                     .size(theme.font_size.sm)
             )
+            .align_y(Vertical::Center)
             .spacing(theme.space.xxs)
             .into()
         })
@@ -585,13 +586,17 @@ impl Tempo {
 }
 
 async fn fetch_location(location: WeatherLocation) -> anyhow::Result<Location> {
+    let client = reqwest::Client::builder()
+        .timeout(Duration::from_secs(10))
+        .build()?;
+
     match location {
         WeatherLocation::City(city) => {
             let url = format!(
                 "https://geocoding-api.open-meteo.com/v1/search?name={}&count=1&language=en&format=json",
                 city
             );
-            let response = reqwest::get(&url).await?;
+            let response = client.get(&url).send().await?;
             let raw_data = response.text().await?;
 
             let data: GeoLocations = serde_json::from_str(&raw_data)?;
@@ -605,7 +610,7 @@ async fn fetch_location(location: WeatherLocation) -> anyhow::Result<Location> {
         WeatherLocation::Current => {
             let find_location = "http://ip-api.com/json/";
 
-            let response = reqwest::get(find_location).await?;
+            let response = client.get(find_location).send().await?;
             let raw_data = response.text().await?;
 
             let data: IpLocation = serde_json::from_str(&raw_data)?;
@@ -672,7 +677,11 @@ pub struct Location {
 }
 
 async fn fetch_weather_data(lat: f32, lon: f32) -> anyhow::Result<WeatherData> {
-    let response = reqwest::get(format!(
+    let client = reqwest::Client::builder()
+        .timeout(Duration::from_secs(10))
+        .build()?;
+
+    let response = client.get(format!(
         "https://api.open-meteo.com/v1/forecast?\
         latitude={}&longitude={}\
         &current=weather_code,apparent_temperature,relative_humidity_2m,temperature_2m,is_day,wind_speed_10m,wind_direction_10m\
@@ -680,7 +689,7 @@ async fn fetch_weather_data(lat: f32, lon: f32) -> anyhow::Result<WeatherData> {
         &daily=weather_code,temperature_2m_max,temperature_2m_min,wind_speed_10m_max,wind_direction_10m_dominant\
         &forecast_days=7", 
         lat, lon
-    )).await?;
+    )).send().await?;
     let raw_data = response.text().await?;
 
     let data: WeatherData = serde_json::from_str(&raw_data)?;


romanstingler:
are your 2 PRs ready for review??

romanstingler:
Hey I had a quick look at that. 

```
[tempo]
clock_format = "%a %d %b %R"
weather_location = { City = "Klagenfurt" }
```

I got the warning 
`WARN [ashell::modules::tempo] Failed to fetch location: missing field `lat` at line 1 column 375`

then I looked at the code and made a call

`curl -s 'https://geocoding-api.open-meteo.com/v1/search?name=Klagenfurt&count=1&language=en&format=json'`
```
{
  "results": [
    {
      "id": 2774326,
      "name": "Klagenfurt",
      "latitude": 46.62472,
      "longitude": 14.30528,
      "elevation": 450,
      "feature_code": "PPLA",
      "country_code": "AT",
      "admin1_id": 2774686,
      "admin2_id": 2774321,
      "admin3_id": 7871497,
      "timezone": "Europe/Vienna",
      "population": 100316,
      "country_id": 2782113,
      "country": "Austria",
      "admin1": "Carinthia",
      "admin2": "Klagenfurt am Wörthersee",
      "admin3": "Klagenfurt"
    }
  ],
  "generationtime_ms": 0.47779083
}
```

Which led to the conclusion that Tempo couldn’t show weather because the Open‑Meteo geocoding response fields (latitude, longitude, etc.) were being deserialized straight into our Location struct, which expects `lat`, `lon`, and `region_name`. 
That mismatch produced the “missing field lat”. 
I added a small adapter([tempo.patch](https://github.com/user-attachments/files/24798939/tempo.patch)): we now deserialize into an internal GeoLocation type and map it into the existing Location struct, falling back to admin1 or country for the region name.

Open‑Meteo’s geocoding response doesn’t provide a `region_name` field, but our Location struct expects one so the UI can show “City, Region” (e.g., “Klagenfurt, Carinthia”). 
The API does return admin1 (first-level administrative division) and country, so in fetch_location we map those optional fields into `region_name`, falling back to the city name if neither is present. 
That’s why we deserialize those fields—just to populate the region label shown next to the city.


Other than that, the code looks good. 
The only complaint that I still have is that the temperature is not aligned with the clock.



I played around a bit basically on my 4k screen 
`Space::with_height(Length::Fixed(2.0)),` makes it fine (I know that this is not the solution, but I have no idea, because the other icons like in Settings, they are wrapped in containers.

just prototyping stuff 
```
    pub fn weather_indicator(&'_ self, theme: &AshellTheme) -> Option<Element<'_, Message>> {
        self.weather_data.as_ref().map(|data| {
            Row::new()
                .push(
                    container(column!(
                        weather_icon(data.current.weather_code, data.current.is_day > 0)
                            .width(Length::Shrink)
                            .height(theme.font_size.md),
                        Space::with_height(Length::Fixed(2.0)),
                    ))
                    .align_y(Vertical::Center),
                )
                .push(
                    container(
                        text(format!("{}°C", data.current.temperature_2m)).size(theme.font_size.sm),
                    )
                    .align_y(Vertical::Center),
                )
                .spacing(theme.space.xxs)
                .align_y(Vertical::Center)
                .into()
        })
    }
```

PS: I added some docs 
https://github.com/MalpenZibo/ashell/pull/448


I guess you want to replace the clock module with the Tempo one, 
we should discuss that in a separate issue, but I would suggest to leave the Clock config as it is remove the clock code, so that we don't use duplicated code and then in the settings basically disable weather, because there are people like me that don't need weather or don't want constant http calls in a bar.

MalpenZibo:
@romanstingler I resolved all the issue (in theory :sweat_smile:)

romanstingler:
> @romanstingler I resolved all the issue (in theory 😅)

tomorrow I will have some time, I will test and review again.  (damn, can't even do as I say :smile:  )
hopefully we will get that merged, 

I still have to finish some fixes for the tray module.

romanstingler:
I just had quick look 
One thing is the icon in the bar 
your code gives me (bottom the same as text but height is bigger than the text)
image

If you intentionally want to have a bigger icon then it should be vertically centered

If you want to have it the same size as the other text/icons then you could try 

```
diff --git a/src/modules/tempo.rs b/src/modules/tempo.rs
index a056e83..31bb2fe 100644
--- a/src/modules/tempo.rs
+++ b/src/modules/tempo.rs
@@ -95,12 +95,14 @@ impl Tempo {
         self.weather_data.as_ref().map(|data| {
             row!(
                 weather_icon(data.current.weather_code, data.current.is_day > 0)
-                    .width(Length::Fixed(15.)),
+                    .width(Length::Fixed(theme.font_size.sm as f32))
+                    .height(Length::Fixed(theme.font_size.sm as f32)),
                 text(format!("{}°C", data.current.temperature_2m))
                     .align_y(Vertical::Center)
                     .size(theme.font_size.sm)
             )
             .spacing(theme.space.xxs)
+            .align_y(Vertical::Center)
             .into()
         })
     }
```

which gives you
image

And the second one is probably not a real issue, because I guess the server will have a 10-30s timeout,
but we could add a client-side timeout just in case (I have 10s in my test)

```
diff --git a/src/modules/tempo.rs b/src/modules/tempo.rs
index a056e83..3136cd7 100644
--- a/src/modules/tempo.rs
+++ b/src/modules/tempo.rs
@@ -585,13 +585,17 @@ impl Tempo {
 }
 
 async fn fetch_location(location: WeatherLocation) -> anyhow::Result<Location> {
+    let client = reqwest::Client::builder()
+        .timeout(Duration::from_secs(10))
+        .build()?;
+
     match location {
         WeatherLocation::City(city) => {
             let url = format!(
                 "https://geocoding-api.open-meteo.com/v1/search?name={}&count=1&language=en&format=json",
                 city
             );
-            let response = reqwest::get(&url).await?;
+            let response = client.get(&url).send().await?;
             let raw_data = response.text().await?;
 
             let data: GeoLocations = serde_json::from_str(&raw_data)?;
@@ -605,7 +609,7 @@ async fn fetch_location(location: WeatherLocation) -> anyhow::Result<Location> {
         WeatherLocation::Current => {
             let find_location = "http://ip-api.com/json/";
 
-            let response = reqwest::get(find_location).await?;
+            let response = client.get(find_location).send().await?;
             let raw_data = response.text().await?;
 
             let data: IpLocation = serde_json::from_str(&raw_data)?;
@@ -672,7 +676,11 @@ pub struct Location {
 }
 
 async fn fetch_weather_data(lat: f32, lon: f32) -> anyhow::Result<WeatherData> {
-    let response = reqwest::get(format!(
+    let client = reqwest::Client::builder()
+        .timeout(Duration::from_secs(10))
+        .build()?;
+
+    let response = client.get(format!(
         "https://api.open-meteo.com/v1/forecast?\
         latitude={}&longitude={}\
         &current=weather_code,apparent_temperature,relative_humidity_2m,temperature_2m,is_day,wind_speed_10m,wind_direction_10m\
@@ -680,7 +688,7 @@ async fn fetch_weather_data(lat: f32, lon: f32) -> anyhow::Result<WeatherData> {
         &daily=weather_code,temperature_2m_max,temperature_2m_min,wind_speed_10m_max,wind_direction_10m_dominant\
         &forecast_days=7", 
         lat, lon
-    )).await?;
+    )).send().await?;
     let raw_data = response.text().await?;
 
     let data: WeatherData = serde_json::from_str(&raw_data)?;
```
 
 Other than that, I can only complain about the bad weather :)

PR #288, opened 2025-10-18
feat(bluetooth): change indicator icon on connected status

sudo-Tiz:
Show Bluetooth "connected" icon in bar indicators when at least one device is connected


MalpenZibo:
Nice!

Only one thing.

I prefer to use the `md` icons if it's possible. So in this case, `f00b1`. Even if it has more of the meaning of "I'm connecting to something". 
Otherwise, we could use the same icon but with a different color. 
- Normal text -> no device connected, 
- success -> at least one device connected. 
Or both, color and a different icon. 

The only thing I prefer not to use `f293` because it's quite different in style from the rest of the icons.

sudo-Tiz:
Okay, I made the changes from the GitHub interface. The icon workflow should regenerate the TTF file once this branch is merged into the main branch.  

There's also the f00b0 icon, which represents "playing music via Bluetooth." I think it would be great to add this one. This could be included in a future PR.    

PS: When no Bluetooth is connected, I prefer not to display the Bluetooth icon (f00b2)

MalpenZibo:
I don't get the PS. You mean that you prefer to only show the Bluetooth icon if some devices are connected? In this way, we could only use the "base" Bluetooth icon.
In case I like the idea, I'm not a big fan of having a Bluetooth icon that only shows me that the Bluetooth is connected (in reality, I don't use the Bluetooth icon at all :sweat_smile:)

sudo-Tiz:
Sorry, I wanted to say :
PS: When Bluetooth is disabled, I prefer not to display the Bluetooth icon (f00b2)

No icon : bt disable
f00af icon: bt enable
f00b1 icon: 1+ device(s) connected

PR #294, opened 2025-10-21
fix(config) Make Default and Deserialize more in sync

### LINKED ISSUE #291: [BUG] Shutdown/reboot/suspend/logout buttons do nothing
I'm not getting any abnormal log output or anything, even with `RUST_LOG=TRACE`. I'm just having the problem that nothing happens when I push any of the power related buttons. This is v0.6.0 off of nixpkgs unstable.

Siprj:
Fix the need for an empty [settings] section in the `config.toml` to get default values for the whole section.

This was caused by inconsistency in default instance and serde instances for the `SettingsModuleConfig` data structure.

Closes #291

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -21,46 +21,33 @@ use tokio::time::sleep;
 pub const DEFAULT_CONFIG_FILE_PATH: &str = "~/.config/ashell/config.toml";
 
 #[derive(Deserialize, Clone, Debug)]
+#[serde(default)]
 pub struct Config {
-    #[serde(default = "default_log_level")]
     pub log_level: String,
-    #[serde(default)]
     pub position: Position,
-    #[serde(default)]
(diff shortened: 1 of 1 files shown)


Siprj:
Actually, when I'm thinking about it. Maybe it is worth rewriting all the types in the config module to have consistent default implementations to the serde derivations. Also, it would help with reducing the amount of small functions that just return a single value.

If there is interest in this change. I'm willing to do it. The question then would be if it should be part of this PR or part of some refactoring PR.

MalpenZibo:
It's a good idea :ok_hand: . We can do it in this PR or in different PRs. 

From my point of view, there's no difference.

Siprj:
Ok, I added the discussed refactoring. I think it is ready to be reviewed.

Issue #156, opened 2025-05-24
Indicator for which workspace is selected on other screen(s)


niveK77pur:
Hello, first off, thank you for this amazing tool! It just works so well, has everything you'd need from a bar, and looks super gorgeous at that :D

Regarding the workspace indicator, I do have a small complaint and feature request:

![Image](https://github.com/user-attachments/assets/586f55b3-39d3-48cc-b88c-ca3c6b4c5ce0)

In the screenshot, you can see that I currently have workspace 2 selected on my second monitor. However, if I want to move back to my first monitor, I wouldn't know whether I need to go to workspace 8 or 9.

Agreeably, in the moment I would somehow just intuitively know most likely, but it would be awesome to have another small indication somehow that Workspace 9 is currently selected on my first monitor.

To follow the currentl style, maybe the selected workspaces on unselected monitors could be ever so slightly more oval, but not quite as oval as the actual selected workspace; if that makes any sense. Otherwise, a colored border around the workspace?

MalpenZibo:
It makes sense to have the "used" workspace on the other monitor slightly oval!

I will try to add this feature in the future

romanstingler:
@niveK77pur I guess this one is resolved for you or?

niveK77pur:
This is indeed resolved now thanks a lot for the work!

PR #306, opened 2025-11-05
Feature: multi-monitor visible indicator

EdgesFTW:
Adds additional information about the inactive but visible workspaces. Currently, just creates a shape between active and inactive for those workspaces.

I did notice that there is also some strange behavior that may or may not be a bug. The workspace that is "active" will never return a special workspace. The current workaround for that leads to having two "active" workspaces when a special workspace is used.

Addresses #156.

### DIFF
--- a/src/modules/workspaces.rs
+++ b/src/modules/workspaces.rs
@@ -22,13 +22,20 @@ use std::{
     sync::{Arc, RwLock},
 };
 
+#[derive(PartialEq, Eq, Debug, Clone)]
+pub enum Displayed {
+    Active,
+    Visible,
+    Hidden,
+}
+
 #[derive(Debug, Clone)]
 pub struct Workspace {
     pub id: i32,
     pub name: String,
     pub monitor_id: Option<i128>,
     pub monitor: String,
-    pub active: bool,
+    pub displayed: Displayed,
     pub windows: u16,
 }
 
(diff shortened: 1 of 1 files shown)


MalpenZibo:
Thanks! I don't understand the possible bug. Could you add some screenshots of these changes and maybe the bug?

EdgesFTW:
Of course!

## Changes

Here, `workspace 2` is active on `monitor 1` and `workspace 11` is visible on `monitor 2` but not active. This is what ashell currently looks like.
image

Here is what this PR currently looks like.
image

## Possible bug

What I believe may be a bug, but also could be intentional, is when a special workspace is active on top of a normal workspace. Here I started with `workspace 1` active. I then switched to the special `workspace Email`
image
Due to a workaround where special workspaces are not considered active, both the special `workspace Email` and `workspace 1` are active at the same time. This is shown in ashell with the additional padding around those workspaces.

I believe this is a bug because hyprland only exposes one active workspace at a time, but due to this workaround we have two active at once. It could however still be useful to show what normal workspace is under the current special workspace.

MalpenZibo:
ok! Now it's clear. It seems like a bug, but it could be useful. I don't know how to treat this behavior :sweat_smile: 

Anyway, if you can, maybe open a dedicated issue about this with this screenshot.

PR #307, opened 2025-11-05
Improve vpn button

### LINKED ISSUE #47: [Suggestion] Improve the VPN button
The discussion started in #44 

it would be great if the last active vpn turned on/off when you click the vpn button. If the latter is missing, the list should open. To open the list of all VPNs, you can add a menu button as done with wifi.

Yes, for this it is necessary to store the last active vpn somehow, but I really hope that this will not be a big problem.

matuck:
Added wireguard vpn's to the vpn button.  Made it so that button will toggle off the first connected VPN if clicked and an arrow to select individual vpn's.  Fixes #47

### DIFF
--- a/src/modules/settings/network.rs
+++ b/src/modules/settings/network.rs
@@ -384,7 +384,7 @@ impl NetworkSettings {
                             "Vpn".to_string(),
                             subtitle,
                             !actives.is_empty(),
-                            if actives.len() >= 1
+                            if !actives.is_empty()
                                 && let Some(first) = actives.first()
                             {
                                 Message::ToggleVpn((*first).clone())


src/modules/settings/network.rs
```diff
@@ -351,6 +351,26 @@ impl NetworkSettings {
...
+                })
+                .into_iter()
+                .filter_map(|conn| {
+                    if activeconnection.is_some_and(|(name, _)| *name == conn.name) {
+                        Some(conn)
+                    } else {
+                        None
+                    }
+                })
+                .next();
```
MalpenZibo:
```rust
        self.service.as_ref().and_then(|service| {
            service
                .known_connections
                .iter()
                .any(|c| matches!(c, KnownConnection::Vpn { .. }))
                .then(|| {
                    let mut known_vpn = service.known_connections.iter().filter_map(|c| match c {
                        KnownConnection::Vpn(c) => Some(c),
                        _ => None,
                    });
                    let actives = service
                        .active_connections
                        .iter()
                        .filter_map(|c| match c {
                            ActiveConnectionInfo::Vpn { name, .. } => {
                                known_vpn.find(|v| v.name == *name)
                            }
                            _ => None,
                        })
                        .collect::<Vec<_>>();

                    let subtitle = if actives.len() > 1 {
                        Some(format!("{} VPNs Connected", actives.len()))
                    } else {
                        actives.first().map(|c| c.name.clone())
                    };

                    (
                        quick_setting_button(
                            theme,
                            StaticIcon::Vpn,
                            "Vpn".to_string(),
                            subtitle,
                            !actives.is_empty(),
                            if actives.len() > 1
                                && let Some(first) = actives.first()
                            {
                                Message::ToggleVpn((*first).clone())
                            } else {
                                Message::ToggleVPNMenu
                            },
                            if !actives.is_empty() {
                                Some((SubMenu::Vpn, sub_menu, Message::ToggleVPNMenu))
                            } else {
                                None
                            },
                        ),
                        sub_menu
                            .filter(|menu_type| *menu_type == SubMenu::Vpn)
                            .map(|_| {
                                Self::vpn_menu(
                                    service,
                                    id,
                                    theme,
                                    self.config.vpn_more_cmd.is_some(),
                                )
                            }),
                    )
                })
        })
```

I rewrote this part. More or less is the same thing; I only adjust the quick button subtitle in cases where more than a single VPN is connected.

I didn't test it!

MalpenZibo (changes requested):

src/modules/settings/network.rs
matuck: Compiles and mostly works.  There is an issue when clicking on the the button doesn't disconnect the 2nd vpn. its doing a toggle so I will work on that

matuck: Small bug was fixed by changing line actives.len() > 1 to actives.len() >= 1

src/modules/settings/network.rs
```diff
@@ -356,18 +356,46 @@ impl NetworkSettings {
...
-                            None,
-                            service
-                                .active_connections
-                                .iter()
-                                .any(|c| matches!(c, ActiveConnectionInfo::Vpn { .. })),
-                            Message::ToggleVPNMenu,
-                            None,
+                            subtitle,
+                            !actives.is_empty(),
+                            if actives.len() >= 1
```
MalpenZibo: Clippy "the boss" said to use `is_empty` 😅 

```suggestion
                            if !actives.is_empty()
```

MalpenZibo (changes requested):

src/modules/settings/network.rs
matuck: does make the code a little more readable.  I have commited the changes and everything still works as expected.

PR #312, opened 2025-11-10
Fuzzy search output names from config

CodedNil:
My system and I can only assume plenty of others, has dynamic names for outputs. My second monitor some boots shows up as HDMI-A-2, sometimes as HDMI-A-3, on rare occasions it can even show up as HDMI-A-1 and switch with my primary monitor.
This PR makes it so you can set the config name to fuzzy search from the outputs description for example "PNP(GWD) - ARZOPA - HDMI-A-2"
So if my config has the output as "ARZOPA" it will always only match to my second monitor.
This shouldn't change anything for existing setups as it will still also match to the output name.
The PR also simplifies outputs to not use an Option<String> as the fallback ones which set that to None already have the WlOutput as None so it makes sense imo to always supply a name to simplify the code.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -362,7 +362,7 @@ impl App {
                     info!("Output created: {info:?}");
                     let name = info
                         .as_ref()
-                        .and_then(|info| info.name.as_deref())
+                        .and_then(|info| info.description.as_deref())
                         .unwrap_or("");
 
                     self.outputs.add(
--- a/src/outputs.rs
+++ b/src/outputs.rs
@@ -27,7 +27,7 @@ struct ShellInfo {
 }
 
 #[derive(Debug, Clone)]
-pub struct Outputs(Vec<(Option<String>, Option<ShellInfo>, Option<WlOutput>)>);
+pub struct Outputs(Vec<(String, Option<ShellInfo>, Option<WlOutput>)>);
 
 pub enum HasOutput<'a> {
     Main,
@@ -44,7 +44,7 @@ impl Outputs {
 
         (
             Self(vec![(
-                None,
+                "Fallback".to_string(),
                 Some(ShellInfo {
                     id,
                     menu: Menu::new(menu_id),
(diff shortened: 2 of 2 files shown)


src/app.rs
```diff
@@ -362,7 +362,7 @@ impl App {
                     info!("Output created: {info:?}");
                     let name = info
                         .as_ref()
-                        .and_then(|info| info.name.as_deref())
+                        .and_then(|info| info.description.as_deref())
```
MalpenZibo, line 365: This shouldn't introduce a breaking change because the description always includes the `name`? Are we 100% sure that this always happens?

MalpenZibo: here https://wayland.freedesktop.org/docs/html/apa.html#protocol-spec-wl_output says that

```
Many compositors can produce human-readable descriptions of their outputs. The client may wish to know this description as well, e.g. for output selection purposes.

The description is a UTF-8 string with no convention defined for its contents. The description is not guaranteed to be unique among all wl_output globals. Examples might include 'Foocorp 11" Display' or 'Virtual X11 output via :1'.

The description event is sent after binding the output object and whenever the description changes. The description is optional, and may not be sent at all.

The description event will be followed by a done event. 
```

But the same is true also for the `name` field, so it should be ok. Let me think about it. Apart from that, the PR seems ok :rocket:

CodedNil: I could make it use name and model merged, instead of description, just to guarantee it contains name?

MalpenZibo: For now let's avoid this. In theory ashell only support Hyprland and it seems that on Hyprland the description contains the name so tomorrow I'll test again everything and then we're ready to merge this change

CodedNil: I'm using niri and for me the description is the same as you describe with hyprland, and niri is using smithay so this applies to cosmic too, I've also tested it and found same behaviour on gnome so I think it can be expected to work this way across all wayland compositors. Though as you say it's not strictly defined by any of them.

PR #315, opened 2025-11-14
Make system_info network selection deterministic

kylesferrazza:
Adds support to specify a network interface for the SystemInfo module.

perhaps later this could be updated to be able to support something like

```toml
[system_info]
interface_name = "eth0"
indicators = [ { IpAddress = { interface_name = "eth0" } } ]
```

so that you can do multiple indicators, each for a different network interface

Closes https://github.com/MalpenZibo/ashell/issues/321

### DIFF
--- a/src/modules/system_info.rs
+++ b/src/modules/system_info.rs
@@ -71,27 +71,54 @@ fn get_system_info(
 
     let elapsed = last_check.map(|v| v.elapsed().as_secs());
 
-    let network = networks.iter().fold(
-        (None, 0, 0),
-        |(first_ip, total_received, total_transmitted), (_, data)| {
-            let ip = first_ip.or_else(|| {
-                data.ip_networks()
-                    .iter()
-                    .sorted_by(|a, b| a.addr.cmp(&b.addr))
-                    .next()
-                    .map(|ip| ip.addr)
-            });
+    let network = networks
+        .iter()
+        .filter(|(name, _)| {
+            name.contains("en")
+                || name.contains("eth")
+                || name.contains("wl")
+                || name.contains("wlan")
+        })
+        .sorted_by_key(|(name, _)| {
+            if name.contains("en") {
+                return 0;
+            }
 
-            let received = data.received();
-            let transmitted = data.transmitted();
+            if name.contains("eth") {
+                return 1;
+            }
 
-            (
-                first_ip.or(ip),
-                total_received + received,
-                total_transmitted + transmitted,
-            )
-        },
-    );
+            if name.contains("wl") {
+                return 2;
+            }
+
+            if name.contains("wlan") {
+                return 3;
+            }
+
+            99
+        })
+        .fold(
+            (None, 0, 0),
+            |(first_ip, total_received, total_transmitted), (_, data)| {
+                let ip = first_ip.or_else(|| {
+                    data.ip_networks()
+                        .iter()
+                        .sorted_by(|a, b| a.addr.cmp(&b.addr))
+                        .next()
+                        .map(|ip| ip.addr)
+                });
+
+                let received = data.received();
+                let transmitted = data.transmitted();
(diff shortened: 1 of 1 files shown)


MalpenZibo:
Currently, I have an accumulator for every possible network. With this change, the accumulator only works for networks with the name specified in `interface_name` (in theory, it should be only one network, right?).

Which problem does this PR resolve (data from other networks in the download speed and upload speed indicator / wrong IP address)?

kylesferrazza:
On `main` I am getting a different IP shown every time I run ashell. I didn't look too much into the current sorting comparator, but it doesn't seem to be deterministic. Regardless, I have a bunch of local interfaces and want to be able to show my LAN IP rather than loopback/tailscale/etc randomly on every boot.

Separately, it'd be nice to be able to see more than one interface. If you'd prefer, I can work on an alternate approach to be able to configure each instance of the IpAddress indicator to show a particular interface? (Like in the description of the PR above)

MalpenZibo:
Ok, because I'm not convinced about the `config` structure changes, but right now I don't have a better idea, and I don't want to complicate things too much with a new fancy configuration. 

What do you think about that:
- filter the network's interfaces searching only: `en`, `eth`, wl`, or `waln`
- sorting the network with this priority: `en`, `eth`, `wl`, `wlan`
- then proceed like before

In theory, we should give priority to the Ethernet connection showing the correct data, also in your case, without adding other configurations.

``` rust
    let network = networks
        .iter()
        .filter(|(name, _)| {
            name.contains("en")
                || name.contains("eth")
                || name.contains("wl")
                || name.contains("wlan")
        })
        .sorted_by_key(|(name, _)| {
            if name.contains("en") {
                return 0;
            }

            if name.contains("eth") {
                return 1;
            }

            if name.contains("wl") {
                return 2;
            }

            if name.contains("wlan") {
                return 3;
            }

            99
        })
        .fold(
            (None, 0, 0),
            |(first_ip, total_received, total_transmitted), (_, data)| {
                let ip = first_ip.or_else(|| {
                    data.ip_networks()
                    
...
```

kylesferrazza:
I think that's probably better than the current approach but I would really appreciate a solution that allows specifying an exact interface or being able to show more than one interface. Any thoughts on an even simpler config that is just `{ IpAddress = "wlp5s0" }` similar to how filesystems are configured for the disk module?

kylesferrazza:
Or another idea, `IpAddress`, `DownloadSpeed` and `UploadSpeed` could be moved out of `SystemInfo` and into a reusable `NetworkInfo` module that you could call repeatedly for multiple interfaces?

```toml
[modules]
left = [ "NetworkInfo.eth0", "NetworkInfo.wlp5s0" ] # not sure what the best syntax would be to reference the array entries

[[NetworkInfo]]
interface = "eth0"
display_name = "Ethernet"
indicators = [ "IpAddress" ]

[[NetworkInfo]]
interface = "wlp5s0"
display_name = "Wifi"
indicators = [ "IpAddress", "DownloadSpeed" ]
```

MalpenZibo:
No, for now let's keep it simple.

If you can, try to add my changes and add you initial config change to provide an additional filter.

In the future we could discuss to let the use specify an interface name foreach indicator.

If you want you can open an issue for this.

kylesferrazza:
I opened two separate issues.

#321 is related to the interface changing on every run

#322 is for being able to show specific/multiple interfaces by name

kylesferrazza:
Incorporated your changes above to make the network selection deterministic. Appears to be working, thanks! I'll try some other stuff for #322 later

kylesferrazza:
Does this one need anything else?

MalpenZibo:
No it's ok, I had to complete the review but sadly recently I don't have much time left 😅

MalpenZibo (approved):

Issue #267, opened 2025-10-07
Media player not recognising spotify


CodedNil:
On the latest 0.6.0 release, it displays a - when playing from spotify. And it also has this ghost "No Title" entry at all times.

<img width="690" height="132" alt="Image" src="https://github.com/user-attachments/assets/f062c16d-f156-4eac-8153-a1f30b6ca2fa" />

<img width="733" height="447" alt="Image" src="https://github.com/user-attachments/assets/206de848-02f9-495f-ae33-546657cb9894" />

```
playerctl --all-players metadata --format '{{playerName}}: {{status}} - {{artist}} - {{title}}'
spotify: Playing - Gracie Abrams - I Love You, I'm Sorry - Live From Vevo
```

MalpenZibo:
Spotify from a web browser seems to work. You are using the desktop application?

<img width="1130" height="466" alt="Image" src="https://github.com/user-attachments/assets/1f6b8a5f-4395-4cbb-8b7b-49b894986c10" />

CodedNil:
Yes desktop app

matuck:
So while testing this noticed some weird behavior.  If spotify was open before ashell was started it seems to work as it should.    Even if i close spotify and reopen.   However, if ashell was opened before spotify it doesnt not respond correctly.

matuck:
Attaching some debug log output
First with spotify started before ashell.

```
INFO [ashell::config] Decoding config file "/home/mitchtuck/.config/ashell/config.toml"
INFO [ashell::config] Config file loaded successfully
DEBUG [ashell::services::mpris] Found MPRIS player services: ["org.mpris.MediaPlayer2.playerctld", "org.mpris.MediaPlayer2.spotify"]
INFO [ashell::services::mpris] MPRIS player service initialized
DEBUG [ashell::services::mpris] Found MPRIS player services: ["org.mpris.MediaPlayer2.playerctld", "org.mpris.MediaPlayer2.spotify"]
DEBUG [ashell::services::mpris] Volume changed: Some(0.5053635461966888)
DEBUG [ashell::services::mpris] Volume changed: Some(0.5053635461966888)
DEBUG [ashell::services::mpris] MPRIS player service receive events: [(), ()]
DEBUG [ashell::services::mpris] Found MPRIS player services: ["org.mpris.MediaPlayer2.playerctld", "org.mpris.MediaPlayer2.spotify"]
DEBUG [ashell::services::mpris] Refreshing MPRIS player data
[...]

PR #319, opened 2025-11-20
Nicer player metadata display

jazzpi:
In the menu view for the players widget:

- show title, artists and album on separate lines
- show album cover (`mpris:artUrl`)

image

I added `reqwest` for downloading the album art. This seems to pull in a lot of extra dependencies, maybe we should use something else instead? It can also only handle HTTP(S) URLs. That's all Spotify uses (which I tested with), but I assume local players might use `file://` URLs instead.

I'm also not too sure about the title/artists/album display, it's a bit hard to tell what's what.

For some reason, `nix fmt` also keeps reformatting the `flake.nix`. @kylesferrazza do you maybe understand why?

### DIFF
--- a/src/services/mpris/mod.rs
+++ b/src/services/mpris/mod.rs
@@ -153,6 +153,12 @@ impl ReadOnlyService for MprisPlayerService {
         match event {
             Event::MetadataChanged(data) => {
                 self.data = data;
+                let desired_urls: HashSet<String> = self
+                    .data
+                    .iter()
+                    .filter_map(|player| player.metadata.as_ref()?.art_url.clone())
+                    .collect();
+                self.covers.retain(|url, _| desired_urls.contains(url));
             }
             Event::CoverFetched(url, bytes) => {
                 self.covers.insert(url, bytes);
@@ -437,11 +443,15 @@ impl MprisPlayerService {
     }
 
     fn check_cover_update(data: &[MprisPlayerData], state_data: &mut ActiveData) {
-        let desired_urls: HashSet<String> = data
+        let mut desired_urls: HashSet<String> = data
             .iter()
             .filter_map(|p| p.metadata.as_ref()?.art_url.clone())
-            .filter(|url| !state_data.fetched_covers.contains(url))
             .collect();
+        // These will be removed in `update()`
+        state_data
+            .fetched_covers
+            .retain(|url| desired_urls.contains(url));
+        desired_urls.retain(|url| !state_data.fetched_covers.contains(url));
 
         for (_, handle) in state_data
             .in_flight


kylesferrazza:
Sorry about that, must have forgotten to run a final `nix fmt` in my last PR -- I just submitted #320 in case you want to rebase over it.

Also, based cory wong

jazzpi:
When you play a YT video, it publishes the thumbnail as `mpris:artUrl` with a `file://` URL. So I've added support for `file://` URLs as well.

MalpenZibo:
About request I had the same "problem" here #279 .
I didn't found a solution without less dependencies. For now we could leave request, maybe in the future I will try to find something with less dependencies.

kylesferrazza:
You can set reqwest to use rustls (with the `rustls-tls` Cargo flag) so that you don't need the openssl dependency.

jazzpi:
Done. I also messed with the widget layout a bit (see updated screenshot).

src/modules/media_player.rs
```diff
@@ -156,6 +228,59 @@ impl MediaPlayer {
         }
     }
 
+    fn check_cover_update(&mut self, data: &[MprisPlayerData]) -> Action {
```
MalpenZibo: I prefer to have this fetch function directly in the service module

src/modules/media_player.rs
```diff
@@ -103,22 +146,51 @@ impl MediaPlayer {
...
+                        None => horizontal_space().into(),
+                    };
+                    let controls = Row::new()
+                        .push(volume_slider)
+                        .push(buttons)
+                        .spacing(theme.space.md)
+                        .align_y(Vertical::Center);
+
+                    // Is it possible to dynamically size the cover to match the buttons?
+                    let buttons_width =
```
MalpenZibo: Maybe you could create two columns with `FillPortion`. In the first column, place the song info and the slider, and in the second column, the buttons and album image.

In this way, you should be able to fill the space used by the button with the image.

src/modules/media_player.rs
jazzpi: I agree, but I'm struggling a bit with how to integrate it into the architecture. I thought about adding it to the `State::Active` arm of `MprisPlayerService::start_listening()`. But I need to somehow get the `Task` back to `App::update()` so it is executed, right? I can't really figure out how to do that. I'd be grateful for some suggestions :)

src/modules/media_player.rs
jazzpi: Ah I used a column before, but didn't know about `FillPortion`. I'll try that, thanks!

src/modules/media_player.rs
MalpenZibo: You can place everything inside the `initialize_data` function in the `MprisPlayerService`. If you need to maintain a `state` to put information about the already downloaded cover, you could alter the service's Active state.

```rust
enum State {
    Init,
    Active(zbus::Connection, Map<String> // or something similar),
    Error,
}
```

MalpenZibo:
Should it work with Spotify from the browser? Because I cannot see any cover

romanstingler:
@MalpenZibo 

the issue is that when the service is initialized (`ServiceEvent::Init`), it only sets the service but doesn't trigger a cover update. The cover update only happens on `ServiceEvent::Update`.

This patch fixes the issue and removes the debug log

```
diff --git a/src/modules/media_player.rs b/src/modules/media_player.rs
index 1d08db4..49627f5 100644
--- a/src/modules/media_player.rs
+++ b/src/modules/media_player.rs
@@ -69,7 +69,13 @@ impl MediaPlayer {
             Message::Event(event) => match event {
                 ServiceEvent::Init(s) => {
                     self.service = Some(s);
-                    Action::None
+                    // Trigger cover update for initial data
+                    if let Some(service) = self.service.as_ref() {
+                        let data: Vec<MprisPlayerData> = service.iter().cloned().collect();
+                        self.check_cover_update(&data)
+                    } else {
+                        Action::None
+                    }
                 }
                 ServiceEvent::Update(d) => {
@@ -84,14 +90,10 @@ impl MediaPlayer {
                 Action::None
             }
             Message::CoverLoaded(url, data) => {
-                log::debug!("Loaded cover from {}", url);
                 self.covers.insert(url, image::Handle::from_bytes(data));
                 Action::None
             }
-            Message::CoverLoadFailed(url, err) => {
-                log::error!("Failed to load cover from {}: {}", url, err);
-                Action::None
-            }
+            Message::CoverLoadFailed(_url, _err) => Action::None,
         }
     }
```

romanstingler:
@jazzpi will you add the code of the patch into your PR? 
Otherwise, I can also create a PR on top

jazzpi:
Sorry for the radio silence. I can integrate it into the PR, but I think the bigger issue is still the separation of concerns, and the fix for that might make this moot.

I should have some time to work on this branch today hopefully.

src/modules/media_player.rs
jazzpi: If we download the covers in `initialize_data`, we'd have to wait for the download to complete (or time out) before forwarding the metadata to the media player module, right? I think it would be quite annoying if the song title update takes a couple of seconds whenever you have a spotty internet connection.

romanstingler:
this issue
> Should it work with Spotify from the browser? Because I cannot see any cover

 is related to that one I guess 
https://github.com/MalpenZibo/ashell/issues/267#issuecomment-3774593882

it updates as soon as you press a button in the menu

jazzpi:
Okay, I've moved the cover fetching to the `mpris` service. It now spawns futures for cover art fetching and `select!`s from either DBus events or these futures in the `Active` state.

There's still a couple of things to clean up, but besides that I'm running into one major problem at the moment. It looks to me like the `media_player` module does not receive the last event sent by the `mpris` service. When I pause/unpause Spotify to trigger a DBus  metadata event, `media_player` then receives the previous event (e.g. the `CoverFetched` for the song, or even the `MetadataChanged` if the song changed). Maybe this is also related to the other missing-update issues?

I've added some `debug!`s to show this:

```
...
[2026-01-21 15:02:19.398992 +01:00] DEBUG [ashell::services::mpris] src/services/mpris/mod.rs:391: tx: MetadataUpdate: Ok(())
...
[2026-01-21 15:02:19.399493 +01:00] DEBUG [ashell::modules::media_player] src/modules/media_player.rs:257: sub: MetadataChanged
...
[2026-01-21 15:02:19.399661 +01:00] DEBUG [ashell::services::mpris] src/services/mpris/mod.rs:412: tx: CoverFetched(file:///home/jasper/.mozilla/firefox/firefox-mpris/1755_2.png): Ok(())
[2026-01-21 15:02:19.399692 +01:00] DEBUG [ashell::modules::media_player] src/modules/media_player.rs:260: sub: CoverFetched(file:///home/jasper/.mozilla/firefox/firefox-mpris/1755_2.png, _)
...
[2026-01-21 15:02:19.486172 +01:00] DEBUG [ashell::services::mpris] src/services/mpris/mod.rs:412: tx: CoverFetched(https://i.scdn.co/image/ab67616d0000b2739ae41e1296c0349c2a900c4f): Ok(())
[2026-01-21 15:02:19.507073 +01:00] DEBUG [ashell::modules::media_player] src/modules/media_player.rs:65: rx: MetadataChanged
[2026-01-21 15:02:19.507152 +01:00] DEBUG [ashell::modules::media_player] src/modules/media_player.rs:68: rx: CoverFetched(file:///home/jasper/.mozilla/firefox/firefox-mpris/1755_2.png, _)
...
[2026-01-21 15:02:39.739351 +01:00] DEBUG [ashell::modules::media_player] src/modules/media_player.rs:260: sub: CoverFetched(https://i.scdn.co/image/ab67616d0000b2739ae41e1296c0349c2a900c4f, _)
...
[2026-01-21 15:02:39.742526 +01:00] DEBUG [ashell::services::mpris] src/services/mpris/mod.rs:391: tx: MetadataUpdate: Ok(())
[2026-01-21 15:02:39.753005 +01:00] DEBUG [ashell::modules::media_player] src/modules/media_player.rs:68: rx: CoverFetched(https://i.scdn.co/image/ab67616d0000b2739ae41e1296c0349c2a900c4f, _)
...
```

The `tx:` logs are printed after the `mpris` service sends an event to the channel. The `rx:` logs are printed after the `media_player` module receives the event in the `update()` function. The `sub:`  logs are printed from `MediaPlayer::subscription()`, in a `.map()` call on the subscription.

As you can see, on startup the covers for both media players are fetched, and for both the `CoverFetched` event is sent just fine. But the second `CoverFetched` is only received 20 seconds later (after I paused/unpaused Spotify).

I'm at a complete loss why this is happening. Any ideas?

<details>

<summary>Full log</summary>

```
[2026-01-21 15:02:19.262962 +01:00] INFO [ashell::config] src/config.rs:747: Decoding config file "/home/jasper/.config/ashell/config.toml"
[2026-01-21 15:02:19.263742 +01:00] INFO [ashell::config] src/config.rs:753: Config file loaded successfully
[2026-01-21 15:02:19.368961 +01:00] DEBUG [ashell::services::mpris] src/services/mpris/mod.rs:184: Found MPRIS player services: ["org.mpris.MediaPlayer2.firefox.instance_1_27", "org.mpris.MediaPlayer2.spotify"]
[2026-01-21 15:02:19.376150 +01:00] INFO [ashell::services::mpris] src/services/mpris/mod.rs:341: MPRIS player service initialized
[2026-01-21 15:02:19.376870 +01:00] DEBUG [ashell::services::mpris] src/services/mpris/mod.rs:184: Found MPRIS player services: ["org.mpris.MediaPlayer2.firefox.instance_1_27", "org.mpris.MediaPlayer2.spotify"]
[2026-01-21 15:02:19.395817 +01:00] DEBUG [ashell::services::mpris] src/services/mpris/mod.rs:290: Volume changed: Some(0.7341573205157549)
[2026-01-21 15:02:19.396003 +01:00] DEBUG [ashell::services::mpris] src/services/mpris/mod.rs:383: MPRIS player service receive events: [()]
[2026-01-21 15:02:19.396657 +01:00] DEBUG [ashell::services::mpris] src/services/mpris/mod.rs:184: Found MPRIS player services: ["org.mpris.MediaPlayer2.firefox.instance_1_27", "org.mpris.MediaPlayer2.spotify"]
[2026-01-21 15:02:19.398883 +01:00] DEBUG [ashell::services::mpris] src/services/mpris/mod.rs:386: Refreshing MPRIS player data
[2026-01-21 15:02:19.398992 +01:00] DEBUG [ashell::services::mpris] src/services/mpris/mod.rs:391: tx: MetadataUpdate: Ok(())
libEGL warning: pci id for fd 57: 10de:2486, driver (null)

[2026-01-21 15:02:19.399493 +01:00] DEBUG [ashell::modules::media_player] src/modules/media_player.rs:257: sub: MetadataChanged
[2026-01-21 15:02:19.399607 +01:00] DEBUG [ashell::services::mpris] src/services/mpris/mod.rs:409: Fetched cover art from file:///home/jasper/.mozilla/firefox/firefox-mpris/1755_2.png
[2026-01-21 15:02:19.399661 +01:00] DEBUG [ashell::services::mpris] src/services/mpris/mod.rs:412: tx: CoverFetched(file:///home/jasper/.mozilla/firefox/firefox-mpris/1755_2.png): Ok(())
[2026-01-21 15:02:19.399692 +01:00] DEBUG [ashell::modules::media_player] src/modules/media_player.rs:260: sub: CoverFetched(file:///home/jasper/.mozilla/firefox/firefox-mpris/1755_2.png, _)
pci id for fd 56: 10de:2486, driver (null)
pci id for fd 59: 10de:2486, driver (null)
libEGL warning: egl: failed to create dri2 screen
pci id for fd 56: 10de:2486, driver (null)
pci id for fd 59: 10de:2486, driver (null)
libEGL warning: egl: failed to create dri2 screen
libEGL warning: pci id for fd 62: 10de:2486, driver (null)

pci id for fd 44: 10de:2486, driver (null)
pci id for fd 63: 10de:2486, driver (null)
libEGL warning: egl: failed to create dri2 screen
pci id for fd 44: 10de:2486, driver (null)
pci id for fd 63: 10de:2486, driver (null)
libEGL warning: egl: failed to create dri2 screen
[2026-01-21 15:02:19.485974 +01:00] DEBUG [ashell::services::mpris] src/services/mpris/mod.rs:409: Fetched cover art from https://i.scdn.co/image/ab67616d0000b2739ae41e1296c0349c2a900c4f
[2026-01-21 15:02:19.486172 +01:00] DEBUG [ashell::services::mpris] src/services/mpris/mod.rs:412: tx: CoverFetched(https://i.scdn.co/image/ab67616d0000b2739ae41e1296c0349c2a900c4f): Ok(())
[2026-01-21 15:02:19.507073 +01:00] DEBUG [ashell::modules::media_player] src/modules/media_player.rs:65: rx: MetadataChanged
[2026-01-21 15:02:19.507152 +01:00] DEBUG [ashell::modules::media_player] src/modules/media_player.rs:68: rx: CoverFetched(file:///home/jasper/.mozilla/firefox/firefox-mpris/1755_2.png, _)
[2026-01-21 15:02:19.507186 +01:00] DEBUG [ashell::services::mpris] src/services/mpris/mod.rs:141: Updating MPRIS player cover for file:///home/jasper/.mozilla/firefox/firefox-mpris/1755_2.png, 71164 bytes
[2026-01-21 15:02:39.739066 +01:00] DEBUG [ashell::services::mpris] src/services/mpris/mod.rs:311: PlaybackStatus changed: Playing
[2026-01-21 15:02:39.739172 +01:00] DEBUG [ashell::services::mpris] src/services/mpris/mod.rs:383: MPRIS player service receive events: [()]
[2026-01-21 15:02:39.739351 +01:00] DEBUG [ashell::modules::media_player] src/modules/media_player.rs:260: sub: CoverFetched(https://i.scdn.co/image/ab67616d0000b2739ae41e1296c0349c2a900c4f, _)
[2026-01-21 15:02:39.740012 +01:00] DEBUG [ashell::services::mpris] src/services/mpris/mod.rs:184: Found MPRIS player services: ["org.mpris.MediaPlayer2.firefox.instance_1_27", "org.mpris.MediaPlayer2.spotify"]
[2026-01-21 15:02:39.742438 +01:00] DEBUG [ashell::services::mpris] src/services/mpris/mod.rs:386: Refreshing MPRIS player data
[2026-01-21 15:02:39.742526 +01:00] DEBUG [ashell::services::mpris] src/services/mpris/mod.rs:391: tx: MetadataUpdate: Ok(())
[2026-01-21 15:02:39.753005 +01:00] DEBUG [ashell::modules::media_player] src/modules/media_player.rs:68: rx: CoverFetched(https://i.scdn.co/image/ab67616d0000b2739ae41e1296c0349c2a900c4f, _)
[2026-01-21 15:02:39.753114 +01:00] DEBUG [ashell::services::mpris] src/services/mpris/mod.rs:141: Updating MPRIS player cover for https://i.scdn.co/image/ab67616d0000b2739ae41e1296c0349c2a900c4f, 86294 bytes
```

</details>

romanstingler:
I am currently busy, I can have a look later in the evening or tomorrow morning.

src/services/mpris/mod.rs
```diff
@@ -277,68 +322,169 @@ impl MprisPlayerService {
...
+    ) -> State {
+        match Self::dbus_events(&conn).await {
+            Ok(dbus_events) => {
+                let mut chunks = dbus_events.ready_chunks(10);
 
-                        match Self::initialize_data(&conn).await {
-                            Ok(data) => {
-                                debug!("Refreshing MPRIS player data");
+                let mut pending_downloads = FuturesUnordered::new();
+                let mut in_flight: HashMap<String, AbortHandle> = HashMap::new();
```
jazzpi: Keeping these here instead of as part of `State` means we will loses them if we exit this function (i.e. if the DBus events returns `None`). I've added a `warn!` below if we ever exit the loop and haven't seen it triggered so far. I'm also not sure what it means if we get a `None` from the DBus events. Does that mean we lost connection to DBus? Then just re-entering the `Active` state with the same connection probably doesn't make sense anyways.

But if you prefer, I think we can move these two into the `State` as well.

src/modules/media_player.rs
```diff
@@ -103,22 +133,52 @@ impl MediaPlayer {
...
+                        .and_then(|m| m.art_url.as_ref())
+                        .map(|url| {
+                            service
+                                .get_cover(url)
+                                .map(|img| {
+                                    // TODO: store image::Handle in the service to avoid cloning
+                                    image(image::Handle::from_bytes(img.clone()))
+                                        .filter_method(image::FilterMethod::Linear)
+                                        .width(buttons_width)
+                                        .into()
```
jazzpi: Cloning the cover art every time `menu_view` is called feels a bit weird. We could directly store `image::Handle`s in the MPRIS service. But then we have iced widget data in the service instead of the module, which also feels a bit weird. Which do you prefer?

src/modules/media_player.rs
jazzpi: Should be done. I've added some comments below about parts of the implementation I'm not so sure about, would love some feedback there :)

jazzpi:
I replaced `pending_downloads.next().fuse()` with `pending_downloads.select_next_some()` and that fixed it. Now ashell also doesn't use a full CPU core anymore :see_no_evil:

src/services/mpris/mod.rs
romanstingler: What about keeping both structures inside the `State::Active` variant, so when the loop exits we hand the existing futures back into the next `active()` run and nothing is lost.
We can also treat a `None` chunk as a signal to transition to `Init` or `Error` and rebuild the connection if needed, but at minimum persisting the state prevents dropped downloads.
maybe something like 
[mpris-state-persistence.patch](https://github.com/user-attachments/files/24779395/mpris-state-persistence.patch)

src/modules/media_player.rs
romanstingler: I agree that cloning on every menu_view render is wasteful, but I’d rather keep the MPRIS service free of iced types. 
I think we can add a small cache inside the media module (`cover_handles: HashMap<String, image::Handle>`) that syncs with `MetadataChanged/CoverFetched `events. 
The service still owns plain Bytes, and the module lazily wraps them into handles once per URL, pruning any that fall out of the metadata. 
This gives us the reuse we wanted without pushing UI concerns into the service. 
Maybe something like 
[cover-handle-cache.patch](https://github.com/user-attachments/files/24778648/cover-handle-cache.patch)

romanstingler:
> I replaced `pending_downloads.next().fuse()` with `pending_downloads.select_next_some()` and that fixed it. Now ashell also doesn't use a full CPU core anymore 🙈

probably you built-in a crypto miner :P

MalpenZibo:
image

It seems stuck loading the cover :sweat_smile:

jazzpi:
Well that's not supposed to happen :sweat_smile: 

If you run with `RUST_LOG=ashell::services::mpris=debug,ashell::modules::media_player=debug"`, do you see anything interesting in the log?

And what media player are you testing with?

romanstingler:
jazzpi https://www.youtube.com/watch?v=lqQLEP1-yOw

[mod.rs.txt](https://github.com/user-attachments/files/24936575/mod.rs.txt)

When ashell starts while no MPRIS players exist, `MprisPlayerService::dbus_events()` builds the DBus event stream with:
the global `NameOwnerChanged` listener, but zero per-player listeners (`receive_metadata_changed`, `receive_playback_status_changed`, etc.) because there are no players yet.

Later, when Spotify appears, the service notices it via `NameOwnerChanged` and does one refresh, but it still isn’t subscribed to per-player signals. 

And therefore, Spotify’s metadata updates (title/artist/artUrl) don’t arrive at all.
The UI stays at “No title/Unknown …” (or “Loading cover…” when empty strings were treated as real values).
As soon as you press a button, `command()` triggers a manual metadata refresh, so the text suddenly appears.


I guess we could detect player set changes and restart the event stream in `active()`.

take an initial snapshot of players and build `dbus_events_for_players(conn, &snapshot)` which subscribes to per-player signals for that snapshot.
Then on events, refresh the player list (if the set differs), break out so `active()` runs again and rebuilds subscriptions for the new set.

This should make Spotify metadata update automatically once it starts playing, without any periodic polling and without needing a button press.

Additionally, Spotify sometimes reports empty strings/empty arrays initially ("", []). 
We normalize those to `None` in `MprisPlayerMetadata::from`, so the UI doesn’t treat `art_url=""` as a real URL.


https://youtu.be/zWbxJ1nGFwY

jazzpi:
@romanstingler that's a different issue from @MalpenZibo's (since there, there is metadata, just no cover). I just tested with the current nixpkgs version (i.e. 0.7.0) and the issue also happens there. So I don't think it's related to the changes in this PR, it's part of #447.

src/modules/media_player.rs
jazzpi: If we use dynamically sized columns, we would need some way to force an aspect ratio of 1. Otherwise the "Loading cover..." text will have a much smaller height than the album art, causing the menu view to "flicker" when a new cover is loaded.

With `iced` 0.14, there's `widget::Grid` that should make that possible. With `cosmic`, there's `widget::aspect_ratio`. But since we have neither, this is the best solution IMO.

jazzpi:
Okay I took a shot at fixing it anyways. I also removed the event filtering in `MprisPlayerService::dbus_events()`. So I think this should now fix #447 in general.

romanstingler:
It works for me, but I had no time for a code review. 
I just got a new coworker and have to onboard him. 

I will see if I'll have some time in the next days.

MalpenZibo:
image
Nice, it seems to be working fine. I'll do a final review, and then we're ready to merge it

MalpenZibo:
image

```rust 
                    let description = column![title, artists, album].spacing(theme.space.xxs);

                    let play_pause_icon = match d.state {
                        PlaybackStatus::Playing => StaticIcon::Pause,
                        PlaybackStatus::Paused | PlaybackStatus::Stopped => StaticIcon::Play,
                    };

                    let buttons = row![
                        icon_button(theme, StaticIcon::SkipPrevious)
                            .on_press(Message::Prev(d.service.clone()))
                            .size(IconButtonSize::Large),
                        icon_button(theme, play_pause_icon)
                            .on_press(Message::PlayPause(d.service.clone()))
                            .size(IconButtonSize::Large),
                        icon_button(theme, StaticIcon::SkipNext)
                            .on_press(Message::Next(d.service.clone()))
                            .size(IconButtonSize::Large),
                    ]
                    .align_y(Vertical::Center)
                    .spacing(theme.space.xs);

                    let volume_slider: Element<'_, _> = match d.volume {
                        Some(v) => slider(0.0..=100.0, v, move |v| {
                            Message::SetVolume(d.service.clone(), v)
                        })
                        .into(),
                        None => horizontal_space().into(),
                    };

                    let cover: Element<_, _> = d
                        .metadata
                        .as_ref()
                        .and_then(|m| m.art_url.as_ref())
                        .map(|url| {
                            self.covers
                                .get(url)
                                .map(|handle| {
                                    image(handle)
                                        .filter_method(image::FilterMethod::Linear)
                                        .into()
                                })
                                .unwrap_or_else(|| {
                                    text("Loading cover...").height(Length::Fill).into()
                                })
                        })
                        .unwrap_or_else(|| horizontal_space().into());

                    let left_width = 3;
                    let right_width = 2;

                    container(
                        column!(
                            row!(
                                container(description).width(Length::FillPortion(left_width)),
                                container(cover)
                                    .align_x(Horizontal::Center)
                                    .width(Length::FillPortion(right_width))
                            )
                            .spacing(theme.space.xs),
                            row!(
                                container(volume_slider).width(Length::FillPortion(left_width)),
                                container(buttons)
                                    .align_x(Horizontal::Center)
                                    .width(Length::FillPortion(right_width))
                            )
                            .align_y(Vertical::Center)
                            .spacing(theme.space.xs),
                        )
                        .spacing(theme.space.md),
                    )
                    .style(move |app_theme: &Theme| container::Style {
                        background: Background::Color(
                            app_theme
                                .extended_palette()
                                .secondary
                                .strong
                                .color
                                .scale_alpha(theme.opacity),
                        )
                        .into(),
                        border: Border::default().rounded(theme.radius.lg),
                        ..container::Style::default()
                    })
                    .padding(theme.space.md)
                    .width(Length::Fill)
                    .into()
                }))
                .spacing(theme.space.md)

```

To avoid the size calculation and remove the IconButton size thing

jazzpi:
Hm I will try this later, but I think I tried a similar solution before and ran into two "problems":

- the "Loading cover..." text will take up less vertical space than the cover, so anything below the cover (buttons or other player metadata) will "flicker" while the cover is loading
- if you have a player without a cover or volume controls, the buttons will be below and to the right of the text

Btw if we manage to get rid of it, should I revert the changes to `src/components/icon.rs` as well?

MalpenZibo:
If we want the the slider and button on the same line vertically center we had to go with a two row solution.

We can change the layout in case of missing cover (description full width) and maybe to resolve the loading message we could use a min height.

MalpenZibo:
Personally I prefer to remove the icon stuff

jazzpi:
Okay I think I covered all cases:

image

I left out the min height for the "Loading..." text because the cover doesn't have to be square anyways -- so even if you make the "Loading..." square, it could still flicker.

jazzpi:
Oops, I missed that the skip song button is a bit squashed. Should be fixed now by using 3/2 instead of 2/1 `FillPortion`s:

image

I also fixed the merge conflicts. So this should finally be ready :crossed_fingers:

clotodex:
Tag me when it is ready for testing then i will try it out a bit

src/services/mpris/mod.rs
```diff
@@ -71,37 +80,84 @@ impl From<HashMap<String, OwnedValue>> for MprisPlayerMetadata {
...
+            art_url,
+        }
     }
 }
 
 #[derive(Debug, Clone)]
 pub struct MprisPlayerService {
     data: Vec<MprisPlayerData>,
     conn: zbus::Connection,
+    pub covers: HashMap<String, Bytes>,
```
romanstingler: is that `pub` necessary ?

romanstingler: I haven't seen any `retain` or `remove` for `MprisPlayerService.covers`. 
you could use a LRU or just check if the size exceeds a max amount and then drop the oldest one.

the loaded_cover urls are not so big but the picture data could be big if you keep a long running system.

src/services/mpris/mod.rs
```diff
@@ -277,68 +334,156 @@ impl MprisPlayerService {
...
         }
     }
+
+    async fn update_data(output: &mut Sender<ServiceEvent<Self>>, state_data: &mut ActiveData) {
+        match Self::initialize_data(&state_data.conn).await {
+            Ok(data) => {
+                debug!("Refreshing MPRIS player data for {} players", data.len());
+                Self::check_cover_update(&data, state_data);
+                let _ = output
+                    .send(ServiceEvent::Update(Event::MetadataChanged(data.clone())))
```
romanstingler: This clone is a bit expensive `MprisPlayerData` contains an `MprisPlayerProxy` which wraps a D-Bus connection, which means that we clone 

- Connection state - D-Bus socket references
- Internal buffers - Message queues and caches
- Metadata structures - Service discovery info

This happens on every metadata, volume, and playback status change.
Volume changes trigger that like 60fps.

You could just wrap that in an Arc like that. 
I did a short benchmark on my debug build I got 40% less time cloning stuff.

```
diff --git a/src/services/mpris/mod.rs b/src/services/mpris/mod.rs
index 252166b..ef564aa 100644
--- a/src/services/mpris/mod.rs
+++ b/src/services/mpris/mod.rs
@@ -17,6 +17,7 @@ use std::{
     any::TypeId,
     collections::{HashMap, HashSet},
     fmt::Display,
+    sync::Arc,
 };
 use url::Url;
 use zbus::{fdo::DBusProxy, zvariant::OwnedValue};
@@ -47,7 +48,7 @@ pub struct MprisPlayerData {
     pub metadata: Option<MprisPlayerMetadata>,
     pub volume: Option<f64>,
     pub state: PlaybackStatus,
-    proxy: MprisPlayerProxy<'static>,
+    proxy: Arc<MprisPlayerProxy<'static>>,
 }
 
 #[derive(PartialEq, Eq, Debug, Clone)]
@@ -252,7 +253,7 @@ impl MprisPlayerService {
                 metadata,
                 volume,
                 state,
-                proxy,
+                proxy: Arc::new(proxy),
             })
         }))
         .await
@@ -424,10 +425,20 @@ impl MprisPlayerService {
     async fn update_data(output: &mut Sender<ServiceEvent<Self>>, state_data: &mut ActiveData) {
         match Self::initialize_data(&state_data.conn).await {
             Ok(data) => {
-                debug!("Refreshing MPRIS player data for {} players", data.len());
+                let start_time = std::time::Instant::now();
+                let cloned_data = data.clone();
+                let clone_duration = start_time.elapsed();
+
+                debug!(
+                    "MPRIS data clone: {} players, {:?}μs",
+                    data.len(),
+                    clone_duration.as_micros()
+                );
+
+                error!("Refreshing MPRIS player data for {} players", data.len());
                 Self::check_cover_update(&data, state_data);
                 let _ = output
-                    .send(ServiceEvent::Update(Event::MetadataChanged(data.clone())))
+                    .send(ServiceEvent::Update(Event::MetadataChanged(cloned_data)))
                     .await;
             }
             Err(err) => {
@@ -509,7 +520,7 @@ impl Service for MprisPlayerService {
             let s = self.data.iter().find(|d| d.service == command.service_name);
 
             if let Some(s) = s {
-                let mpris_player_proxy = s.proxy.clone();
+                let mpris_player_proxy = Arc::clone(&s.proxy);
                 let conn = self.conn.clone();
                 iced::Task::perform(
                     async move {
```

src/services/mpris/mod.rs
```diff
@@ -71,37 +80,84 @@ impl From<HashMap<String, OwnedValue>> for MprisPlayerMetadata {
...
     conn: zbus::Connection,
+    pub covers: HashMap<String, Bytes>,
 }
 
-impl Deref for MprisPlayerService {
-    type Target = Vec<MprisPlayerData>;
-
-    fn deref(&self) -> &Self::Target {
+impl MprisPlayerService {
+    pub fn players(&self) -> &Vec<MprisPlayerData> {
```
romanstingler: this could be 
` pub fn players(&self) -> &[MprisPlayerData] {`

src/modules/media_player.rs
```diff
@@ -191,4 +259,29 @@ impl MediaPlayer {
     pub fn subscription(&self) -> Subscription<Message> {
         MprisPlayerService::subscribe().map(Message::Event)
     }
+
+    fn sync_cover_handles(&mut self) {
```
romanstingler, line 263: This is optional and open for discussion (we are talking about `ns` and fewer allocations. 
For small numbers <20 players (I would argue the most probable use-case), 
a Vec is faster than a Hashmap. 


```
    fn sync_cover_handles(&mut self) {
        let Some(service) = &self.service else {
            return;
        };

        let mut desired_urls = HashSet::new();
        let mut missing_urls = Vec::new();

        for player in service.players() {
            if let Some(art_url) = player.metadata.as_ref().and_then(|m| m.art_url.as_ref()) {
                desired_urls.insert(art_url.clone());
                if !self.covers.contains_key(art_url) {
                    missing_urls.push(art_url);
                }
            }
        }

        self.covers.retain(|url, _| desired_urls.contains(url));

        for url in missing_urls {
            if let Some(cover) = service.get_cover(url) {
                self.covers
                    .insert(url.clone(), image::Handle::from_bytes(cover.clone()));
            }
        }
    }
```

| Scenario | Old (2 HashSets) | New (HashSet + Vec) | Delta |
|---|---|---|---|
| 1 player, cold | 1,035 ns | 937 ns | -9.5% |
| 1 player, warm | 975 ns | 898 ns | -7.9% |
| 3 players, 1 stale | 3,492 ns | 3,312 ns | -5.2% |
| 5 players, 3 stale | 6,865 ns | 4,664 ns | -32.1% |
| 10 players, 5 stale | 11,651 ns | 11,234 ns | -3.6% |

src/services/mpris/mod.rs
jazzpi: huh, wonder why clippy doesn't catch that

src/services/mpris/mod.rs
jazzpi: The clone is unnecessary in general, we can just move `data` here. I'll remove it. I guess `data` will be cloned by iced later, but that was the case before as well. And this PR is almost 3 months old, I'd really like to just get it merged. IMHO this can be changed in a later PR.

src/modules/media_player.rs
jazzpi: I think readability/clarity of intent > saving a couple of nanoseconds

PR #323, opened 2025-11-21
Dynamic menu wrapper

MalpenZibo:
After these changes, the menu wrapper no longer needs a fixed width to position the menu

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -2,7 +2,7 @@ use crate::{
     HEIGHT,
     config::{self, AppearanceStyle, Config, Modules, Position},
     get_log_spec,
-    menu::{MenuSize, MenuType},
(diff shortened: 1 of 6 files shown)


romanstingler:
I just had a quick look, because I am not good at UI stuff (I have no passion for that)
but 


I tried to find the issue and as far as I have understood it is because both the row and each child use Length::Shrink, the container’s `max_width(MenuSize::XLarge)` simply limits how large the overlay can grow, but it doesn’t scale down children that collectively exceed that cap. 
When a child needs more room than the available max width, Iced lets it overflow rather than forcefully shrinking it.

But I have no idea if it is the hourly scroll area and daily forecast who request such a wide size

the default tempo branch looks fine 
image

PR #328, opened 2025-11-24
niri-support

clotodex:
Looking for testers! If you are on nixos, use this url for ashell: `ashell.url = "github:clotodex/ashell-iwd/feature/niri-support";`
- created a new CompositorService
- implemented niri support for:
  - [x] workspaces
  - [x] window title
  - [x] keyboard layout (module still todo)
- pulled hyprland implementation out of the modules into the service
- The CompositorService will first try and load hyprland - if that fails it will load niri

I ran some inital tests on my machine to validate it working, but I did not put it through the ringer yet:
- [x] verify colors of the workspaces
- [x] test (and fix) commands -> scrolling and clicking does not work yet (broken for both)
  - [x] this also requires forwarding the commands to the chosen backend 
- [x] test hyprland

**Open decisions:**
- [x] should the backend be something to be "failed over to" like it is right now and with network, or a config option?
- [x] Readme, intro, website, doc adjustments - this is no longer just limited to hyprland :)
- [x] debate empty / full init state of the service - right now niri immediately fills it, so would like to keep it empty
- [x] changed the scrolling on worksapces to be in the right direction - was this a bug or intended? I assumee it was "if I scroll up I expect the workspace number to increase" while scrolling layouts (also on hyprland) scroll down with increasing workspace index. This will depend on how you configure workspace animation direction and might just be a config preference - or something to be handled on a compositor to compositor basis - what are your takes?

Problems that remain open (but would merge anyway, potentially with doc comment):
- niri can lead to an infinite loop with enable_workspace_filling, since niri removes empty ones - should be checked or mentioned somehow
- there is no submap for niri atm https://github.com/YaLTeR/niri/issues/846 and https://github.com/MaxVerevkin/wlr-which-key (which is what most are using I assume) does not have a dbus / ipc

This also allows implementing other compositors and wayland porotocols later.
*(I did not implement the ext/workspaces yet - I could not get the smithay client to work on yielding the right events (the control loop is different - tried to copy idle_inhibitor but failed so far)
I focused on a full ipc implementation, since wayland afaik has no way of discovering window titles atm - so an implementation was needed anyways*

Any feedback, especially on architectural choices (@MalpenZibo ) are very welcome!

### DIFF
--- a/src/modules/workspaces.rs
+++ b/src/modules/workspaces.rs
@@ -45,6 +45,7 @@ pub enum Message {
     ChangeWorkspace(i32),
     ToggleSpecialWorkspace(i32),
     Scroll(i32),
+    ConfigReloaded(WorkspacesModuleConfig),
 }
 
 pub struct Workspaces {
@@ -313,6 +314,11 @@ impl Workspaces {
                 if let Some(next) = next_workspace {
                     return self.update(Message::ChangeWorkspace(next.id));
                 }
+                iced::Task::none()
+            }
+            Message::ConfigReloaded(cfg) => {
+                self.config = cfg;
+                self.recalculate_ui_workspaces();
                 iced::Task::none()
             }
         }
--- a/src/app.rs
+++ b/src/app.rs
@@ -159,10 +159,29 @@ impl App {
         self.custom = custom;
         self.updates = config.updates.map(Updates::new);
         self.clipboard = config.clipboard_cmd.map(Clipboard::new);
-        self.workspaces = Workspaces::new(config.workspaces);
-        self.window_title = WindowTitle::new(config.window_title);
+
+        // ignore task, since config change should not generate any
+        let _ = self
+            .workspaces
+            .update(modules::workspaces::Message::ConfigReloaded(
+                config.workspaces,
+            ))
+            .map(Message::Workspaces);
+
+        self.window_title
+            .update(modules::window_title::Message::ConfigReloaded(
+                config.window_title,
+            ));
+
         self.system_info = SystemInfo::new(config.system_info);
-        self.keyboard_layout = KeyboardLayout::new(config.keyboard_layout);
+
+        let _ = self
+            .keyboard_layout
+            .update(modules::keyboard_layout::Message::ConfigReloaded(
+                config.keyboard_layout,
+            ))
+            .map(Message::KeyboardLayout);
+
         self.keyboard_submap = KeyboardSubmap::default();
         self.clock = Clock::new(config.clock);
         self.settings
--- a/src/modules/keyboard_layout.rs
+++ b/src/modules/keyboard_layout.rs
@@ -12,6 +12,7 @@ use iced::{Element, Subscription, Task, widget::text};
 pub enum Message {
     ServiceEvent(ServiceEvent<CompositorService>),
     ChangeLayout,
+    ConfigReloaded(KeyboardLayoutModuleConfig),
 }
 
 pub struct KeyboardLayout {
@@ -47,6 +48,10 @@ impl KeyboardLayout {
                         .command(CompositorCommand::NextLayout)
                         .map(Message::ServiceEvent);
                 }
+                Task::none()
+            }
+            Message::ConfigReloaded(new_config) => {
+                self.config = new_config;
                 Task::none()
             }
         }
--- a/src/modules/window_title.rs
+++ b/src/modules/window_title.rs
@@ -12,6 +12,7 @@ use iced::{
 #[derive(Debug, Clone)]
 pub enum Message {
     ServiceEvent(ServiceEvent<CompositorService>),
+    ConfigReloaded(WindowTitleConfig),
 }
 
 pub struct WindowTitle {
@@ -44,6 +45,10 @@ impl WindowTitle {
                 }
                 _ => {}
             },
+            Message::ConfigReloaded(cfg) => {
+                self.config = cfg;
+                self.recalculate_value();
+            }
         }
     }
 


jazzpi:
I've been [working on an `ext-workspace-v1` implementation](https://github.com/jazzpi/ashell/blob/ext-workspace-v1/src/services/workspaces/wayland.rs) (still very much WIP). I didn't use the smithay client, but yeah as far as I understand the idle inhibitor doesn't have an event loop at all (doesn't need one either since it doesn't receive any events after the initial `roundtrip()`).

I'm not sure how to best include this. I suppose the cleanest way would be to fully rely on the workspace protocol for workspace information. But the protocol doesn't expose all the information that the workspaces module currently uses (i.e. no way to determine which workspaces are empty, or which workspaces are visible but not focused). OTOH it exposes some information that isn't available to the workspaces module at the moment (urgent state, and hidden state for special workspaces).

Maybe we could use the workspace protocol as the common denominator and have it accept messages from the compositor services with additional information?

clotodex:
> I've been [working on an `ext-workspace-v1` implementation](https://github.com/jazzpi/ashell/blob/ext-workspace-v1/src/services/workspaces/wayland.rs) (still very much WIP). I didn't use the smithay client, but yeah as far as I understand the idle inhibitor doesn't have an event loop at all (doesn't need one either since it doesn't receive any events after the initial `roundtrip()`).
> 
> I'm not sure how to best include this. I suppose the cleanest way would be to fully rely on the workspace protocol for workspace information. But the protocol doesn't expose all the information that the workspaces module currently uses (i.e. no way to determine which workspaces are empty, or which workspaces are visible but not focused). OTOH it exposes some information that isn't available to the workspaces module at the moment (urgent state, and hidden state for special workspaces).
> 
> Maybe we could use the workspace protocol as the common denominator and have it accept messages from the compositor services with additional information?

I would love to have it as a fall back compositor implementation for sure.
In terms of what it exposes and how we deal with it - i am already fighting with niri, since it does not expose any kind of monitor numbering - nor has an idea of what a primary_monitor is.
So I will have to figure out some hacks or defaults.

This PR will likely still move a lot in its API. Right now my plan would be to clean this up and merge it - then add new compositor implementations as PRs on top. If this is too slow for you, let me know and we can somehow merge earlier.

src/services/compositor/niri.rs
```diff
@@ -0,0 +1,252 @@
...
+    UnixStream::from_std(std_stream).context("Failed to convert stream")
+}
+
+async fn send_command_request(stream: &mut UnixStream, request: Request) -> Result<()> {
+    let mut json = serde_json::to_string(&request)?;
+    json.push('\n');
+    stream.write_all(json.as_bytes()).await?;
+    stream.flush().await?;
+
+    let mut reader = BufReader::new(stream);
```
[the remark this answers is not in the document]
clotodex: not ready yet, but cool reference code

src/modules/workspaces.rs
```diff
@@ -161,432 +367,46 @@ fn get_workspaces(config: &WorkspacesModuleConfig) -> Vec<Workspace> {
...
-                    self.workspaces
-                        .iter()
-                        .filter(|w| w.id < current_id)
-                        .max_by_key(|w| w.id)
-                };
-                let Some(next_workspace) = next_workspace else {
-                    return;
-                };
-                Self::update(self, Message::ChangeWorkspace(next_workspace.id));
+    if config.enable_workspace_filling && !result.is_empty() {
```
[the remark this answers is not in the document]
clotodex: this is anyways not supported with niri - since ashell auto-creates the workspaces

src/services/compositor/hyprland.rs
```diff
@@ -0,0 +1,156 @@
...
+            let d = if dir > 0 { "+1" } else { "-1" };
+            Dispatch::call(DispatchType::Workspace(
+                WorkspaceIdentifierWithSpecial::Relative(d.to_string().parse()?),
+            ))?;
+        }
+        CompositorCommand::NextLayout => {
+            hyprland::ctl::switch_xkb_layout::call(
+                "all",
+                hyprland::ctl::switch_xkb_layout::SwitchXKBLayoutCmdTypes::Next,
+            )?;
```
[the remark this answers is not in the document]
clotodex: actually it IS the keyboard!

src/services/compositor/niri.rs
```diff
@@ -0,0 +1,252 @@
...
+
+pub async fn execute_command(cmd: CompositorCommand) -> Result<()> {
+    let mut stream = connect().await?;
+
+    let action = match cmd {
+        CompositorCommand::FocusWorkspace(id) => match u8::try_from(id) {
+            Ok(idx) => Action::FocusWorkspace {
+                reference: WorkspaceReferenceArg::Index(idx),
+            },
+            Err(_) => return Err(anyhow!("Invalid workspace index for Niri (must be u8)")),
```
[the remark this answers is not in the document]
clotodex: done

clotodex:
@MalpenZibo whenever you get to it, i think this is ready for initial review :heart_hands: 
(The git diff is often matching on weird lines so that diffs are harder to read than necessary ...)
In the meantime i'll daily drive this on niri to test it :)

jazzpi:
> I would love to have it as a fall back compositor implementation for sure. In terms of what it exposes and how we deal with it - i am already fighting with niri, since it does not expose any kind of monitor numbering - nor has an idea of what a primary_monitor is. So I will have to figure out some hacks or defaults.

With the workspace protocol we can map workspaces (`ext_workspace_handle_v1`) to outputs (`wl_output`) in a compositor-agnostic way. `wl_output`s also don't have a monitor numbering though, but they do have names and descriptions. As far as I can see, the monitor ID is only used for the workspace colors, right? That could be mapped from the name or description instead (but that would be a breaking change of course).

> This PR will likely still move a lot in its API. Right now my plan would be to clean this up and merge it - then add new compositor implementations as PRs on top. If this is too slow for you, let me know and we can somehow merge earlier.

I think this PR is much further along than my branch, so I'd rather base the workspace protocol implementation off this :)

src/services/compositor/niri.rs
```diff
@@ -0,0 +1,275 @@
...
+            } else {
+                None
+            }
+        })
+        .collect();
+    // INFO: this is how niri sorts the outpus internally (niri msg outputs - in client.rs)
+    let outputs = output_to_active_ws
+        .keys()
+        .sorted_unstable()
+        .collect::<Vec<_>>();
```
clotodex, line 182: @jazzpi this is how I handled the monitor ordering in niri (since they also dont have a default order)
It seems like the default for niris cli is sorting it.
I could also imagine an optional config value marking the main montior and otherwise using eDP-1 or sth. What do you think?

Otherwise, in the current implementation, this is a choice up to the compositor implementation

clotodex: Actually @jazzpi lets create a new issue and discuss a generic wayland compositor there if you dont mind :D
 #329

MalpenZibo:
I will look as soon as possible. Thank you so much, this is very appreciated. Sadly, I haven't had much free time lately, but I'll try my best to check this out!

clotodex:
> I will look as soon as possible. Thank you so much, this is very appreciated. Sadly, I haven't had much free time lately, but I'll try my best to check this out!

To help the review:

What I did:
- pulled out hyprland into the CompositorService
- implemented niri using the niri-ipc
- rewrote the affected modules to use the compositorservice

How to find yourself around the files:
- start with types.rs to see the new interface / message the service can expose
- then move on to compositor/mod.rs
- the CompositorService has a single Message which is CompositorStateChanged
- there is also an ActionPerformed Message that gets triggered e.g. on workspace click. This is useful if in the future a compositor needs a hint to recompute - with the IPC this is a noop.
- the service keeps track of its `CompositorChoice`, currently only hyprland and niri and otherwise delegates its work to the specific implementation
- the modules now only listen to the same service subscription and consume the state - this also needed changes in message communication since the commands cant be performed only within a module anymore (hyprland specific code was there before)

Hope this helps :)

What to look out for especially:
- I see two risks: 1) in moving out hyprland I could have made some mistakes and lost edge cases (I tested everything but still)
- and 2) In pulling the hyprland-specific logic from the modules into the services the same applies (again also tested)
- I feel like I made a few tradeoffs to get this working soon, leaving some architectural choices for later => anything you think we should resolve now, let me know

EDIT: a few notes, going over differences again:
- keyboard layout in the new hyprland implementation directly only watches the active keyboard and skips internal state of multiple layouts (that wasn't used so far) - so all good here
- in hyprland the submap state needs to be persisted via `internal_state` since the IPC only exposes the change via an event, not via a poll. (We could also seperate out specific messages (SubMapChanged) and dont couple it with the rest of the state - however either way is an assumption of where future features might go. Right now i am a fan of tight coupling to leave everything up to the compositor implementation - might change my mind with the next PR on making submaps work for third party tools :P)

Otherwise looking forward to see what I don't :)

MalpenZibo:
About 

> **Open decisions:**
> 
>     * [ ]  should the backend be something to be "failed over to" like it is right now and with network, or a config option?
> 
>     * [ ]  Readme, intro, website, doc adjustments - this is no longer just limited to hyprland :)
> 
>     * [ ]  debate empty / full init state of the service - right now niri immediately fills it, so would like to keep it empty
> 
>     * [ ]  changed the scrolling on worksapces to be in the right direction - was this a bug or intended? I assumee it was "if I scroll up I expect the workspace number to increase" while scrolling layouts (also on hyprland) scroll down with increasing workspace index. This will depend on how you configure workspace animation direction and might just be a config preference - or something to be handled on a compositor to compositor basis - what are your takes?

- It would be great to have some sort of env var to know which compositor is in use, but probably it's not so easy, so we could leave it as it is for now.
- Readme and other stuff can be done in a dedicated PR.
- I still need to see the code. Sorry, I haven't had enough free time recently :cry:   
- I don't know really XD I personally never use the scrolling feature. Maybe @EdgesFTW could help us.

MalpenZibo:
Another thing that I notice. With these changes, we create a subscription for each module that uses compositor information. I think that for now it could be ok, but probably it would be better to:
- Move the subscription to the app level and pass the data down to the module. (Pro: easy solution, Downside: breaks module isolation, still requires information that could be unnecessary (eg: someone doesn't use the workspace module). Not a big deal
- Make the `CompositorService` "dynamic", letting the subscriber specify only which data is required. (Pro: preserve module isolation, only required data are fetched, Downside: more complex solution, we still need to open a connection for each kind of data requested (more or less like it is right now)

clotodex:
> It would be great to have some sort of env var to know which compositor is in use, but probably it's not so easy, so we could leave it as it is for now.

uh env var is actually quite clean I think. Both niri and hyprland have env vars for their IPC
via: HYPRLAND_INSTANCE_SIGNATURE
and NIRI_SOCKET
So this could be a better way to choose than to hard fail over :D
We can extend this in the future if setting this with env vars outside of ashell makes sense (to support more complex setups or overrides like in #329 )

> Readme and other stuff can be done in a dedicated PR.

cool, happy to do some ripgrepping etc already though while waiting for the code review. Let me know if you want to keep it separate and agree on it, otherwise I can check that out as well

> I still need to see the code. Sorry, I haven't had enough free time recently 😢

~regarding the empty / full state - I am experiencing two bugs that might be related to this.~
~1. if I live-change the config it loses the connection to niri until I restart ashell => this is likely unrelated and potentially a general ashell issue (not bothering me though)~
~2. if I have multi-monitor setups it sometimes does not reload correctly when unplugging - this might also actually be a subscription issue - maybe you have more info on how multi-monitor sets things up~
~If neither are related, then empty state is fine -~ I am not experiencing any other issues
EDIT: it was event parsing (left a comment in the code) - the state was unrelated

> I don't know really XD I personally never use the scrolling feature. Maybe @EdgesFTW could help us.

Me either - but for niri the previous implementation is plain wrong - for hyprland it depends on interpretation and could technically be both - deferred to the compositor implementation and config-overridden. But yeah @EdgesFTW if you have some argument to weigh in, let me know. Worst case we can see if we get an issue after report to revert back - then we have more real world examples.

clotodex:
> Another thing that I notice. With these changes, we create a subscription for each module that uses compositor information. I think that for now it could be ok, but probably it would be better to:

Since they are all subscribing to the same compositor service this is actually just a single subscription (ice deduplicates them)

> - Move the subscription to the app level and pass the data down to the module. (Pro: easy solution, Downside: breaks module isolation, still requires information that could be unnecessary (eg: someone doesn't use the workspace module). Not a big deal

Not quite sure how this helps vs the current approach? Either you or I misunderstand something I think :D

> - Make the CompositorService "dynamic", letting the subscriber specify only which data is required. (Pro: preserve module isolation, only required data are fetched, Downside: more complex solution, we still need to open a connection for each kind of data requested (more or less like it is right now)

Was thinking about this too and in #329 there might be some extensions on this we might need to build since e.g. hyprland IPC does not expose as much as the wayland protocols - and yet the protocols dont have all the info and need some from wayland. So some composability based on what is needed might be relevant -> this could stay on service level though, not on subscription level.

With regards to "requesting what you need" the downside as you said is a subscription for each module. Technically they could be just an indirection to another subscription but yeah.. Potential upside for #329 is that different backends could be lazy-loaded depending on what is needed - e.g. no title, only workspaces? use wayland only. Title but no workspaces? use hypr. Both? use both or just hypr. Etc. However I would see this as a new PR based on #329 and @jazzpi's wayland protocol implementation.

My take on this for now:
- the IPC (both hypr and niri) exposes everything anyways so there is nothing gained from splitting subscriptions in any way
- if we care about separation of concerns then I would likely not adjust the subscription, but the messages => instead of communicating the whole state, have different messages for what each module needs to know (e.g. a submap message, a keyboard message, ...). Not quite sure about the up and downsides of this.

What do you think @MalpenZibo ?

MalpenZibo:
> Since they are all subscribing to the same compositor service, this is actually just a single subscription (ice deduplicates them)

Yes but I made a simple test, I placed a `println` here
```rust 
    macro_rules! add_refresh_handler {
        ($listener:expr, $method:ident, $output:expr) => {
            $listener.$method({
                let output = $output.clone();
                let internal_state = internal_state.clone();
                move |_| {
                    let output = output.clone();
                    let internal_state = internal_state.clone();
                    Box::pin(async move {
                        if let Ok(state_guard) = internal_state.read()
                            && let Ok(state) = fetch_full_state(&*state_guard)
                            && let Ok(mut o) = output.write()
                        {
                            println!("event hyprland");
                            let _ = o.try_send(ServiceEvent::Update(
                                CompositorEvent::StateChanged(state),
                            ));
                        }
                    })
                }
            });
        };
    }
```
And I notice that with two open windows, moving the mouse between the two causes two `"event hyprland"` log using both `workspaces` and `window_title` module, and only one using only the `window_title` module. So, there's something I don't understand. Could you try to see if it's something that happens also on your side?

clotodex:
> > Since they are all subscribing to the same compositor service, this is actually just a single subscription (ice deduplicates them)
> 
> Yes but I made a simple test, ...
> 
> And I notice that with two open windows, moving the mouse between the two causes two `"event hyprland"` log using both `workspaces` and `window_title` module, and only one using only the `window_title` module. So, there's something I don't understand. Could you try to see if it's something that happens also on your side?

It is actually not true that iced deduplicates the subscription - it might be a race condition though. In my logs:
```
WARN [ashell::services::compositor] Failed to listen to hyprland: Could not get socket path! (Is Hyprland running??)
WARN [ashell::services::compositor] Failed to listen to hyprland: Could not get socket path! (Is Hyprland running??)
WARN [ashell::services::compositor] Listening for niri instead
WARN [ashell::services::compositor] Listening for niri instead
```
Each should obviously only happen once. Will read up on iced a bit.

clotodex:
Okay, I was wrong, you are right. (And iced explanations are confusing as hell). The app level subscription is the straightforward solution - but I don't like that pattern generally for ashell. Since one might not want to use some modules, an "always" subscription is not great.
The only way I can see iced supporting (that is compatible with what we want) is an app level subscription with potentially custom logic that only enables the subscription if one of the modules is present.

MalpenZibo:
Let's try with a sub-optimal solution for now. 

I don't think that many users will have a configuration without a workspace module or a title module, so because the service, in any case, provides information for everything compositor-related, we could move that to the app level.

clotodex:
Alright so summarizing
- i will implement the env var failover to auto-detect what compositor is in use - instead of relying on errors
- we do the renaming/docs as a separate cleanup pr
- the init state of the service stays as is - no errors from few weeks of use
- scrolling stays as implemented until we hear back from @EdgesFTW 
- i will pull the subscription to app level (will check the fancyness when coding)
- any extension - wayland protocols, submaps, etc stay for future work

Other than that, any major annoyances you have with the code so far?

clotodex:
First feeling when looking at pulling the subscription to the app is that it feels very ugly :D
As far as I can tell subscriptions in iced are generally not designed for services. Starting a subscription multiple times should not cause problems as per best practice.
<details>
<summary>[outdated] thoughts before implemenation </summary>

Now diving a bit deeper into potential options we have and issues in iced, I arrived at a lazily spawned tokio spawn.
This should not have any performance penalty (tokio spawn, not read thread). It keeps everything clean. The only issue would come with cleanup - but ashell does not have this issues since modules cant just disappear

This is how it could look like - will test it out tomorrow:
```rust
struct IpcState {
    tx: Option<broadcast::Sender<IpcMessage>>,
}

static IPC_STATE: once_cell::sync::Lazy<Arc<Mutex<IpcState>>> = 
    once_cell::sync::Lazy::new(|| {
        Arc::new(Mutex::new(IpcState { tx: None }))
    });
 ...
...
subscription::channel(
        std::any::TypeId::of::<IpcConnection>(), // Use TypeId as unique ID
        100,
        |mut output| async move {
            let mut rx = {
                let mut state = IPC_STATE.lock();
                
                // If no sender exists, create connection and start work
                if state.tx.is_none() {
                    let (tx, _) = broadcast::channel(100);
                    state.tx = Some(tx.clone());
                    
                    // Spawn the actual work in background
                    tokio::spawn(async move {
                        let connection = connect_ipc().await;
                        loop {
                            match connection.recv().await {
                                Ok(msg) => {
                                    let state = IPC_STATE.lock();
                                    if let Some(tx) = &state.tx {
                                        let _ = tx.send(msg);
                                    }
                                }
                                Err(_) => break,
                            }
                        }
                    });
                }
                
                state.tx.as_ref().unwrap().subscribe()
            };
            
            // Forward messages from broadcast to this subscriber
            loop {
                match rx.recv().await {
                    Ok(msg) => {
                        let _ = output.send(msg).await;
                    }
                    Err(_) => break,
                }
            }}}
```
</details>

*EDIT*: @MalpenZibo please check out the new compositor/mod.rs
I am very happy with it as of now.
Instead of trying to shoehorn it into the iced approach - which is not designed for any kind of broadcasting pattern - I used the tokio broadcasting pattern. Then used iced subscriptions to subscribe to it multiple times.
This follows both, the best practices of iced AND tokio.
I like this approach and it feels relatively clean. In the future for other services we could even pull out / generify the broadcaster itself.

*EDIT 2*: fixed ~Currently buggy - still happy with it but found some interesting timing issue. On release my PC sometimes is fast enough with init, that only one subscriber receives the init event, while the second does not - resulting in empty workspaces, since the service is never initialized. Checking if this is an easy fix (replaying on subscription, different init mechanism, service passing on update, etc).~

clotodex:
Alright. The multi-consumer situation is a bit cursed. Currently the setup is like this:

module1 -> subscribes service -> service subscribes to broadcast -> niri/hypr start listening and manage state -> state passed back -> message triggered in module 1 -> module1 updates service instance -> service instance state changes
The problem: with a module2, only the internal listening is shared - state is managed a second time, init etc needs to be carefully managed.

This leads me to two (three) paths:
1. As @MalpenZibo suggested - pull everything to app, to specialized message passing - always subscribe to compositor IPC (which you likely always want) - manage all submodules via specialized messages managed in the service **main pro: simplicity and effeciency, main con: VERY tight coupling and putting the work to modules**
2. Keep the broadcasting idea, split the service. We could have a Workspaces Service, a Window Title Service, etc - all listening to the same IPC broadcast and constructing their states independently. **main pro: services manager their own responsibility, fully-declarative by modules, subscription and UI code separated nicely per module, main con: state duplication, potentially no real gain**
3. bonus: keep as-is. Broadcast system with per-module instantiations of CompositorService. This keeps 2 open for the future and works for now. Plus is pretty clean, i.e. if you check out any other service implementation this one behaves the same, while serving multiple modules

src/modules/workspaces.rs
```diff
@@ -163,161 +168,162 @@ fn get_workspaces(config: &WorkspacesModuleConfig) -> Vec<Workspace> {
...
+                iced::Task::none()
             }
             Message::Scroll(direction) => {
+                /* TODO: should we use the native service implementation instead?
+                if let Some(service) = &mut self.service {
+                    return service
+                        .command(CompositorCommand::ScrollWorkspace(direction))
+                        .map(Message::ServiceEvent);
+                }
+                return iced::Task::none();*/
```
MalpenZibo, line 292: using something like a simple `CompositorCommand::ScrollWorkspace(direction)` should be cleaner. Anyway, I would like to refactor this part because using the touchpad is too sensitive.

clotodex: Yeah that was a point i raised before but did not want to touch the existing code that actively chose to not use it
But i am with you here :) If someone wants custom behavior a future extension could be a simple config scroll direction

clotodex: Yeah that was a point i raised before but did not want to touch the existing code that actively chose to not use it
But i am with you here :) If someone wants custom behavior a future extension could be a simple config scroll direction

MalpenZibo: ok, we could use the ScrollWorkspace command but we should add a way to check if the scroll arrives from the touchpad and filter the sensitivity.

It could be done in a dedicated PR

MalpenZibo:
I notice that changing the configuration (the log_level in this case) breaks in some way the compositor service (I don't see the workspaces and window title)

clotodex:
> I notice that changing the configuration (the log_level in this case) breaks in some way the compositor service (I don't see the workspaces and window title)

Noted will check this out tomorrow

clotodex:
Ok I can reproduce this.
The service is actually working fine and the subscription is still yielding all events.
But somehow the modules are not resubscribing ... Don't know why that is happening.

clotodex:
Okay - My hunch after debugging is that iced does not rewire the new subscription - for some reason, idk why. But to the best of my knowledge this "issue" is either within iced or a misuse of iced (which btw is way to easy...).
What I am noticing is that currently in the codebase, existing services are not being recreated in app.rs on a config change, but just receive an update message with the new config instead. @MalpenZibo I assume there was a reason to this. I would go ahead and implement the same for the affected modules to fix this bug.

MalpenZibo:
Yes, the resubscription should happen if the id changes in my understanding but we are using typeid so maybe that's the problem

clotodex:
Check out the latest fix commit
I copied the approach you have in the other services
Works fine now for me

MalpenZibo:
Seems to be working fine! I will test it in the next few days, then I think that we could merge it

bbb651:
I tested it and unfortunately it seems to be broken on nvidia, only rendering the first frame and staying frozen and printing a bunch of `ERROR [iced_winit::program] Error Outdated when presenting surface.`, probably the same underlying issue as https://github.com/YaLTeR/niri/issues/2335.

I'm not sure there's much you can do, and unlike regular xdg-shell clients the user cannot workaround it, maybe as a hack ashell can resize itself right after it launches (only on niri + nvidia)?

MalpenZibo:
Ok, but is this kind of bug related to this PR?

I mean, this problem also happens with the main branch of `ashell`? Because it seems something related to `ashell/iced` is not running correctly under `niri/nvidia`, and not something related to the `ashell` service that retrieves the compositor data

MalpenZibo:
Let's go ahead with the merge. We could open a dedicated issue in the future.

PR #332, opened 2025-11-26
Allow hiding special workspaces

fdev31:
Controls if special workspaces shall be displayed using `enable_special_workspaces`

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -86,7 +86,6 @@ pub enum WorkspaceVisibilityMode {
 pub struct WorkspacesModuleConfig {
     pub visibility_mode: WorkspaceVisibilityMode,
     pub enable_workspace_filling: bool,
-    #[serde(default)]
     pub disable_special_workspaces: bool,
     pub max_workspaces: Option<u32>,
     pub workspace_names: Vec<String>,


MalpenZibo:
Hi, can you add an entry in the documentation?

Also, can I ask why this requirement?

fdev31:
I use special workspaces for:

- hidding/staging/minimizing apps
- scratchpads

I use always "direct access" for it, they have relatively long names and look like pollution on the widget. 

Just to be sure, you want me to update "modules/workspaces.md" in the website ? any additional file ?

fdev31:
Illustration of what I avoid using such patch:
image

vs a minimal / more expected:

image

MalpenZibo:
Ok thanks! 

Only one request. Could you flip the logic? A flag to hide special workspaces? 

To avoid introducing breaking changes in the configuration.

fdev31:
Sure! Should I keep "workspace" in the name of the configuration item by the way ? (it's in the "workspaces" module so I'm not so sure it makes sense, and different items seem to follow a different rule...

MalpenZibo:
Keep `workspaces`

addicust:
I was just about to ask for this, I use a special workspace as a dropdown terminal in hyprland and it doesnt look good seeing (Scratchpad) (1)(2)(3)

src/config.rs
```diff
@@ -86,6 +86,8 @@ pub enum WorkspaceVisibilityMode {
 pub struct WorkspacesModuleConfig {
     pub visibility_mode: WorkspaceVisibilityMode,
     pub enable_workspace_filling: bool,
+    #[serde(default)]
```
MalpenZibo: This shouldn't be necessary

PR #354, opened 2025-12-16
Feature: VPN service extensions for Tailscale support

zaolin:
Hey folks,

I am not sure if you wanna take that code, just generated it with agentic coding.
For me it works quite nice, so I thought it might be a good start.
I designed the module to be similar to the Trayscale software.
It still has some rough edges. I am going to work a bit more the next weeks on it.

If there is not interest to integrate the code that's also fine, just let me know.
Also if you have early feedback I appreciate it.

- Add disconnect/connect toggle with loading states
- Add exit node selection with 'None (This device)' option
- Add profile switching with auto-connect
- Show status indicator in bar (green/red dot)
- Smaller togglers for compact UI

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -375,6 +375,10 @@ dependencies = [
  "flexi_logger",
  "freedesktop-icons",
  "hex_color",
+ "http-body-util",
+ "hyper",
+ "hyper-util",
+ "hyperlocal",
  "hyprland",
  "iced",
  "inotify",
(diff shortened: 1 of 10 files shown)


MalpenZibo:
Hi, thanks for the contribution. 

Tailscale is a VPN? In that case, I prefer some sort of integration with the current VPN handler. The problem with these sorts of modules is that they are way too specific and increase the codebase and the maintenance effort.

For this reason, I don't know if it makes sense to have this like it is now in upstream.

zaolin:
Hi,

No problem I am going to do the rework the code. Maybe I can fit it into the VPN handler.
I am new to project and didn't know where to put it.

Tailscale is a quite specific VPN but I think we can probably somehow make it possible.
If it doesn't work out, I can keep it just in my fork. :)

Best Zaolin

MalpenZibo:
I can see 3 ways to integrate this into `ashell`
- In the VPN service
- as a custom module using external bash scripts to launch a command and get information
- as a tray element (only if a tailscale GUI app or service exists that provides a tray support)

zaolin:
> I can see 3 ways to integrate this into `ashell`
> 
> * In the VPN service
> * as a custom module using external bash scripts to launch a command and get information
> * as a tray element (only if a tailscale GUI app or service exists that provides a tray support)

I decided to go for the frist option. Going to push some updates the next days. The latest change already implements it as part of the VPN service as additional event-based backend. Now we have two NetworkManager and Tailscale backends. Not sure if this works out for you?

image

image

image

MalpenZibo:
Woh, ok, I didn't expect so much stuff. I was thinking more like a way to add the VPN to the normal list with its toggle button, but here we have a lot of options.

Right now, I don't think that I will merge it upstream. It seems more like something that should have its own GUI that maybe exposes a tray item

zaolin:
Ok thanks for letting me know, closing the PR then.

Issue #301, opened 2025-10-26
Add additional Clock config options


JettScythe:
I am in the process of moving over from waybar, and the only thing that this seems to be lacking for 1:1 functionality is the ability to configure multiple timezones in my clock that I can switch between by hovering + scrolling

fdev31:
I would like to be able to launch a calendar app when clicking on the Clock...

romanstingler:
@JettScythe I have pushed a PR you can define multiple datetime formats and change/cycle through them by clicking on the clock. 
https://github.com/MalpenZibo/ashell/pull/361

@fdev31 idk my waybar had click on it and change the formats and onHover you get the calendar and with right-click you can cycle/toggle between the 2 calendar layouts. 

I would suggest that you open a new feature request. and lets discuss that there

PR #361, opened 2025-12-19
Tempo: Add clock format cycling feature

romanstingler:
@MalpenZibo

This change extends the Tempo module with two features previously built for the deprecated Clock module:

Format cycling via right-click: Configure multiple datetime formats in the formats array. Right-clicking the clock cycles through them. The update interval adjusts automatically (1 second for formats with seconds, 5 seconds otherwise).
Timezone cycling via scroll: Configure timezone identifiers in the timezones array. Scrolling up/down cycles forward/backward. Supports both IANA names (e.g., "UTC", "America/New_York") and fixed offsets (e.g., "+00:00", "-05:00"). When the format contains `%Z`, named zones are preferred for abbreviations; otherwise numeric offsets are used. (chrono is a bit special here, sorry for the dependency but .... https://docs.rs/chrono/latest/chrono/format/strftime/index.html#fn6)

The implementation refactors the Tempo view logic to:
Detect `%Z` in the format string to choose between named zones and numeric offsets
Parse timezone strings using `chrono_tz::Tz` for IANA names and `FixedOffset` for offset strings
Apply the selected timezone to UTC time before formatting.

Documentation has been updated in `tempo.md` with examples for both features. 

closes https://github.com/MalpenZibo/ashell/issues/301

---
**Old**

Add support for multiple datetime formats in the clock module that can be cycled through by clicking. 

When the `formats` array is configured, clicking the clock cycles through each format in sequence. 
The update interval automatically adjusts based on the current format. 
If `formats` is empty or not provided, the clock uses the single `format` string as before (backward compatible).

I sometimes use different TZ sometimes I need with seconds and sometimes something else (like local time and UTC), 
so I can have all the formats that I need handy without the need for changing the config all the time. 

If the comments in the example doc should be removed just let me know.
As I said, thanks for the cool project.

### DIFF
--- a/src/modules/mod.rs
+++ b/src/modules/mod.rs
@@ -29,6 +29,12 @@ pub mod workspaces;
 pub enum OnModulePress {
     Action(Box<Message>),
     ToggleMenu(MenuType),
+    ToggleMenuWithExtra {
+        menu_type: MenuType,
+        on_right_press: Option<Box<Message>>,
+        on_scroll_up: Option<Box<Message>>,
+        on_scroll_down: Option<Box<Message>>,
+    },
 }
 
 impl App {
@@ -102,6 +108,26 @@ impl App {
                         button.on_press_with_position(move |button_ui_ref| {
                             Message::ToggleMenu(menu_type.clone(), id, button_ui_ref)
                         })
+                    }
+                    OnModulePress::ToggleMenuWithExtra {
+                        menu_type,
+                        on_right_press,
+                        on_scroll_up,
+                        on_scroll_down,
+                    } => {
+                        let mut button = button.on_press_with_position(move |button_ui_ref| {
+                            Message::ToggleMenu(menu_type.clone(), id, button_ui_ref)
+                        });
+                        if let Some(msg) = on_right_press {
+                            button = button.on_right_press(*msg);
+                        }
+                        if let Some(msg) = on_scroll_up {
+                            button = button.on_scroll_up(*msg);
+                        }
+                        if let Some(msg) = on_scroll_down {
+                            button = button.on_scroll_down(*msg);
+                        }
+                        button
                     }
                 }
                 .into()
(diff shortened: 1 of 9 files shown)


MalpenZibo:
Hey, thanks, it's a great addition.

The only thing is that I plan to replace the current Clock module with #279 after the next release, which I should create today or tomorrow.

What do you think if we postpone this feature in the next release, rebasing it on top of the new "Tempo" module?

romanstingler:
Hey sorry, I have some busy days at work :) 

absolutely fine for me. 
I will have time at night for the other issue(Overlay).

romanstingler:
image
image



image
image

src/modules/mod.rs
```diff
@@ -278,10 +278,29 @@ impl App {
...
+                    Message::ToggleMenu(MenuType::Tempo, id, button_ui_ref)
+                })
+                .on_right_press(Message::Tempo(tempo::Message::CycleFormat))
+                .on_scroll_up(Message::Tempo(tempo::Message::CycleTimezone(
+                    tempo::TimezoneDirection::Forward,
+                )))
+                .on_scroll_down(Message::Tempo(tempo::Message::CycleTimezone(
+                    tempo::TimezoneDirection::Backward,
+                )));
+                Some((button.into(), None))
```
MalpenZibo: Let's extend the `OnModulePress` instead of this. This is a sort of singleton that breaks the consistency with how we treat the otehr module

src/widgets/position_button.rs
```diff
@@ -221,6 +271,15 @@ where
                     }
                 }
             }
+            Event::Mouse(mouse::Event::ButtonPressed(mouse::Button::Right)) => {
```
MalpenZibo, line 275: Here, maybe we should use `ButtonRelease` instead of Press. also we could use publish_on_`release` method also for the left click

src/modules/tempo.rs
```diff
@@ -77,15 +93,81 @@ impl Tempo {
...
+                    }
+
+                    if let Ok(offset) = tz_name.parse::<FixedOffset>() {
+                        return Some(
+                            offset
+                                .from_utc_datetime(&utc_now.naive_utc())
+                                .format(format)
+                                .to_string(),
+                        );
+                    }
```
MalpenZibo: Here we could try `FixedOffset` first if %Z isn't in format and use Tz as fallback. To simplify the code because branch 1 and 3 are identical

website/docs/configuration/modules/tempo.md
```diff
@@ -20,6 +20,8 @@ Tempo combines a highly configurable clock with an optional weather summary in t
 | Option             | Type     | Default       | Description                                                                                                                                                                                           |
 | ------------------ | -------- | ------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | `clock_format`     | `string` | `%a %d %b %R` | Strftime-compatible format used for the clock in the bar and in the menu header. See the [chrono formatting guide](https://docs.rs/chrono/latest/chrono/format/strftime/index.html) for placeholders. |
+| `formats`          | `array`  | `[]`          | Multiple datetime formats that can be cycled through by clicking the clock. When provided, clicking cycles through each format in sequence.                                                           |
```
MalpenZibo:
```suggestion
| `formats`          | `array`  | `[]`          | Multiple datetime formats that can be cycled through by right-clicking the clock. When provided, clicking cycles through each format in sequence.                                                           |
```

romanstingler:
Hey I cleaned up the branch and pushed the changes. 

Thank you for reviewing. 

> `The update interval adjusts automatically` I cannot see any changes to the subscription method, so this is true?
Missed that in one of my stashes.

PR #362, opened 2025-12-19
add configurable Wayland layer support 

### LINKED ISSUE #356: [FEATURE REQUEST] add the ability to keep the bar visible over fullscreen programs
waybar supports this by putting layer = overlay, and is the only reason I don't use ashell on my laptop

romanstingler:
Add `layer` configuration option to control which Wayland layer the bar is rendered on. 

Supports 
`Bottom` (default), 
`Background`, 
`Top`, and 
`Overlay` layers. 

The layer setting is now threaded through the app initialization, config reloading, and output management.

A bit of cleanup, remove the `.clone()` because we are using the `Copy` Trait 

closes #356

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -44,6 +44,7 @@ use wayland_client::protocol::wl_output::WlOutput;
 pub struct GeneralConfig {
     outputs: config::Outputs,
     pub modules: Modules,
+    pub layer: config::Layer,
     enable_esc_key: bool,
 }
 
@@ -101,6 +102,7 @@ impl App {
             let (outputs, task) = Outputs::new(
                 config.appearance.style,
                 config.position,
+                config.layer,
                 config.appearance.scale_factor,
             );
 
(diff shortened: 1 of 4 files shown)


MalpenZibo:
Ok, I don't get why someone would have the status bar on top of a full-screen window. 

Anyway it make sense to have all this configuration? I think that the `Background` and `Top` layers are not so useful in this context. What do you think?

romanstingler:
I added all just to have a full list, but I see why some of them might be questionable. 
I pushed a change, just 'Bottom' and 'Overlay' . 

I have no clue why the user needs it, 
but I would guess there are cases where you would want to have the bar as a HUD or during Streaming/Presenting you want to show something. 

Keep time or system status or media-player 'now playing' during streams or full-screen videos or presentations (maybe a timer) so. 
Or as a business on a TV where you want to have the bar with a logo or a custom module. 
...

MalpenZibo (approved):

PR #364, opened 2025-12-20
Fix the reported SystemBattery percentage.

kiryl:
When there are multiple batteries in the system, the percentage calculation overvalues the contribution of the smaller battery.i

This can lead to very strange results in extreme cases. For example, the Thinkpad T480 reports two batteries, but the second one is not real, with zero Energy and zero EnergyFull. With the current percentage calculation, the reported value ends up being half of the actual value.

To derive the SystemBattery percentage accurately, it should be based on the ratio of total Energy to total FullEnergy.


kiryl:
Updated.

PR #370, opened 2025-12-28
Fix VPN list when there are too many VPNs

kushagharahi:
When there are too many VPNs, the list doesn't render anything (I have about 200):
image

With this fix, it makes it into a scroll-able list and renders the VPNs. Sorted alphabetically
image

### DIFF
--- a/src/modules/settings/network.rs
+++ b/src/modules/settings/network.rs
@@ -546,7 +546,8 @@ impl NetworkSettings {
         show_more_button: bool,
     ) -> Element<'a, Message> {
         // Collect and sort the VPNs
-        let mut vpns: Vec<_> = service.known_connections
+        let mut vpns: Vec<_> = service
+            .known_connections
             .iter()
             .filter_map(|c| match c {
                 KnownConnection::Vpn(vpn) => Some(vpn),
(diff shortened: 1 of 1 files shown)


src/modules/settings/network.rs
```diff
@@ -545,13 +545,21 @@ impl NetworkSettings {
...
         show_more_button: bool,
     ) -> Element<'a, Message> {
-        let main = Column::with_children(
-            service.known_connections
-                .iter()
-                .filter_map(|c| match c {
-                    KnownConnection::Vpn(vpn) => Some(vpn),
-                    _ => None,
-                })
+        // 1. Collect and sort the VPNs first
```
MalpenZibo: let's remove `1.` and `2.` from the comments 😄

src/modules/settings/network.rs
```diff
@@ -569,6 +577,11 @@ impl NetworkSettings {
         )
         .spacing(theme.space.xs);
 
+        let main = container(
+            scrollable(vpn_list))
+            .height(Length::Shrink)
+            .max_height(500.0);
```
MalpenZibo: Could you also reduce the max height? We had to add some padding from the toggle and the scrollbar. It's quite tricky, see the update module for an example

src/modules/settings/network.rs
kushagharahi: Done, sorry about that

src/modules/settings/network.rs
kushagharahi: made the max height 300 to match updates.rs and added the same scrollbar padding, here's what it looks like now
image

PR #308, opened 2025-11-05
Feature: Mouse Scrolling

EdgesFTW:
Added some basic mouse scrolling functionality for audio sliders and workspaces.

This works by wrapping each of the scrolling areas in an iced MouseArea widget.

Addresses #268 and #304.

PR #374, opened 2025-12-29
add scroll support for brightness slider and fix UI sync issues

romanstingler:
The brightness slider currently suffers from stuttering and jumping caused by conflicting updates between the UI and hardware, often stemming from rounding discrepancies, external brightness changes, and interference from hardware feedback loops. 

To resolve this and prevent UI flicker, the system will now track user interaction via an `is_user_adjusting` flag and a dedicated `ui_percentage` field that remains independent of hardware values during active input. 

While the user is interacting, including using new mouse wheel support for ±5% increments(related to #308), hardware updates will be ignored and a 500ms debounced reset timer will be used to safely re-enable hardware synchronization only after the interaction ends. 

This separation of UI state from hardware feedback ensures smooth slider behavior and eliminates the unexpected jumping caused by conflicting update cycles.

PS: I wish I could made it simpler. I spent some time trying to iron out the UI and HW value rounding differences but the slider still jumped around while dragging. 
This is the smoothest and visually most pleasing version that I could come up with.

### DIFF
--- a/src/modules/settings/brightness.rs
+++ b/src/modules/settings/brightness.rs
@@ -8,7 +8,8 @@ use crate::{
 };
 use iced::{
     Alignment, Element, Length, Subscription, Task,
-    widget::{container, row, slider},
+    futures::stream,
+    widget::{MouseArea, container, row, slider},
 };
 
 #[derive(Debug, Clone)]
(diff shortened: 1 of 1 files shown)


MalpenZibo:
> PS: I wish I could made it simpler. I spent some time trying to iron out the UI and HW value rounding differences but the slider still jumped around while dragging.
This is the smoothest and visually most pleasing version that I could come up with.

I know I tried in the past without much luck :sweat_smile:

PR #392, opened 2026-01-03
Fix workspaces on multi-monitor Niri setups

jazzpi:
Niri reuses workspace indices for multiple monitors. E.g. I currently have:

```sh
$ niri msg workspaces
Output "DP-1":
   1
 * 2
   3
   4
   5

Output "DP-2":
   1
   2
 * 3
   4

Output "HDMI-A-1":
   1
 * 2
   3
```

The current Niri implementation uses the workspace index as the ID. Since `modules::workspaces::calculate_ui_workspaces()` expects workspace IDs to be unique, using the workspace index as the ID meant I only saw one of each workspace (i.e. I only saw 5 workspaces in total instead of 12).

This PR

- uses the workspace ID as the ID on Niri
- sets the workspace name to the workspace index on Niri (unless a name is set already)
- displays the workspace name if it is set (i.e. the workspace index on Niri)
- moves the workspace sorting into the compositor backends instead of the UI

I hope I caught all uses of the workspace index. @clotodex could you double check please? :point_right: :point_left:

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -85,6 +85,7 @@ pub enum WorkspaceVisibilityMode {
 #[serde(default)]
 pub struct WorkspacesModuleConfig {
     pub visibility_mode: WorkspaceVisibilityMode,
+    pub group_by_monitor: bool,
     pub enable_workspace_filling: bool,
     pub disable_special_workspaces: bool,
     pub max_workspaces: Option<u32>,
--- a/src/modules/workspaces.rs
+++ b/src/modules/workspaces.rs
@@ -212,6 +212,10 @@ fn calculate_ui_workspaces(
         }
     }
 
+    if config.group_by_monitor {
+        result.sort_by_key(|w| w.monitor.clone());
+    }
+
     result
 }
 
--- a/src/services/compositor/niri.rs
+++ b/src/services/compositor/niri.rs
@@ -185,7 +185,7 @@ fn map_state(niri: &EventStreamState) -> CompositorState {
         .workspaces
         .workspaces
         .values()
-        .sorted_by_key(|w| (w.output.clone().unwrap_or_default(), w.idx))
+        .sorted_by_key(|w| w.idx)
         .map(|w| {
             CompositorWorkspace {
                 id: w.id as i32,


clotodex:
Hey @jazzpi, nice catch and thanks for the first stab at this!
I just tested and it works, however I have two nit-picks:
1. when connecting a new screen, the new workspaces dont show up until i open a window on the other screen (this could be a niri thing, is not too annoying and i could live without, but noticed it)
2. I really dont like the screens being "interweaved" like that and would argue to put the screens in sequence
so instead of 1 1 2 2 3 3 ... we do 1 2 3 1 2 3. I would really like this - what do you think?

jazzpi:
1. Do you mean the extra empty workspaces? That seems to be a Niri thing, it only spawns one empty workspace until you open something on the monitor (looking at the overview and `niri msg workspaces`). Or do you mean not even the one empty workspace shows up?
2. Good point, I use `visibility_mode = "MonitorSpecific"` so I didn't notice that. I've modified the sort so it sorts by monitor first and then workspace index.

clotodex:
Nice, I like it now: 
image

and to 1. I think you are right with that being what I saw - i thought i saw none at all, but now cant reproduce

So LGTM
@MalpenZibo could you (review and) merge this?

jazzpi:
I've made the group-by-monitor a config option for the workspace module (so it also works for Hyprland).

Hyprland:

swappy-20260105_205307

Niri:

Screenshot from 2026-01-05 20-58-21

(top: no `group_by_monitor`, bottom: `group_by_monitor = true`)

romanstingler:
@MalpenZibo  I added that in https://github.com/MalpenZibo/ashell/pull/423

MalpenZibo:
Ok this PR causes this behavior in my hyprland config 
image

We need to check this out

src/services/compositor/hyprland.rs
```diff
@@ -142,6 +143,7 @@ pub async fn run_listener(tx: &broadcast::Sender<ServiceEvent<CompositorService>
 fn fetch_full_state(internal_state: &HyprInternalState) -> Result<CompositorState> {
     let workspaces = Workspaces::get()?
         .into_iter()
+        .sorted_by_key(|w| w.id)
```
clotodex, line 146: @MalpenZibo what happens if you drop that line on your end? This is the only hyprland specifc change and looking at the code only this is my guess. 

In workspaces.rs it sorts again. So i thought this would not change anything when not grouping
But please try that out

clotodex: Nvmd just saw the issue
Github app hid that from me :P

clotodex: Nvmd just saw the issue
Github app hid that from me :P

Issue #229, opened 2025-09-07
Tuxedo Control Center icon not shown in tray


zimbipw:
The app icon for the Tuxedo Control Center doesn't show up in the tray, instead there is just a dot. Tray seems to work otherwise as Signal app tray icon shows up as expected.
Arch Linux
Hyprland 0.50.1
Ashell 0.5.0 from AUR

MalpenZibo:
Can you post here the log (when the problem happens) using this config
```
log_level = "ashell=warn, ashell::services::tray=debug"
```

zimbipw:
Am Freitag, dem 26.09.2025 um 05:13 -0700 schrieb Simone Camito:
> "ashell=warn, ashell::services::tray=debug"
there you are

MalpenZibo:
Hem... I don't get it. Where is the log?

Usually are under fhe `/tmp/ashell` folder

Modificator:
same problem occurs with JetBrains Toolbox

romanstingler:
I pushed a fix, the issue is that `tuxedo-control-center` sends `tuxedo-control-center1_1` as icon_name but the icon is named `tuxedo-control-center` in the system. 

<img width="279" height="27" alt="Image" src="https://github.com/user-attachments/assets/498cfd0a-f07f-44e1-b60d-d2e72cf655a2" />

@Modificator in which DE/WM do you have a tray icon for JetBrains Toolbox? 
I don't get anything in ashell via dbus and there is none in waybar.

PR #407, opened 2026-01-11
[Tray: 1/2] Improve tray icon lookup with fuzzy matching and system icon indexing

romanstingler:
Add once_cell and walkdir dependencies. 
Implement system-wide icon name indexing at startup, fuzzy matching for similar icon names, and prefix-based fallback matching to improve icon resolution for system tray applications.

closes https://github.com/MalpenZibo/ashell/issues/229

The issue is that `tuxedo-control-center` is the name of the icon in the system but dbus 
```
$ dbus-send --session --print-reply --dest=:1.89 /org/ayatana/NotificationItem/tuxedo_control_center1 org.freedesktop.DBus.Properties.Get string:"org.kde.StatusNotifierItem" string:"IconName"

method return time=1768084199.379954 sender=:1.89 -> destination=:1.109 serial=57 reply_serial=2
   variant       string "tuxedo-control-center1_1"
   ```


MalpenZibo:
Could we find a way to avoid at least the walkdir dependency?

romanstingler:
removed walkdir

Issue #99, opened 2025-02-12
Tray apps are not reinitialized after ashell kill and restart


Gigas002:
When you have a system running with some apps in tray, killing the ashell process and restarting it results in empty tray, though tray apps are still up and running

MalpenZibo:
Hum... I'm not able to replicate the issue. I tried with slack and vlc and in both cases, if I kill ashell with `killall ashell` and then restart the process the tray icon reappears.

Which ashell version and OS are you using?

Gigas002:
I am on arch linux and using `hyprland` as compositor. `ashell` installed through aur, version `0.4.0`. In some rare cases tray icons reappears. Sometimes (mostly) not, or not every app reappear.
I had the same experience with `waybar`, actually, don't know if it's related to something more deep, like some wayland protocol implementation itself.

P.S.
I'm starting to notice a pattern here. Let's say, I have a `telegram` running on a 2nd monitor. When I restart the `ashell` with 2nd monitor being active at the moment - telegram reappears in tray. Otherwise - not.

MalpenZibo:
I will try to replicate the issue on my machine to try to understand the cause of the problem.

MalpenZibo:
I'm not experiencing any problem with Slack. Sadly, the status of the tray dbus services on Linux is not great. My implementation should be similar to that of Waybar, so if the problem also appears there, I probably can do much.

In general, the shell module creates a "server" for the "org.kde.StatusNotifierWatcher" dbus. The "server" maintains a list of apps that send "tray data". So when you kill `ashell` and restart the process, this list is empty. In theory, the first time that an application sends "tray data" the tray icon should reappear.

You could try this test:
- start telegram
- the tray icon should appear on ashell
- kill ashell: `killall ashell`
- restart ashell
- if the telegram icon does not appear
  - completely kill all telegram task -> killall telegram
  - restart telegram
  - the tray icon on ashell should now appear

Gigas002:
Yeah, I understand that, just didn't see which implementation for try stuff you have created and also how `waybar` implements it. If this is on `dbus` side, then there's not much we can do, I suppose. I wonder if [busd](https://github.com/dbus2/busd) will become a better alternative in the future though.
Thank you for checking this out :)

MalpenZibo:
I mean. 
With the ashell aur version you could try the following test: 
```
    - start telegram
    - the tray icon should appear on ashell
    - kill ashell: killall ashell
    - restart ashell
[...]

PR #408, opened 2026-01-11
[Tray: 2/2] Implement proactive StatusNotifierItem discovery

romanstingler:
This PR addressed the issue where tray icons would not reappear after ashell was restarted.

Problem: ashell acted as a passive StatusNotifierWatcher, relying entirely on applications to call RegisterStatusNotifierItem. 
When ashell restarted, it lost its internal state, and many applications did not automatically re-register with the new watcher instance.

Implemented a proactive discovery mechanism in StatusNotifierWatcher that:
**Initial Scan**: Scans the D-Bus session bus upon startup to find existing StatusNotifierItem providers and registers them automatically.
**Periodic Refresh**: Spawns a background task that re-scans the bus every 30 seconds to ensure the tray stays in sync with running applications.
**D-Bus Name Filtering**: Identifies potential items by looking for bus names containing StatusNotifierItem.
**Registration Logic**: Refactored item registration to handle both manual discovery and standard D-Bus registration calls while preventing duplicates.

closes https://github.com/MalpenZibo/ashell/issues/99

### DIFF
--- a/src/services/tray/dbus.rs
+++ b/src/services/tray/dbus.rs
@@ -1,5 +1,6 @@
 use iced::futures::StreamExt;
 use log::{info, warn};
+use std::time::Duration;
 use zbus::{
     Connection, Result,
     fdo::{DBusProxy, RequestNameFlags, RequestNameReply},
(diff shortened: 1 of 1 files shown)


MalpenZibo (approved):
🚀

Issue #416, opened 2026-01-13
[Feature] Togglable visibility of the bar


samheiden:
To toggle visibility, one could just kill the process and spawn a new one. But this isn't ideal for certain use cases. Here's an example of how Waybar handles this:

https://github.com/Alexays/Waybar/wiki/FAQ#how-can-I-temporarily-hide-the-bars

PR #417, opened 2026-01-13
Add SIGUSR1 signal handling for visibility toggle

romanstingler:
Adds support for toggling ashell bar visibility using `killall -SIGUSR1 ashell`, similar to how Waybar handles it.
 
Usage:
```bash
killall -SIGUSR1 ashell
```

closes https://github.com/MalpenZibo/ashell/issues/416

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,7 +39,10 @@ iced = { git = "https://github.com/MalpenZibo/iced", rev = "daccb115c0827bd67fb1
   "svg",
   "canvas",
 ] }
-chrono = { version = "0.4", default-features = false, features = ["clock", "serde"] }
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -70,6 +70,10 @@ inotify = "0.11.0"
 pin-project-lite = "0.2.16"
 niri-ipc = "25.11.0"
 parking_lot = "0.12.5"
+signal-hook = "0.4.1"
+signal-hook-tokio = { version = "0.4", features = ["futures-v0_3"] }
+futures = "0.3"
```
MalpenZibo: hum... the iced library doesn't re-export futures?

romanstingler: sorry missed that

Issue #376, opened 2025-12-31
[Feature request] Microphone mute status


hecto600:
Same as audio sound menu, would love to have it for microphone.

romanstingler:
I could add that. 
Probably like the audio icon in the bar just with a mic icon, 
and a slider for mic volume and the mic mute button and the source select for mics if there are multiple. 

is that correct/what you want ??

hecto600:
Yes, that sounds good to me

romanstingler:
I pushed some code. 
you can build the version yourself or wait until the owner of the project has reviewed it and pushed a new version. 

<img width="364" height="500" alt="Image" src="https://github.com/user-attachments/assets/bed32e3c-55f5-45c2-b4a4-082c44e15261" />

hecto600:
Thanks! Now I can tell whether I am speaking with the microphone muted or not. 🙂

MalpenZibo:
I'm back, I will do my best to review all the PR in the next weeks 😅

PR #419, opened 2026-01-13
[Audio 1/2] add microphone indicator to settings module

romanstingler:
- Add Microphone to SettingsIndicator enum
- Implement source_indicator() method for microphone status
- Add Sources trait with get_icon() for microphone icon selection
- Update default indicators order to include Microphone after Audio
- Show microphone selection menu if multiple Microphones are detected
- Update documentation with Microphone indicator examples

closes https://github.com/MalpenZibo/ashell/issues/376

@MalpenZibo
maybe you find some time to discuss the audio/sound volume icon.
```
Some((true, _)) => StaticIcon::Speaker0,  // Explicitly muted
Some((false, volume)) => {
    if volume > 0.66 { StaticIcon::Speaker3 }
    else if volume > 0.33 { StaticIcon::Speaker2 }
    else if volume > 0.000001 { StaticIcon::Speaker1 }
    else { StaticIcon::Speaker0 }  // Volume at 0 shows mute icon
}
```
when I click the mute button in the menu you can scroll the mouse as much as you want it stays muted (mute is a separate state).

but when I have some volume and scroll down and make it more silent and we get to 0 volume,
you show the mute icon in the bar, but inside the menu it is not muted, because mute is not 0 volume.

In my opinion the mute icon at 0 volume is wrong, because that indicates that no matter what I press (volUpBtn on the keyboard or so) the volume will stay muted.


MalpenZibo:
Yes, you're probably right. It's better to use the mute icon only when the mic is actually muted.

I suppose that we should also change the speaker icon behavior right?

romanstingler:
@MalpenZibo today I have no more time, I will work tomorrow on some stuff

romanstingler:
ready, I made a separate PR for the audio icon

PR #430, opened 2026-01-16
Add support for initialTitle and initialClass for Window Title Mode

lkcv:
Adds InitialTitle and InitialClass as options under [window_title] -> mode

Niri doesn't expose these via IPC (from what I could see), but I added them anyway with a comment above (says that we're just using title and class for these) rather than creating a new ActiveWindow struct for niri. 

let me know if this approach is preferred and we can use something like ActiveWindowHyprland and ActiveWindowNiri.

Tested on Hyprland.

### DIFF
--- a/src/services/compositor/hyprland.rs
+++ b/src/services/compositor/hyprland.rs
@@ -73,7 +73,9 @@ pub async fn run_listener(tx: &broadcast::Sender<ServiceEvent<CompositorService>
 
         match fetch_full_state(&state_guard) {
             Ok(state) => {
-                let _ = tx.send(ServiceEvent::Update(CompositorEvent::StateChanged(Box::new(state))));
+                let _ = tx.send(ServiceEvent::Update(CompositorEvent::StateChanged(
+                    Box::new(state),
+                )));
             }
             Err(e) => {
                 log::error!("Failed to fetch initial compositor state: {}", e);
@@ -95,8 +97,9 @@ pub async fn run_listener(tx: &broadcast::Sender<ServiceEvent<CompositorService>
                         if let Ok(state_guard) = internal_state.read()
                             && let Ok(state) = fetch_full_state(&*state_guard)
                         {
-                            let _ =
-                                tx.send(ServiceEvent::Update(CompositorEvent::StateChanged(Box::new(state))));
+                            let _ = tx.send(ServiceEvent::Update(CompositorEvent::StateChanged(
+                                Box::new(state),
+                            )));
                         }
                     })
                 }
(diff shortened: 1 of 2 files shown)


MalpenZibo:
Ok but can you add the config documentation section about this changes? Specifying that using niri this config don't do anything?

MalpenZibo:
@clotodex what do you think?

clotodex:
@ikcv if i understand correctly this is a hyprland feature keeping thw first title the window had (so e.g. in terminal not updating to the newest command) correct?

@MalpenZibo i have a few thoughts
- runs on my machine, i dont personally dont need this use case, but that does not mean others might not
- if we want to maintain this feature for niri we could just cache the initial window title and introspect the events (right now the implementaton just wraps around niris crate, so the actual event parsing is abstracted into state) - tldr: i think can be coded into ashell, do you want this? What is your take on maintenance?
- since this is not a niri "intended" feature i would probably prefer an error or something similar explaining this feature not being supported on niri instead of it behaving unexpected (changing title even though setting for initial title is on) - but is a log message enough? How do we want to treat "not supported" errors?

Sorry for blowing this up :P have no strong opinion but this is what i think. Out of a) not implement at all b) implement spoof for niri or c) fail somewhat gracefully i would like c

lkcv:
@clotodex,

That's right.

For your suggestion, I can create a new ActiveWindow struct for Niri that doesn't have these 2 new fields. That way, it'll print an error if a user tries to use these 2 for niri.

lkcv:
I installed Niri to test this and it appears to work as expected. It'll print a warning log message if InitialTitle and InitialClass are used under a Niri session.

Documentation has been updated too.

On a related note, it looks like on niri that ashell only shows the title or class of the active window upon startup. It doesn't dynamically update afterwards. It does this on both my branch and on main.

MalpenZibo:
@clotodex @lkcv maybe for this niri issue let's open a separate PR. In the meantime, we could proceed with merging this

PS: format issue :smile:

lkcv:
@MalpenZibo i ran cargo fmt and pushed another commit, so hopefully the formatting issues are fixed now.
Not sure why CI is unhappy. I made a test branch based on main and merged my changes into that. It compiled.

MalpenZibo:
still some issues

MalpenZibo:
again! Run clippy on your local machine to check if everything is ok

lkcv:
Sorry, I was just running `cargo clippy` instead of `cargo clippy --all-features -- -D warnings`.
I made its suggested changes and am no longer getting any errors.

MalpenZibo:
Format issue

lkcv:
Formatting issues fixed. I looked at the workflow file since all these checks don't auto run, and both clippy and fmt pass now

PR #433, opened 2026-01-18
Add indicator_format option to MediaPlayer module

gwiazdorrr:
Adds a new `indicator_format` configuration option to the MediaPlayer module. Users can choose from `IconAndTitle` (default) and `Icon`.

My motivation was I don't actually want the title on my status bar, but I want to have a convenient way of controlling media that's played. Currently the only option is to set `max_title_length` to something very small, but even then you get three dots in the status bar and it also affects the panel being shown on click, which sucks.

Example config:
```
[media_player]
indicator_format = "Icon"
```

Tested on hyprland and niri.

### DIFF
--- a/src/modules/media_player.rs
+++ b/src/modules/media_player.rs
@@ -166,8 +166,8 @@ impl MediaPlayer {
     pub fn view(&'_ self, theme: &AshellTheme) -> Option<Element<'_, Message>> {
         self.service.as_ref().and_then(|s| {
             s.first().map(|player| {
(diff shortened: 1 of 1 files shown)


romanstingler:
@MalpenZibo 
Looks good to me 
I added some docs

MalpenZibo:
@gwiazdorrr the PR it's ok. Could you just fix the formatting?

gwiazdorrr:
@MalpenZibo no problem, done! I'll make sure pass PRs through `cargo fmt` in the future.

PR #449, opened 2026-01-26
AudioService improvements

levitatingpineapple:
### [Smart Filters](https://pipewire.pages.freedesktop.org/wireplumber/policies/smart_filters.html) Support

Smart filters are virtual sources and sinks which provide ability to use system-wide audio effects like EQ, background noise cancellation, feedback prevention etc. Most of the time these filter nodes will have no ports [^1]. According to pipewire docs smart filters can be recognized by having their `node.link-group` property being set. This PR adds ability to select such filters as default audio route from ashell.

### Volume Slider Jitter

Currently we update volume slider twice per each user input:
1. The cached volume gets [scaled locally](https://github.com/MalpenZibo/ashell/blob/485ff80f3d471033f984784697c1a60863d9114e/src/services/audio.rs#L87)
2. Then we call `libpulse::(...)::set_source_volume` which causes sink list to be updated.

Since the second update can be slightly delayed (for BT headsets for example) - pulseaudio events can start arriving after the next local update, creating jittery appearance. 

The solution is to remove the local update and create a copy of channel volumes to send to pulse audio. This ensures a unidirectional data flow and keeps the libpulse as a single source of truth.

### `AudioService` Interface

Due to inherent complexity of selecting audio routes like multi-port or virtual sinks/sources - I've made the actual sink/source arrays private and exposed methods to access them in a coherent way. Removes some code duplication too.

### Preview

The top bar is `main` and the botton one is on this branch. Notice that the volume slider no longer "jumps around" and audio selectors are now working.

https://github.com/user-attachments/assets/0a06a7e2-8344-43ae-bdd2-10b481370733

[^1]: There is a single default port in case of blue-tooth headphones, where the filter is used for profile switching

### DIFF
--- a/src/modules/settings/audio.rs
+++ b/src/modules/settings/audio.rs
@@ -315,7 +315,6 @@ impl AudioSettings {
                     }
                 }
             })
-        // })
     }
 
     pub fn sliders<'a>(
--- a/src/services/audio.rs
+++ b/src/services/audio.rs
@@ -19,7 +19,7 @@ use libpulse_binding::{
     proplist::{Proplist, properties::APPLICATION_NAME},
     volume::ChannelVolumes,
 };
-use log::{debug, error, trace};
+use log::{debug, error, trace, warn};
 use std::{
     any::TypeId,
     cell::RefCell,
@@ -285,7 +285,7 @@ impl AudioService {
             if device.is_filter {
                 if device.ports.len() > 1 {
                     let device_name = device.name.as_str();
-                    error!("Unexpected multiple ports in a filter node: {device_name}")
+                    warn!("Unexpected multiple ports in a filter node: {device_name}")
                 }
                 Either::Left(std::iter::once(Route { device, port: None }))
             } else {


denqxotl (approved):

denqxotl:
@MalpenZibo would u mind to check and merge this PR, please?

src/modules/settings/audio.rs
```diff
@@ -104,11 +104,11 @@ impl AudioSettings {
                     if let Some(service) = self.service.as_mut() {
                         service.update(data);
 
-                        if service.sinks.len() < 2 {
+                        if service.has_multiple_sinks() {
```
romanstingler: these 2 blocks have exactly inverted logic to the original one

romanstingler:
Filters seem to work 
xdph_screenshot_387113a6

I also like the refactoring. 
The only thing is (but we can fix that another day) is the Volume jitter.
This is not working. 
While I refactored the brightness slider, I tried so many things and I don't like the solution, 
but the best that I came up with was this one https://github.com/MalpenZibo/ashell/pull/374

But the brightness slider was jumping and was not just sluggish as the audio/mic sliders are. 
If you drag the audio/mic fast left to right they always get lost in an increasing delay. 

Just try the brightness slider for comparison.

Maybe at some point we can find an easy fix for that. 

We could merge that and open an issue for the sliders
and then debug the [UI position in %, read volume in %]
and see what happens during rendering. 

In cosmic the rendering of the brightness slider in the submenu is also a little bit sluggish,
not extremely but still slower than the settings menu with the audio sliders.

3bnz:
> But the brightness slider was jumping and was not just sluggish as the audio/mic sliders are.
> If you drag the audio/mic fast left to right they always get lost in an increasing delay.

The audio/mic sliders also had the jumping around issue - maybe harder to reproduce. It can be seen in the video 00:01-00:04 seconds.

But yes - I also wanted to propose a separate PR to make the sliders responsive. What you describe in https://github.com/MalpenZibo/ashell/pull/374 sounds good - I have done something similar for IOT sliders in a mobile app. Basically show local value on touch/finger down and then animate/jump to _source of truth_ after the interaction ends.

I suspect that calls to `pa_context_set_[sink|source]_volume_by_name`  should also be throttled. They are indeed _piling up_ causing the ever increasing delay. Either by time or by checking if the new audio state has been received.

src/modules/settings/audio.rs
3bnz: Patched - not sure how I missed this 🤦

romanstingler:
sounds good, maybe open a separate ticket and mention that you will work on it,
so people find the details via the issue tracker and don't implement it in parallel. 

@MalpenZibo for me the PR is good for now

MalpenZibo:
Sorry for the delay. I will look the code tomorrow and if everything it's good we can proceed with the merge

src/modules/settings/audio.rs
```diff
@@ -390,15 +404,19 @@ impl AudioSettings {
...
-                            active: p.active && s.name == service.server_info.default_source,
-                            msg: Message::DefaultSourceChanged(s.name.clone(), p.name.clone()),
-                        })
+                    .source_iter()
+                    .map(|route| SubmenuEntry {
+                        name: route.to_string(),
+                        device: route
+                            .port
+                            .map(|p| p.device_type)
+                            .unwrap_or(DeviceType::Speaker),
```
3bnz, line 412: Here it likely makes sense to introduce `DeviceType::Mic` - but that could be done in a separate PR, which expands the supported [`pa_device_port_type_t`](https://docs.rs/libpulse-sys/latest/libpulse_sys/def/enum.pa_device_port_type_t.html) icons.

clotodex:
tested and works for me too
the slider is still not snappy though (it does not jump around anymore, but since there is no debouncing it "lags" behind) - could also be a future improvement

romanstingler:
> future

@clotodex there is already an open ticket for that  
https://github.com/MalpenZibo/ashell/issues/487

I think that is fine to tackle that one and also the mic one suggested by @levitatingpineapple 
in separate. 

@MalpenZibo said new release in about 1 month. 
so meanwhile people using -git could test this while UI fixes and additions are pouring in.

src/modules/settings/audio.rs
```diff
@@ -292,6 +315,7 @@ impl AudioSettings {
                     }
                 }
             })
+        // })
```
MalpenZibo: ?

src/services/audio.rs
```diff
@@ -236,6 +222,80 @@ impl AudioService {
...
+
+    /// Iterates over all audio routes which user should be able to select
+    /// This includes devices with multiple available hardware ports as well as
+    /// [smart-filters](https://pipewire.pages.freedesktop.org/wireplumber/policies/smart_filters.html)
+    fn route_iter(devices: &[Device]) -> impl Iterator<Item = Route<'_>> {
+        devices.iter().flat_map(|device| {
+            if device.is_filter {
+                if device.ports.len() > 1 {
+                    let device_name = device.name.as_str();
+                    error!("Unexpected multiple ports in a filter node: {device_name}")
```
MalpenZibo: Maybe a `warn` in this case is better?

PR #456, opened 2026-01-30
create a built-in Notifications module for ashell 

Yoimiya-Naganohara:
This pull request adds a new notifications module to the application, integrating notifications into the app's configuration, UI, and event handling. The changes include new configuration options, menu integration, message handling, and icon support for notifications.

**Notifications Module Integration:**

* Added a new `Notifications` module, including `NotificationsModuleConfig` with customizable options such as format, timestamps, maximum notifications, body display, and view mode. (`src/config.rs`, `src/modules/notifications.rs`) [[1]](diffhunk://#diff-cba64c21ab992eaad29fce147a08f4560a4769bc14682b8a96081a5fd02dbecdR258-R283) [[2]](diffhunk://#diff-90d2c2330dd750ac18b6a3a2b13fcb041216ac7e0efef9a2ff729ebf2184f686R19)
* Registered `Notifications` in the main `App` struct, message enum, initialization, update, and event handling logic, ensuring the module is properly initialized and updated with configuration reloads. (`src/app.rs`) [[1]](diffhunk://#diff-0b979b6de560b29c1d97336878db56179224aa21d96fe099630c6628479ed020R13) [[2]](diffhunk://#diff-0b979b6de560b29c1d97336878db56179224aa21d96fe099630c6628479ed020R69) [[3]](diffhunk://#diff-0b979b6de560b29c1d97336878db56179224aa21d96fe099630c6628479ed020R90-R91) [[4]](diffhunk://#diff-0b979b6de560b29c1d97336878db56179224aa21d96fe099630c6628479ed020R116-R117) [[5]](diffhunk://#diff-0b979b6de560b29c1d97336878db56179224aa21d96fe099630c6628479ed020R141-R164) [[6]](diffhunk://#diff-0b979b6de560b29c1d97336878db56179224aa21d96fe099630c6628479ed020R216-R219) [[7]](diffhunk://#diff-0b979b6de560b29c1d97336878db56179224aa21d96fe099630c6628479ed020R444-R451)
* Integrated the notifications menu into the UI, allowing users to open the notifications menu and view notifications. (`src/menu.rs`, `src/app.rs`, `src/modules/mod.rs`) [[1]](diffhunk://#diff-aeb300263b37f95d7d14af8f5f3aa31740c92a10fd7a174635c1f916ab3b9e02R19) [[2]](diffhunk://#diff-0b979b6de560b29c1d97336878db56179224aa21d96fe099630c6628479ed020R559-R566) [[3]](diffhunk://#diff-90d2c2330dd750ac18b6a3a2b13fcb041216ac7e0efef9a2ff729ebf2184f686R295-R300)

**Configuration and Subscription Support:**

* Extended the main `Config` struct and default implementation to include `notifications` configuration, and updated module name parsing to recognize `Notifications`. (`src/config.rs`) [[1]](diffhunk://#diff-cba64c21ab992eaad29fce147a08f4560a4769bc14682b8a96081a5fd02dbecdR35) [[2]](diffhunk://#diff-cba64c21ab992eaad29fce147a08f4560a4769bc14682b8a96081a5fd02dbecdR56) [[3]](diffhunk://#diff-cba64c21ab992eaad29fce147a08f4560a4769bc14682b8a96081a5fd02dbecdR622) [[4]](diffhunk://#diff-cba64c21ab992eaad29fce147a08f4560a4769bc14682b8a96081a5fd02dbecdR649)
* Added notification subscription handling to the main app, wiring up notification events through the message system and subscriptions. (`src/app.rs`, `src/modules/mod.rs`) [[1]](diffhunk://#diff-0b979b6de560b29c1d97336878db56179224aa21d96fe099630c6628479ed020R602-R604) [[2]](diffhunk://#diff-90d2c2330dd750ac18b6a3a2b13fcb041216ac7e0efef9a2ff729ebf2184f686R339-R343)

**Icon Support:**

* Added new static icons for notifications (`Bell` and `BellBadge`) to be used in the UI. (`src/components/icons.rs`) [[1]](diffhunk://#diff-cfe9b2fb6cfce1602e8c3f1f69f55b90f73847b5ee04071a0f26786199f82825R114-R115) [[2]](diffhunk://#diff-cfe9b2fb6cfce1602e8c3f1f69f55b90f73847b5ee04071a0f26786199f82825R214-R215)


website/docs/configuration/modules/notifications.md
```diff
@@ -0,0 +1,49 @@
+---
+sidebar_position: 10
```
romanstingler: privacy has already position 10 we need to find another position

romanstingler:
Notification dismiss/Clear buttons don’t actually remove entries from the notification service/daemon. `Message::NotificationClicked` now removes only from the module’s local `self.notifications` map, never from the service (`self.service`) or the global `NOTIFICATIONS` store that NotificationsService polls. 
Any subsequent event will clone the service’s still-populated map back into the UI, causing “cleared” notifications to reappear. 
The Clear button has the same issue (local `clear()` only). Users can no longer dismiss notifications persistently.


Why did you disable the DBus signals?? 
In my opinion, the 100ms polling adds constant CPU wakeups and copies the whole map each tick.


`show_bodies`  = false

still shows the last empty line

xdph_screenshot_1eb1c914

Yoimiya-Naganohara:
> Notification dismiss/Clear buttons don’t actually remove entries from the notification service/daemon. `Message::NotificationClicked` now removes only from the module’s local `self.notifications` map, never from the service (`self.service`) or the global `NOTIFICATIONS` store that NotificationsService polls. Any subsequent event will clone the service’s still-populated map back into the UI, causing “cleared” notifications to reappear. The Clear button has the same issue (local `clear()` only). Users can no longer dismiss notifications persistently.
> 
> Why did you disable the DBus signals?? In my opinion, the 100ms polling adds constant CPU wakeups and copies the whole map each tick.
> 
> `show_bodies` = false
> 
> still shows the last empty line
> xdph_screenshot_1eb1c914

Thanks! I hadn't noticed that before. I'm working on fixing those issues now.

src/modules/notifications.rs
```diff
@@ -0,0 +1,257 @@
...
+                        let action_key = notification.actions[0].clone();
+                        tokio::spawn(async move {
+                            NotificationDaemon::invoke_action(id, action_key).await.ok();
+                        });
+                    }
+                // Close the notification properly (removes from both daemon and global state)
+                tokio::spawn(async move {
+                    let _ = NotificationDaemon::close_notification_by_id(id).await;
+                });
+                // Remove from local state immediately for responsive UI
```
MalpenZibo: This should be done by returning a Task to perform an async function

src/modules/notifications.rs
```diff
@@ -0,0 +1,257 @@
...
+                self.notifications.remove(&id);
```
MalpenZibo: if we return a task to perform the async operation we could also place this code in the task "callback"

src/modules/notifications.rs
```diff
@@ -0,0 +1,257 @@
...
+            }
+            Message::ClearNotifications => {
+                // Close each notification through the daemon (cleans up both daemon and global state)
+                let notification_ids: Vec<u32> = self.notifications.keys().copied().collect();
+                tokio::spawn(async move {
```
MalpenZibo: same as before

src/modules/notifications.rs
```diff
@@ -0,0 +1,257 @@
...
+        }
+    }
+
+    pub fn subscription(&self) -> Subscription<Message> {
+        use crate::services::ReadOnlyService;
+        NotificationsService::subscribe().map(Message::Event)
+    }
+
+    pub fn menu_view<'a>(&'a self, _id: Id, theme: &'a AshellTheme) -> Element<'a, Message> {
+        let mut notifications_data: Vec<_> = self.notifications.values().cloned().collect();
```
MalpenZibo: there isn't a way to avoid this clone? Maybe we could directly save in the state the vector

src/modules/notifications.rs
```diff
@@ -0,0 +1,257 @@
...
+
+        // Sort by timestamp (newest first) and limit
+        notifications_data.sort_by(|a, b| b.timestamp.cmp(&a.timestamp));
```
MalpenZibo: time consuming operation should go in the update method if it's possible

src/services/notifications/dbus.rs
```diff
@@ -0,0 +1,215 @@
...
+use zbus::{
+    Connection,
+    fdo::{DBusProxy, RequestNameFlags, RequestNameReply},
+    interface,
+    names::WellKnownName,
+    zvariant::{self, OwnedValue},
+};
+
+// Global static for storing notifications - accessed by service layer
+pub static NOTIFICATIONS: std::sync::OnceLock<std::sync::Mutex<HashMap<u32, Notification>>> =
```
MalpenZibo: Maybe we could avoid this static global object. I haven't checked all the code, but if I understand correctly, you should take inspiration from the tray service

MalpenZibo (commented):
nice addition!! Thanks. There's some work to do, but it could be a great module.

I left a preliminary review with my main concern

Yoimiya-Naganohara:
Thanks for your review! I’ll get those issues sorted out. I’m also working on adding a grouped notification feature.

Yoimiya-Naganohara:
Thanks for the review. Addressing the requested changes now. Also, the grouped notification feature is currently being integrated.

src/modules/notifications.rs
Yoimiya-Naganohara: fixed

src/modules/notifications.rs
Yoimiya-Naganohara: fixed

src/modules/notifications.rs
Yoimiya-Naganohara: solved

src/modules/notifications.rs
Yoimiya-Naganohara: moved to the update

src/services/notifications/dbus.rs
Yoimiya-Naganohara: fixed

src/modules/notifications.rs
Yoimiya-Naganohara: fixed

src/services/notifications/dbus.rs
Yoimiya-Naganohara: this has been replaced

romanstingler:
I just tested it, the functionality is fine for me. 
Good to see the Vec preference for small numbers.
I did a small benchmark and it is as expected.

```
warning: `ashell` (bin "bench_notifications") generated 1 warning
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 9.33s
     Running `target/debug/bench_notifications`
Benchmarking notification storage: Vec vs HashMap
Sizes: [10, 20, 50, 100], Iterations per measurement: 1000

Size: 10
  Insert: Vec 2340.00 ns, HashMap 5115.00 ns
  Lookup: Vec 204.00 ns, HashMap 1773.00 ns
  Remove: Vec 2820.00 ns, HashMap 8371.00 ns

Size: 20
  Insert: Vec 7288.00 ns, HashMap 10726.00 ns
  Lookup: Vec 373.00 ns, HashMap 1769.00 ns
  Remove: Vec 8079.00 ns, HashMap 14112.00 ns

Size: 50
  Insert: Vec 35731.00 ns, HashMap 18541.00 ns
  Lookup: Vec 575.00 ns, HashMap 1300.00 ns
  Remove: Vec 34662.00 ns, HashMap 21208.00 ns

Size: 100
  Insert: Vec 117876.00 ns, HashMap 37377.00 ns
  Lookup: Vec 1034.00 ns, HashMap 1307.00 ns
  Remove: Vec 119005.00 ns, HashMap 39805.00 ns
  ```
  
  Well done and thanks for the contribution.
  
  
  @MalpenZibo 
  regarding the documentation 
  should we re-order all the modules in the docs alphabetically? 
  
image

MalpenZibo:
I will check tomorrow the fixes. @romanstingler it make sense!

romanstingler:
@Yoimiya-Naganohara I added a PR for reordering of all the modules. 
I mentioned that above, please use
`sidebar_position: 5`. 

Additionally, please run followind commands before pushing
`cargo clippy --all-features -- -D warnings`  and
`cargo fmt --all -- --check`

Thanks

@MalpenZibo please merge that one before this one
- https://github.com/MalpenZibo/ashell/pull/456

Yoimiya-Naganohara:
It's done

clotodex:
Tested and working.

Noticed a small nitpick. A single notification is a smaller window than no notifications.
This is purely preference but I would love the empty notifications case to be a smaller spacer.
When you clear the last notification the window gets bigger :P 
Feel free to ignore, this shouldn't block a merge.

I have an open question though which we should also address in docs:
What is the strategy for notification popups? Does dunst alongside work?
Since this registers itself as handling notifications (or is blocked from doing so when others are registered) this is mutually exclusive with my current swaync setup (which poweres both the notification list and the popups).
I could not care less about swaync but for me to move over I would need some generic popup support as well. Since it breaks my current swaync I would at least mention that this happens or an alternative in the docs.

Yoimiya-Naganohara:
The notification toast is under development. The UI will be improved.

Yoimiya-Naganohara:
The feature is almost done, but the toast layer still needs proper handling. Currently it's implemented as a basic computed layer with manually calculated size and position.

MalpenZibo:
Sorry, it's a big change, so I will not merge it into the next release. Anyway, the toast surface makes sense. The problem right now is that iced takes some time to create and correctly render a new surface (taking into consideration the HiDPI monitor). For this reason, the menu surface is always open.

You are "correctly" creating a new surface to show a notification when necessary. I need to check if this solution will work. After the release, I will update the iced library to the latest version to see if this behavior has been improved.

MalpenZibo:
Ok here I am! Sorry for the delay.

image

First, the layout, I'd like to have a bigger icon in the first column. Then it could be ok for now, but I'd like to have an X button on the top right to close the notification and reserve the click on notification for an optional call to action

## Scaling issue

image

Sadly, as I thought, we have the same scaling issue that I had with the menu. On a HiDPI monitor, the time between the first frame and the point where iced fixes the scaling of the surface is not trivial. 

I will try with the new version of iced to see if this issue is resolved:
- If it's resolved, we could pause this PR, wait until the update (I'm already working on that), then rebase the PR after the update.
- If it's not resolved, the only solution will be to maintain a third surface always open, like for the menu :(. In that case, we should find a strategy to check the memory consumption of this solution, and in case, try to find a way to reduce it

MalpenZibo:
Ok, great news. The new version of iced seems to fix this issue.

This means that we can leave the surface creation logic as it is here and, in the future, resolve the menu surface problem. That's great news, we could also simplify that `outputs.rs` awful system :sweat_smile:

Yoimiya-Naganohara:
I will improve this

Yoimiya-Naganohara:
There is an issue with the toast layer's dimensions: it's blocking mouse clicks on the underlying window. Implementing click-through functionality or resizable layer would solve this.

MalpenZibo:
ok, so let's block this PR until the update of iced is merged. So we can easily open and close the notification surface

romanstingler:
I had a look at the branch at the current state, I think it looks fine for a first merge, so people can use it and give feedback. 
Sure, features like click to open should be implemented, but not necessarily now. 
Idk if you @MalpenZibo want to have some of the commits stashed.

MalpenZibo:
Yes! Now that we've merged the iced update the main blocker here is gone and we can merge it.

@Yoimiya-Naganohara are you able to fix the conflict?

MalpenZibo:
I double-checked the code. There are a few things that I would like to change to make everything more in line with the rest of the codebase.
I can handle everything in a dedicated PR merging this as it is, if it's ok for you @Yoimiya-Naganohara. Otherwise, I can make a deep review to point out everything that I'd like to change and let you do the work. Let me know what you prefer. 
I'd like the idea of working on this, but I prefer to ask you first

MalpenZibo:
Nevermind! There are some things that I prefer to fix by myself!

I will merge this PR as soon as I can.

Yoimiya-Naganohara:
I’ve been working on other things recently, so I’ve just resolved the conflicts.

Yoimiya-Naganohara:
> I double-checked the code. There are a few things that I would like to change to make everything more in line with the rest of the codebase. I can handle everything in a dedicated PR merging this as it is, if it's ok for you @Yoimiya-Naganohara. Otherwise, I can make a deep review to point out everything that I'd like to change and let you do the work. Let me know what you prefer. I'd like the idea of working on this, but I prefer to ask you first

surely

PR #458, opened 2026-01-30
Show Update button only when updates available

denqxotl:
2026-01-30-184359_hyprshot


romanstingler:
looks good to me

denqxotl:
@MalpenZibo rebased and fixed column width to MenuSize::Small

PR #464, opened 2026-02-05
adjust menu width based on weather data availability

romanstingler:
@MalpenZibo

sorry missed that during review. 

xdph_screenshot_160f82ef
  
  
when there is no weather enabled we want to shrink the menu 
xdph_screenshot_6a1f0409

### DIFF
--- a/src/modules/tempo.rs
+++ b/src/modules/tempo.rs
@@ -107,11 +107,18 @@ impl Tempo {
     }
 
     pub fn menu_view<'a>(&'a self, theme: &'a AshellTheme) -> Element<'a, Message> {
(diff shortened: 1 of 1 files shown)


MalpenZibo:
Have you check with the latest main? Because in theory I have done #462 just to resolve this issue. Maybe I miss something, I will check

MalpenZibo:
Ok I saw the fix. Probably I forgot to use max width or something like that 😅

MalpenZibo:
I fixed this problem and a build problem here https://github.com/MalpenZibo/ashell/pull/467

PR #465, opened 2026-02-05
fix playback state

qmmp123:
Sorry for my english

You've already got changed state from dbus so there is no need to compare value with previous state. Also strings below did not work as you expected
```rust
let volume = s.volume;
```
```rust
let state = s.state;
```
These strings run once and do not change variable's value


qmmp123:
Oh, I did not see this PR https://github.com/MalpenZibo/ashell/pull/319

PR #468, opened 2026-02-06
Fix workspace scroll

megabyte6:
To me, it makes more sense to only scroll through visible items.

When using `visibility_mode` as `MonitorSpecific` or `MonitorSpecificExclusive`, you only see the workspaces for the current monitor. This PR doesn't let the mouse scroll to other monitors when those workspaces aren't being displayed on the current `ashell`.

If there are any adjustments you wish to be made to this PR, feel free to ping me!


flake.nix
```diff
@@ -70,14 +70,14 @@
           inherit ldLibraryPath;
           buildInputs = buildInputs ++ [
             pkgs.rust-analyzer-unwrapped
-            pkgs.nixfmt-rfc-style
```
MalpenZibo, line 73: Why these changes to the nix file?

MalpenZibo (commented):
This will change the default behavior. I need to understand if it's ok or if we should place a config flag

flake.nix
megabyte6: I got the following warning for `nixfmt-rfc-style` when entering the dev shell.

`evaluation warning: nixfmt-rfc-style is now the same as pkgs.nixfmt which should be used instead.`

It's honestly quite minimal. I just figured I might as well bundle it in so it gets added.
If you're against it I can revert the change.

megabyte6:
> This will change the default behavior. I need to understand if it's ok or if we should place a config flag

@MalpenZibo I get where you're coming from. If you decide on adding a config flag, just let me know; I'm happy to get started on implementation.

clotodex:
My opinion is the default to be only visible makes a lot of sense from UX
I would add a config flag though so the old behavior can be used.

PR #469, opened 2026-02-06
implement simple custom background option for menus

Follpvosten:
as per the title.

for my specific usecase, this allows me to style the top bar exactly like the one in GNOME - solid black bar, drop-down menus in grey/gtk colour. it remains backwards comspatible by using an Option and falling back to `background_color` if menu_background was not set in config.

I'm open for criticism or input/design suggestions, but also open to just getting a "no" (in which case I'd just maintain a downstream fork for this :)

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -538,6 +538,7 @@ impl App {
                     if let Some(updates) = self.updates.as_ref() {
                         self.menu_wrapper(
                             id,
+                            &self.theme,
                             updates.menu_view(id, &self.theme).map(Message::Updates),
                             *button_ui_ref,
                         )
@@ -547,11 +548,13 @@ impl App {
                 }
                 Some((MenuType::Tray(name), button_ui_ref)) => self.menu_wrapper(
                     id,
+                    &self.theme,
                     self.tray.menu_view(&self.theme, name).map(Message::Tray),
                     *button_ui_ref,
                 ),
                 Some((MenuType::Settings, button_ui_ref)) => self.menu_wrapper(
                     id,
+                    &self.theme,
                     self.settings
                         .menu_view(id, &self.theme, self.theme.bar_position)
                         .map(Message::Settings),
@@ -559,6 +562,7 @@ impl App {
                 ),
                 Some((MenuType::MediaPlayer, button_ui_ref)) => self.menu_wrapper(
                     id,
+                    &self.theme,
                     self.media_player
                         .menu_view(&self.theme)
                         .map(Message::MediaPlayer),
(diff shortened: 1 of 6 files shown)


Follpvosten:
~~TODO: should probably also affect the new Tempo module (~~
nevermind it already does, good thing I fetched upstream before I wrote the patch :sweat_smile:

MalpenZibo:
Ok, for these kinds of changes I would like to open a discussion on improving the styling configuration in the next release (post March)

MalpenZibo:
Ok. Generally speaking, I'm ok with this change. About the configuration, I have some doubts.

I would like to maintain the palette as a 1to1 port of the iced palette. For this reason, the `menu_background` should go under the MenuAppearance section. 
And then I think that the best solution would be to add a link to the color of a specific part of the interface to a palette color. For example, in this case, `menu_background: Background | Primary  | ...`, maybe with the option of specifying a custom color. 

I need to think about this also because I plan to have a similar solution for the status bar surface background in the future.

Follpvosten:
interesting! being able to set a background for the status bar itself would also solve my problem, perhaps even more elegantly. I hadn't considered that option. From my view, both achieve a very similar goal though: Being able to specify different backgrounds for different major parts of the UI.

for now I will rework my existing patch to implement things as requested. Further discussion does indeed sound reasonable though.

Follpvosten:
For now I have kept the `menu_background` in the internal `AshellTheme` for simplicity (otherwise I'd have to define an internal-specific version of `MenuAppearance` where `background` is resolved to a `palette::Background`).

But in the public API, I have moved `appearance.menu_background` to `appearance.menu.background`.

Docs-wise, I have removed the line I had previously (where it was documented directly under palette -> background), and instead created a new section further down in `palette.md`, because I felt that it fit best there as it's colour related. I'd be happy to change that again of course, in case it makes more sense to put it under General.

clotodex:
@Follpvosten the palette rework is largely done. Feel free to check #574 and play around with it

MalpenZibo:
A little update here. I'm working on this https://github.com/MalpenZibo/ashell/pull/867.

So I'm making changes to the `style` system of the bar. 

I'm not sure about making the surface color (for both bar and menus) fully customizable (right now it is transparent or solid). Anyway, I prefer to centralize the changes around this topic to PR on top of 867 (also, I prefer to make these PR personally right now).

So I will block this PR for now, sorry, but I need to calibrate how many config and personalization add to the bar. Maybe in the near future I can add a `transparent | solid | CustomColor` feature. If you want to open an issue with your specific requirement to discuss about it would be great.

PR #470, opened 2026-02-06
reduce unnecessary allocations

Follpvosten:
# rationale
calls to `Row::new()`/`Column::new()` followed by a bunch of `.push`/`.push_maybe` result in a lot of unnecessary allocations as buffers grow; with the way stuff like opening/closing a menu is currently implemented, this happens every time a menu is opened.

this is mostly relevant for low-end mobile end user hardware, such as my tiny N100 laptop. in addition to the randomly occuring delays I'm still trying to figure out the cause of, without this patch there is a consistent noticeable delay when opening menus.

# my current/first approach
for a first go at fixing this, I picked the option that's the least intrusive to the codebase: replacing calls to `Row::new()` and `Column::new` with `::with_capacity` calls instead (since the max length is almost always known in advance). this should be fine in terms of memory usage (since at least on linux-shaped kernels, allocated memory isn't actually used until it's written to), it will reduce overall allocation count, and it might actually cause smaller allocations than pushing without setting the capacity (https://doc.rust-lang.org/std/vec/struct.Vec.html#capacity-and-reallocation). I have also replaced instances of `::new().push().push().push()` with the row!/column! macro where possible.

# possible alternatives
alternative approaches to this, for example ones which always allocate exactly as much memory as needed, are possible and not hard to implement, but they would require touching a lot more lines of code; I think the easiest replacement for `::new().push().push_maybe()` chains would be creating an iterator chain using `Option::into_iter()`, `Iterator::chain` and `std::iter::once()`, and then passing that iterator to Row/Column's `with_children`. I am open to changing to this if it is preferred.

### DIFF
--- a/src/modules/settings/mod.rs
+++ b/src/modules/settings/mod.rs
@@ -23,9 +23,7 @@ use crate::{
 };
 use iced::{
     Alignment, Background, Border, Element, Length, Padding, Subscription, Task, Theme,
-    widget::{
-        Column, MouseArea, Row, Space, button, column, container, horizontal_space, row, text,
-    },
+    widget::{Column, MouseArea, Row, Space, button, container, horizontal_space, row, text},
     window::Id,
 };
 
@@ -495,7 +493,7 @@ impl Settings {
                 .power
                 .battery_menu_indicator(theme)
                 .map(|e| e.map(Message::Power));
-            let right_buttons = Row::new()
+            let right_buttons = Row::with_capacity(2)
                 .push_maybe(
                     self.lock_cmd
                         .as_ref()
@@ -514,7 +512,7 @@ impl Settings {
                 )
                 .spacing(theme.space.xs);
 
-            let header = Row::new()
+            let header = Row::with_capacity(3)
                 .push_maybe(battery_data)
                 .push(Space::with_width(Length::Fill))
                 .push(right_buttons)
@@ -617,7 +615,7 @@ impl Settings {
                 Position::Bottom => (None, source_slider.map(|e| e.map(Message::Audio))),
             };
 
-            Column::new()
+            Column::with_capacity(11)
                 .push(header)
                 .push_maybe(
                     self.sub_menu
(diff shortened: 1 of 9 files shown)


romanstingler:
Yes, the iterator approach is indeed hard to read and maintain, especially with conditionals. 
Let's stick with `::with_capacity()` for now; it's clean, reduces allocations without overcomplicating the code, and we can profile/iterate later if needed. 

can you also post your config so that I can see what parts of the code people are using.

Thanks for the feedback!

Follpvosten:
> can you also post your config so that I can see what parts of the code people are using.

sure! my config currently lives here :) https://github.com/Follpvosten/sylvieFlake/blob/tzuyu/home-manager/modules/niri/ashell.nix#L18 (except for theming/styling, that's in colours.nix)

src/modules/settings/mod.rs
megabyte6, line 1: @Follpvosten I agree that this is a little hard to read. This is what I came up with, which is a little clearer to me. I haven't tested it, and I'm not 100% confident about this, but I thought I'd just put it out there.

```rust
let mut iter = buttons.into_iter();

while let Some((left_button, left_menu)) = iter.next() {
    let (right_button, right_menu): (Element<'a, Message>, _) = iter
        .next()
        .unwrap_or_else(|| (horizontal_space().into(), None));

    section = section.push(
        row![left_button, right_button]
            .width(Length::Fill)
            .spacing(theme.space.xs),
    );

    if let Some(menu) = left_menu {
        section = section.push(sub_menu_wrapper(theme, menu));
    }
    if let Some(menu) = right_menu {
        section = section.push(sub_menu_wrapper(theme, menu));
    }
}
```

We could replace the code between the definition of `section` and the `section.into()` at the end.

Also, is my type for `(right_button, right_menu)` even correct?

MalpenZibo (commented):
need rebase, then let's merge

src/modules/settings/mod.rs
Follpvosten: I think this looks better indeed, but I don't know if this PR is the right place to fix it :sweat_smile:

Follpvosten:
rebased

megabyte6:
> Sorry, my bad I merged another PR before this one and now needs to be rebased again

sorry :sweat_smile:

Issue #437, opened 2026-01-20
Tray has still some issues


romanstingler:
If I close and open ashell multiple times I can sometimes trigger some issues. 

I will investigate this weekend.

Follpvosten:
just gonna collect some logs here.

```
ERROR [ashell::services::tray] Failed to listen for tray events: org.freedesktop.DBus.Error.ServiceUnknown: The name :1.413 was not provided by any .service files
ERROR [ashell::services::tray] Tray service error
```

romanstingler:
I hope I can finalize my fixes this weekend.

romanstingler:
hey @Follpvosten :P 
maybe you can build my 
```
git clone https://github.com/romanstingler/ashell.git
cd ashell
git checkout improve-tray-discovery

cargo b --release
./target/release/ashell
```

andradei:
Same as @Follpvosten is happening in Void linux, building release 0.7.0 locally.

romanstingler:
yes I have already a PR in progress, I am on a conference and will be back on thursday and will try to finish it for the next release.

romanstingler:
@Follpvosten @andradei 

if you have the opportunity to build it on your system, 
could you please try my refactor (just pushed) 

```
git clone https://github.com/romanstingler/ashell.git
cd ashell
git checkout improve-tray-discovery

cargo b --release
./target/release/ashell
```

andradei:
@romanstingler I haven't tried your fork yet. I noticed I don't have the issue on Arch. In my case, on Void, it could be that ashell starts before DBUS starts up completely, adding a `sleep` of 1-3 seconds to it fixes it.

PR #475, opened 2026-02-08
Improve tray icon discovery and reliability

romanstingler:
### Improved tray service discovery and icon resolution

This PR fixes issues where certain apps (specifically **Spotify, Telegram, and JetBrains Toolbox**) weren't showing up or were missing icons in the tray because they don't follow the standard `StatusNotifierItem` naming conventions.

### What’s changing:

* Instead of just looking for names that explicitly say "StatusNotifierItem," the service now uses introspection. It checks common paths like `/StatusNotifierItem` and `/MenuBar` to see if they actually implement the required interface. If it looks like a tray item, we use it.
* I added a fallback chain for icons. It now tries to pull from the DBus pixmap first, then the icon name property, and finally falls back to the app's `.desktop` file.
* If one tray item fails to initialize, it won't hang the whole loop anymore. It’ll log the error and move on to the next item.

### Technical Notes:

The discovery flow now scans registered DBus services and runs an introspection check for the `org.kde.StatusNotifierItem` interface. For icons, the priority is now:

1. Raw pixmap data
2. Theme lookup via `icon_name`
3. ID/Title matching
4. Desktop file lookup.

closes https://github.com/MalpenZibo/ashell/issues/437

### DIFF
--- a/src/services/tray/dbus.rs
+++ b/src/services/tray/dbus.rs
@@ -1,5 +1,5 @@
 use iced::futures::StreamExt;
-use log::{info, warn};
+use log::{debug, info, warn};
 use std::time::Duration;
 use zbus::{
     Connection, Result,
@@ -18,7 +18,7 @@ const OBJECT_PATH: &str = "/StatusNotifierWatcher";
 
 #[derive(Debug, Default)]
 pub struct StatusNotifierWatcher {
-    items: Vec<(UniqueName<'static>, String)>,
+    items: std::collections::HashMap<String, String>,
 }
 
 impl StatusNotifierWatcher {
@@ -65,19 +65,20 @@ impl StatusNotifierWatcher {
                             }
                         } else if let BusName::Unique(name) = &args.name {
                             let mut interface = internal_interface.get_mut().await;
-                            if let Some(idx) = interface
-                                .items
-                                .iter()
-                                .position(|(unique_name, _)| unique_name == name)
-                            {
-                                let emitter =
-                                    SignalEmitter::new(&internal_connection, OBJECT_PATH).unwrap();
-                                let service = interface.items.remove(idx).1;
-                                StatusNotifierWatcher::status_notifier_item_unregistered(
+                            if let Some(service) = interface.items.remove(&name.to_string()) {
+                                let emitter = match SignalEmitter::new(&internal_connection, OBJECT_PATH) {
+                                    Ok(emitter) => emitter,
+                                    Err(err) => {
+                                        warn!("Failed to create signal emitter for unregistration: {err}");
+                                        continue;
+                                    }
+                                };
+                                if let Err(err) = StatusNotifierWatcher::status_notifier_item_unregistered(
                                     &emitter, &service,
                                 )
-                                .await
-                                .unwrap();
+                                .await {
+                                    warn!("Failed to emit status notifier item unregistered signal for service '{service}': {err}");
+                                }
                             }
                         }
                     }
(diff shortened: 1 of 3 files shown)


src/services/tray/mod.rs
```diff
@@ -272,13 +447,29 @@ impl StatusNotifierItem {
...
+            Err(_) => {
+                // Try each fallback candidate
+                let mut found_icon = None;
+                for candidate in fallbacks.iter() {
+                    if let Some(icon) = get_icon_from_name(candidate) {
+                        found_icon = Some(icon);
+                        break;
+                    }
+                }
+                found_icon
```
MalpenZibo: code duplication

src/services/tray/dbus.rs
```diff
@@ -129,13 +130,65 @@ impl StatusNotifierWatcher {
                         &emitter,
                     )
                     .await;
+            } else {
+                // Try introspection for other services
+                if let Some(path_hint) = Self::find_status_notifier_path(conn, name_str).await {
```
MalpenZibo, line 130: could make sense to have a timeout here?

src/services/tray/mod.rs
```diff
@@ -392,7 +598,12 @@ impl TrayService {
         let mut menu_layout_change = Vec::with_capacity(items.len());
 
         for name in items {
-            let item = StatusNotifierItem::new(conn, name.to_string()).await?;
+            let item = match StatusNotifierItem::new(conn, name.to_string()).await {
+                Ok(item) => item,
+                Err(_) => {
```
MalpenZibo, line 562: Maybe we should place a `warn`?

src/services/tray/mod.rs
```diff
@@ -520,34 +731,27 @@ impl TrayService {
...
+                    State::Active(conn)
                 }
-            }
+                Err(err) => {
+                    error!("Failed to listen for tray events: {err}");
+                    State::Error
+                }
+            },
             State::Error => {
-                error!("Tray service error");
```
MalpenZibo, line 549: Is it ok to completely remove the log here?

src/services/tray/dbus.rs
```diff
@@ -129,13 +130,65 @@ impl StatusNotifierWatcher {
...
+                        .await;
+                }
             }
         }
         Ok(())
     }
 }
 
 impl StatusNotifierWatcher {
+    async fn find_status_notifier_path(conn: &Connection, name: &str) -> Option<String> {
```
MalpenZibo, line 172: I'm a bit worried about the performance of this check, but let's try and see ifit  causes a problem

1randomguy:
Tested it, and it works well on my setup!

romanstingler:
today and tomorrow I have not that much time, but I will try to clean up as much PRs on Thu-Sa

romanstingler:
> Tested it, and it works well on my setup!


if you have the opportunity to build it on your system, 
could you please try my refactor (just pushed) 

```
git clone https://github.com/romanstingler/ashell.git
cd ashell
git checkout improve-tray-discovery

cargo b --release
./target/release/ashell
```

with the tray apps I tested with it works on my machine, but I want to have some more people testing it. 

thanks

1randomguy:
> > Tested it, and it works well on my setup!
> 
> if you have the opportunity to build it on your system, could you please try my refactor (just pushed)
> 
> ```
> git clone https://github.com/romanstingler/ashell.git
> cd ashell
> git checkout improve-tray-discovery
> 
> cargo b --release
> ./target/release/ashell
> ```
> 
> with the tray apps I tested with it works on my machine, but I want to have some more people testing it.
> 
> thanks

I tried it, but some icons are wrong. Somehow my spotify thinks it is a ghostty terminal :( 
image

romanstingler:
@1randomguy 
thanks for reporting 

can you please run 
`RUST_LOG=ashell::services::tray=debug cargo r > /tmp/log.txt`
and then open you tray apps quickly - there is a lot of debugging output - and close ashell and upload that file.

1randomguy:
[tray.log](https://github.com/user-attachments/files/25813507/tray.log)

https://github.com/user-attachments/assets/edf0a613-78b9-4c33-ba26-71aa764b1033

in this case the spotify tray icon was fine, but opening spotify broke the others 😔

clotodex:
#495 built a new xdg icon discovery, and solved all icons being successfully discovered. not sure if this applies to or also works for tray, but wanted to let you know @romanstingler

romanstingler:
yes exactly because of #495 I set this one as draft for now. 
I wanted to check after the other code is ready if this one is still needed. I will close that as soon as I had time to have a look at the other code

Issue #482, opened 2026-02-10
calender of tempo module is off by one day.


Abdisto:
<img width="248" height="261" alt="Image" src="https://github.com/user-attachments/assets/d4dac946-2db1-4106-bcf0-30ef49aa94b2" />

I know, that this module is currently under heavy development and this should be caught earlier or later. Still wanted to report on that even though I am unsure if issue reporting is welcomed. I am unsure if this a local problem or something being wrong upstream, either case I don't have any reproduction step. So what is the matter. 

I am currently on the latest build via the flake (I also cloned the repo, updated the lock file, and built it myself, issue persists). The Calender displayed is off. Like it stating that the 10th of February being Wednesday. Other applications (cal) report it back correctly so doesn't seem to be a localization issue (Don't even know if there even anything that could cause smth. like this). I don't know about your guys implementation but i saw in the commit history that you use an external dependency so I don't think that the values fetched are wrong but prob. something like a wrong Iteration pushing everything 1 day to the side.

romanstingler:
hey, thank you for your issue report, always nice to have some -git users who test new code.
It seems that we totally missed that. 
I will have a look into it 

PS: I pushed a fix for it.

PR #485, opened 2026-02-11
Tempo: Remove primary color highlight for current date in calendar if not current month

romanstingler:
found due to  https://github.com/MalpenZibo/ashell/issues/482
xdph_screenshot_10249fca

### DIFF
--- a/src/modules/tempo.rs
+++ b/src/modules/tempo.rs
@@ -200,8 +200,6 @@ impl Tempo {
                                             .color_maybe({
                                                 if day == self.date.date_naive() {
                                                     Some(theme.iced_theme.palette().success)
-                                                } else if day == selected_date {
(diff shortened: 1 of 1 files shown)


MalpenZibo:
That's actually to indicate the current date. I'd like the idea of having a specific color on "today"

romanstingler:
no, 
This is correct for the CURRENT month
but go to the next month 
if this is intentional then it is fine for me, just didn't know that
xdph_screenshot_743d3cbc

xdph_screenshot_44ed86f3


I am currently at work, I will fix the other things in the evening

MalpenZibo:
A ok the blu one! In my mind, when you change the month is like selecting a day, so I place the selection on the same day the next month.

It's something that I like to leave as it is. I hope it's not confusing. Let me know

PR #486, opened 2026-02-11
Improve WiFi scanning reliability for both NetworkManager and iwd backends

### LINKED ISSUE #91: Wifi connection problem
The procedure for connecting to a Wi-Fi network has some issues. Sometimes, it asks for a password even if it's not required, and sometimes, it doesn't connect at all.

romanstingler:
UPDATED: Reworked the scanning implementation.

Fixed `scan_nearby_wifi` to use `wireless_devices()` instead of `wireless_access_points()`. The old approach couldn't discover networks when none were currently visible, since it was trying to get device paths from an empty access point list.

Scans now run in parallel across all wireless devices using `join_all`. Tested with two cards (internal + USB dongle) and both scan simultaneously. Each device waits for the LastScan D-Bus property to change before returning, so we know exactly when the scan finishes.

```
❯ cargo r
   Compiling ashell v0.7.0 (/srv/opensource/ashell)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 13.30s
     Running `target/debug/ashell`
INFO [ashell::config] Decoding config file "/home/roman/.config/ashell/config.toml"
INFO [ashell::config] Config file loaded successfully
WARN [ashell::modules::clock] Clock module is deprecated and will be removed in a future release. Please migrate to the Tempo module.
WARN [ashell::services::upower] Failed to get power profile: org.freedesktop.DBus.Error.ServiceUnknown: The name is not activatable
WARN [ashell::services::network::dbus] Scan requested for device /org/freedesktop/NetworkManager/Devices/3, LastScan before: 25067423
WARN [ashell::services::network::dbus] Scan requested for device /org/freedesktop/NetworkManager/Devices/6, LastScan before: 25236251
WARN [ashell::services::network::dbus] Scan finished for device /org/freedesktop/NetworkManager/Devices/6, LastScan changed from 25236251 to 25238551
WARN [ashell::services::network::dbus] Scan finished for device /org/freedesktop/NetworkManager/Devices/3, LastScan changed from 25067423 to 25241983
```

Added cross-device SSID deduplication in wireless_access_points() — when multiple wireless devices see the same network, we keep the one with the strongest signal. Without this, the same SSID showed up multiple times in the list.
xdph_screenshot_62fc7ac3

Simplified the scanning state management. ScanningNearbyWifi(bool) replaces the old ScanningNearbyWifi + ScanCompleted pair. The state is set true synchronously in the command handler and false via the Task::perform callback after scans complete. Removed `scan_started_at, SCAN_TIMEOUT`, and the timeout wrapper — the LastScan stream handles completion detection properly so we don't need artificial timeouts.

Kept `scan_completed_at` and `SCAN_RESULT_DISPLAY_DURATION` for the "Scan complete" text that shows for 3 seconds after a scan finishes.

Graceful error handling per device — if request_scan fails on one device (e.g. scanning too frequently), it logs a warning and continues without breaking the others.

@clotodex maybe you have time to test that on your end as well. 

closes #91

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -88,7 +88,7 @@ fn subset_text(input: &str, text: &str, output_path: &str) -> Result<(), Box<dyn
     glyphs.insert(0, Some(notdef));
 
     let mut glyphs: Vec<RawGlyph<()>> = glyphs.into_iter().flatten().collect();
-    glyphs.sort_by(|a, b| a.glyph_index.cmp(&b.glyph_index));
+    glyphs.sort_by_key(|a| a.glyph_index);
     let mut glyph_ids = glyphs
         .iter()
         .map(|glyph| glyph.glyph_index)
--- a/src/modules/settings/brightness.rs
+++ b/src/modules/settings/brightness.rs
@@ -6,19 +6,18 @@ use crate::{
         brightness::{BrightnessCommand, BrightnessService},
     },
     theme::AshellTheme,
+    utils::remote_value,
 };
 use iced::{
-    Alignment, Element, Length, Subscription, Task,
-    futures::stream,
-    widget::{MouseArea, container, row, slider, text},
+    Alignment, Element, Subscription, Task,
+    mouse::ScrollDelta,
+    widget::{MouseArea, Text, container, row, slider, text},
 };
 
 #[derive(Debug, Clone)]
 pub enum Message {
     Event(ServiceEvent<BrightnessService>),
-    Change(u32),
-    MenuOpened,
-    ResetUserAdjusting,
+    Changed(remote_value::Message<u32>),
     ConfigReloaded(SettingsFormat),
 }
 
@@ -30,9 +29,6 @@ pub enum Action {
 pub struct BrightnessSettings {
     config: SettingsFormat,
     service: Option<BrightnessService>,
-    ui_percentage: u32,
-    is_user_adjusting: bool,
-    reset_timer_active: bool,
 }
 
 impl BrightnessSettings {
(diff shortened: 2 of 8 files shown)


clotodex:
Small nitpick (not sure if worth solving)
If i run a second bar, it tries to connect with the same agent name to iwd, which fails
Failing to listen to network events..

However the ui is not aware of any failure
Scanning just switches to complete immediately etc
The error is printed to the log though

A quick fix could be randomly generating the name of the agent

clotodex:
Tested it and it still seems to work on iwd
The experience is confusing though
It counts down from 10s but is already "done" after three

What are the 10s supposed to communicate?

romanstingler:
it has a 10s timeout (max duration) but the systems response could be already arriving after a few seconds. 

I don't know how to display it without showing a whole paragraph .

But yeah, maybe it is enough just to show "Scanning..." as long as it ends in a sensible amount of time and we completely remove the countdown. 
 Thanks @clotodex

clotodex:
Agreed, would just not show the countdown

romanstingler:
> Agreed, would just not show the countdown

Thanks, removed the timer

clotodex:
looking at the code, lgtm.

Does it make sense to remove the timeout? If iwd and nm have an internal timeout then potentially yes. If not then we could wait in your task for a max number of time by selecting a future and display that scanning failed in time.

src/services/network/dbus.rs
```diff
@@ -76,19 +77,37 @@ impl super::NetworkBackend for NetworkDbus<'_> {
...
+                let last_scan_before = device.last_scan().await.unwrap_or(-1);
+
+                if let Err(e) = device.request_scan(HashMap::new()).await {
+                    warn!("Scan request failed for device {device_path_str}: {e}");
+                    return Ok(());
+                }
+
+                // Wait for LastScan property to change, indicating scan finished
+                let mut last_scan_stream = device.receive_last_scan_changed().await;
+                while let Some(change) = last_scan_stream.next().await {
```
MalpenZibo: Here we should probably place a tokio timeout for safety

romanstingler:
So I started again, after the phone call with @clotodex and a deeper understanding of the network code. 

What I did is testing the scanning behavior on NM and IWD with the built-in nw-card and additionally a USB nw-card to simulate 2 active WIFI adapters.

As mentioned before we would duplicate the Network names in the nw-(sub)menu.

I extended the shared AccessPoint model with `max_bitrate` and `frequency`. 
On the NetworkManager side, both values are now read from D-Bus and used when collapsing duplicate SSIDs. 

The selection logic now prefers the access point with the higher `max_bitrate`, then the `higher frequency` (not entirely sure here, theoretically the bitrate check should keep a balance), and finally the stronger signal. 

This applies both when reducing multiple APs on a single device and when reducing APs across multiple devices.

In the NetworkManager path I now track scan requests per device and only show scanning for devices that really started a scan. When a scan finishes, that device is removed from the pending list. Old scan complete events are ignored if the device was not in that list.

I also fixed how signal strength updates work after deduplication. Before, strength updates were matched too broadly by SSID, so the wrong access point could be updated when the same network was seen by multiple devices. Now the strength event can include the AP path, and when it is available the update is applied to the correct access point.

The IWD backend was updated to mirror the same deduplication behavior as NetworkManager so both backends behave consistently. 
In practice, IWD still falls back to signal strength because the metadata we need for full parity is not available in a reliable per-visible-AP way from the D-Bus objects we inspected. 

I investigated whether IWD exposes usable frequency or bitrate data over D-Bus. Live inspection showed connected-link diagnostics such as frequency and TX/RX bitrate, but not a clear equivalent of NetworkManager’s per-access-point `max_bitrate` for scanned candidates, so that part was intentionally left unchanged.
```
> busctl call net.connman.iwd /net/connman/iwd/1/7 net.connman.iwd.StationDiagnostic GetDiagnostics
a{sv} 15 "ConnectedBss" s "d0:ea:11:15:c6:f1" "Frequency" u 2412 "Channel" q 1 "Security" s "WPA3-Personal + FT" "PairwiseCipher" s "CCMP-128" "RSSI" n -27 "AverageRSSI" n -25 "RxMode" s "802.11ax" "RxMCS" y 11 "TxMode" s "802.11ax" "TxMCS" y 11 "TxBitrate" u 2867 "RxBitrate" u 1433 "InactiveTime" u 114 "ConnectedTime" u 3528
```

Even so, the backend now deduplicates visible APs and known connections consistently, so duplicate same-SSID entries are no longer shown just because they came from different devices.

clotodex:
> In practice, IWD still falls back to signal strength because the metadata we need for full parity is not available in a reliable per-visible-AP way from the D-Bus objects we inspected.

Is this reliable for NM? I have not checked its implementation, but since NM can actively wrap iwd it either
- is also unreliable and just seems reliable form the api
- is reliable and infers its information from iwd - which we can choose to copy or ignore

For now this looks great though and would simply revisit if someone has an issue with this and get this to merge

src/services/network/dbus.rs
```diff
@@ -639,8 +675,14 @@ impl NetworkDbus<'_> {
...
+                    let max_bitrate = ap.max_bitrate().await.unwrap_or_default();
+                    let frequency = ap.frequency().await.unwrap_or_default();
                     if let Some(access_point) = aps.get(&ssid)
-                        && access_point.strength > strength
+                        && (access_point.max_bitrate > max_bitrate
+                            || (access_point.max_bitrate == max_bitrate
+                                && (access_point.frequency > frequency
+                                    || (access_point.frequency == frequency
+                                        && access_point.strength > strength))))
                     {
```
MalpenZibo: This check is duplicated in 4 different places. Could we create a common function?

src/services/network/dbus.rs
```diff
@@ -639,8 +675,14 @@ impl NetworkDbus<'_> {
...
                     let strength = ap.strength().await?;
+                    let max_bitrate = ap.max_bitrate().await.unwrap_or_default();
+                    let frequency = ap.frequency().await.unwrap_or_default();
                     if let Some(access_point) = aps.get(&ssid)
-                        && access_point.strength > strength
+                        && (access_point.max_bitrate > max_bitrate
+                            || (access_point.max_bitrate == max_bitrate
+                                && (access_point.frequency > frequency
+                                    || (access_point.frequency == frequency
+                                        && access_point.strength > strength))))
```
MalpenZibo:
```suggestion
                                        && access_point.strength >= strength))))
```

To avoid AP replacement if everything is equal?

romanstingler:
I refactored the duplicated code. 

It works for NM and on IWD
not tested for NM with IWD backend.

Issue #493, opened 2026-02-15
Ashell Layer Options


1randomguy:
Is there a reason for only allowing "Bottom" and "Overlay" for the layers on which ashell may be displayed?
From my perspective, choosing the Bottom Layer as a default seems like a bit of an odd choice to be honest.

I use niri, and when going to the overview, where you zoom out, the bar also scrolls out, and when you switch workspaces, graphically it looks like changing to a different status bar.
As I understand it (https://variety4me.github.io/niri_docs/Layer-Shell-Components/), putting ashell on the Top layer would give a more logical behavior, where the status bar is always stuck in place, but gets covered by fullscreen apps (unlike when I put it on the overlay layer, where it covers part of the fullscreen application).

I don't know about other window managers behaviours, so maybe there is a reason why Bottom is the default, in that case I don't mind keeping Bottom as the default, otherwise I would argue for making Top the default.

Either way adding the option to put it on the Top layer seems logical to me, and I suspect that it would not require many codechanges, in fact it's something I might try adding myself, as a first contribution to this great project.

PR #494, opened 2026-02-15
Allow configuration of ashell on top-layer

1randomguy:
#493 for reference.

In my opinion putting a bar on the Top layer, such that it is always on top except for fullscreen windows, makes the most sense for the default setting of a status bar.

In niri, for example, scrolling out to the overview also makes ashell small and appear on each workspace individually.
Also, when comparing to something like windows, the bar is always displayed above any window that you move over it.

If, however, it is your personal wish @MalpenZibo to keep bottom as the default for some reason, I will of course accept that.
I hope I made it reasonably clear why the Top layer might be the most sensible choice for some (if not most) users, and therefore the value of adding this at least as a configurable option.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -618,6 +618,7 @@ pub enum Position {
 pub enum Layer {
     #[default]
     Bottom,
+    Top,
     Overlay,
 }
 
--- a/src/outputs.rs
+++ b/src/outputs.rs
@@ -82,6 +82,7 @@ impl Outputs {
         let height = Self::get_height(style, scale_factor);
 
         let iced_layer = match layer {
+            config::Layer::Top => Layer::Top,
             config::Layer::Bottom => Layer::Bottom,
             config::Layer::Overlay => Layer::Overlay,
         };
--- a/website/docs/configuration/main.md
+++ b/website/docs/configuration/main.md
@@ -102,6 +102,7 @@ Configure the bar position and Wayland layer.
 ### Layer Options
 
 - `"Overlay"` - Above everything including fullscreen
+- `"Top"` - Above everything excluding fullscreen
 - `"Bottom"` - Above background, below windows (default)
 
 ### Examples


1randomguy:
I looked at some other status bars, and the most common default seems to be the top layer, such as in:
- noctalia-shell (not even configuarable)
- hyprpanel [(ref)](https://github.com/Jas-SinghFSU/HyprPanel/blob/0e73df1dfedf0f6fa21ed0ae5e031b0663c8f400/src/components/bar/layout/BarLayout.tsx#L171)
- ironbar [(ref)](https://github.com/JakeStanger/ironbar/wiki/configuration-guide)

just to name a few.

The only bar that I saw in my little research which used the Bottom Layer as default was waybar, but I wouldn't call that a status bar with particularly great defaults, I don't think anyone uses that without changing a lot of the config.

MalpenZibo:
I don't know much about Niri, so for the `scrolling out to the overview` I had to see a screenshot of the problem to get an idea.

The reason behind the default to the `Bottom` layer was if I remember correctly because I prefer to have the bar under floating window. That said we could add the option to have the bar on the Top layer. 

If the Bottom layer doesn't create issues I prefer to leave it as default value. 

Can you share some screenshot of the problem?

1randomguy:
this is what the overview looks like with the default bottom layer
image

and this is what overlay/top layer look like
image

Another difference is that when switching between workspaces without going to the overview the bar stays stationary at the top of the screen in top/overlay layer.
But on the bottom layer is behaves more like its part of the background, so you can see it disappear to the top bottom with the resto of the windows/ background in the animation, and then reappear with all of the next workspaces stuff, I hope you can understand what I mean, not sure how to illustrate that.

clotodex:
I tested it on niri, never even thought about this.
It works.
My personal preference (also since I am used to it) is the background layer. But it does really *feel* like two experiences so I would be in favor of configuration

1randomguy:
I changed it back to bottom as the default layer, as I get the feeling that you still prefer it like that, and only added the top layer as a new option now.

clotodex:
lgtm

Issue #347, opened 2025-12-07
feature request: active workspace diffrent colour


maikelthedev:
Just that, being able to have the activ eowrkspace with a different colour instead of different width.

PR #495, opened 2026-02-16
Add configurable window icons to workspace indicators

gwiazdorrr:
Display application icons next to workspace names, similar to waybar's window-rewrite feature.

New config options:
```
[workspaces]
indicator_format = "NameAndIcons"  # or "Name" (default)
window_icons = { "firefox" = "󰈹", "code" = "󰨞" }
default_window_icon = "󰏗"
```

Window class to icon mappings use case-insensitive matching with an optional default icon. Window class collection from compositors is gated behind an AtomicBool flag so the feature has no cost when disabled.

This is how it looks on my setup. I find this extremely useful:
image


1randomguy:
I like the idea of being able to see the what windows are open on which workspace.
But the way I understand it with you implementation you have to specify an icon for every application or it is gonna use the fallback icon?

I feel like that goes against the idea of ashell being a "ready to go" status bar that works out of the box without any configuration.

clotodex:
Agreeing with @1randomguy 
However, the tray implementation already has a way to look up xdg icons from the application and display them. Maybe that works for this case too?

gwiazdorrr:
I hijacked XDG icon handling from the tray module and added some caching. 

image

Got to say, I was skeptical initially. I like nerdfonts, but it turns out I like things just working out of the box without config even more. I removed the ability to override icons completely, so I guess the PR name should be updated?

1randomguy:
I tried it (niri), and I noticed that not all programs are shown, for example nautilus, gnome-calendar, and spotify do not show an icon.

One more thing that I noticed is that it shows only one icon per program per workspace, I think this is something worth thinking about whether we could show that there are multiple windows open.

Lastly, I think that the icons making the indicators bigger makes it kind of difficult to distinguishe the currently active workspace from the others.
I think there was an idea sometime to show the active workspace in a different color, maybe this would be something worth revisiting in this context.

clotodex:
@1randomguy @gwiazdorrr potentially #475 has some fixes in icon discovery that also help here. Not sure if what you mention is an icon or process discovery issue though

src/modules/workspaces.rs
```diff
@@ -31,12 +32,38 @@ pub struct UiWorkspace {
...
+
+    let mut icons = Vec::new();
+    let mut seen = HashSet::new();
+
+    for class in window_classes {
+        let class_lower = class.to_lowercase();
+        if !seen.insert(class_lower.clone()) {
+            continue;
+        }
+        if let Some(xdg) = xdg_icons::get_icon_from_name(&class_lower) {
```
1randomguy: if we can't get an icon, we might want to put a placeholder?

1randomguy:
It does indeed seem like it can't find the icon for a bunch of them, so those fixes will probably help a lot. 

Idk if it will be able to find every icon though, maybe we could consider adding a placeholder icon like the circle that appears in the tray for programs where the icon is not found.
image

1randomguy:
I had a look at #475 , and it seems like he uses the statusnotifier icon first and falls back to the `.desktop` icon if not available. 

For this one, I'd argue that using the `.desktop` icon as the first choice probably makes more sense, right?
But it could still use the functions for getting that `.desktop` icon from the other pr.

gwiazdorrr:
> I tried it (niri), and I noticed that not all programs are shown, for example nautilus, gnome-calendar, and spotify do not show an icon.

From the list I can verify that spotify works here, but the icon for nautilus is missing. I'll investigate.

In the meantime, I rebased and pushed support for both fallback icon (using `StaticIcon::Point`) and `active_workspace_colors` appearance option

1randomguy:
> From the list I can verify that spotify works here

That's very interesting, still doesn't work here :P
But I would still vote for using .desktop icons as a first choice for this one, or do you think that thats not a good idea?

The fallback icon and color for active workspaces (#347 ) work great though :+1: 

image

1randomguy:
One more thing that I was wondering about is whether or not a small space or visual divider between the workspace name and the icons would help with legibility, as I feel like the numbers of the workspaces can get a bit harder to make out when there are a lot of icons. But thats just an idea

gwiazdorrr:
Fixed the missing nautilus icon, hopefully spotify now also works for you :)

1randomguy:
unfortunately still not working for me 😖
image

gwiazdorrr:
I've just added parsing of .desktop files that fixed the last "dot icon" for me (cura).

1randomguy:
image
🥳

src/modules/workspaces.rs
```diff
@@ -31,12 +32,38 @@ pub struct UiWorkspace {
...
     pub active: bool,
     pub windows: u16,
+    pub window_classes: Vec<String>,
+}
+
+fn resolve_workspace_icons(
+    window_classes: &[String],
+    config: &WorkspacesModuleConfig,
+) -> Vec<XdgIcon> {
+    if config.indicator_format != WorkspaceIndicatorFormat::NameAndIcons {
```
clotodex: i would optionally call this function instead of this
especially since the config parameter is literally only necessary for this if i read the code correctly

src/modules/workspaces.rs
```diff
@@ -31,12 +32,38 @@ pub struct UiWorkspace {
...
+        return Vec::new();
+    }
+
+    let mut icons = Vec::new();
+    let mut seen = HashSet::new();
+
+    for class in window_classes {
```
clotodex: I feel like there is a cleaner implementation using iterators (collecting inbetween to hashset if itertools is not a dependency)

```
let icons = window_classes
    .iter()
    .map(|class| class.to_lowercase())
    .unique()
    .map(|class_lower| {
        xdg_icons::get_icon_from_name(&class_lower).unwrap_or_else(xdg_icons::fallback_icon)
    })
    .collect::<Vec<_>>();
```
This _should_ also be a tiny bit more performant compared to pushing and inserting.

src/modules/workspaces.rs
```diff
@@ -31,12 +32,38 @@ pub struct UiWorkspace {
     pub monitor: String,
     pub displayed: Displayed,
     pub windows: u16,
+    pub icons: Vec<XdgIcon>,
```
clotodex: Would make this an Option<Vec<...>> to not collect it if the setting is off

src/modules/workspaces.rs
```diff
@@ -240,8 +273,13 @@ fn calculate_ui_workspaces(
     result
 }
 
+fn needs_window_classes(config: &WorkspacesModuleConfig) -> bool {
```
clotodex: this function makes me feel funny, not sure if worth the deduplication

src/modules/workspaces.rs
```diff
@@ -240,8 +273,13 @@ fn calculate_ui_workspaces(
...
+    config.indicator_format == WorkspaceIndicatorFormat::NameAndIcons
+}
+
 impl Workspaces {
     pub fn new(config: WorkspacesModuleConfig) -> Self {
+        crate::services::compositor::set_collect_window_classes(needs_window_classes(&config));
```
clotodex: please use a `use` for this function

src/modules/workspaces.rs
```diff
@@ -433,48 +472,93 @@ impl Workspaces {
                                 w.monitor_id
                             };
 
+                            let is_active = w.displayed == Displayed::Active;
+
                             let color = color_index.map(|i| {
-                                if w.id > 0 {
-                                    theme.workspace_colors.get(i as usize).copied()
-                                } else {
+                                let i = i as usize;
```
clotodex: please inline this to line 492
`colors.get(i as usize).copied()`
i think it is the only use

src/modules/workspaces.rs
```diff
@@ -433,48 +472,93 @@ impl Workspaces {
...
-                                .padding(if w.id < 0 {
-                                    match w.displayed {
-                                        Displayed::Active => [0, theme.space.md],
-                                        Displayed::Visible => [0, theme.space.sm],
-                                        Displayed::Hidden => [0, theme.space.xs],
+                            let has_icons = !w.icons.is_empty();
+                            let dynamic_width = w.id < 0 || has_icons;
+
+                            let content: Element<'a, Message> = if has_icons {
+                                let mut children: Vec<Element<'a, Message>> =
```
clotodex: a non-mut non-extend approach could be:
```
let children: Vec<Element<'a, Message>> = iter::once(
    text(w.name.as_str()).size(theme.font_size.xs).into()
)
.chain(w.icons.iter().map(|i| match i {
    XdgIcon::Svg(handle) => Svg::new(handle.clone())
        .height(Length::Fixed(theme.font_size.xs as f32))
        .width(Length::Shrink)
        .into(),
    XdgIcon::Image(handle) => Image::new(handle.clone())
        .height(Length::Fixed(theme.font_size.xs as f32))
        .width(Length::Shrink)
        .into(),
    XdgIcon::NerdFont(glyph) => text(*glyph)
        .size(theme.font_size.xs)
        .font(Font::with_name("Symbols Nerd Font"))
        .into(),
}))
.collect();```

**Review by clotodex — CHANGES_REQUESTED** [review]
As far as I got today with review

**Comment by clotodex** [comment]
please fix the conflict :)

**Comment by clotodex** [comment]
also generally check out #537 for performance, maybe this PR can adopt some of the same improvements

File: src/services/compositor/mod.rs [hunk 9]
```diff
@@ -17,6 +18,16 @@ const BROADCAST_CAPACITY: usize = 64;
 static BROADCASTER: OnceCell<broadcast::Sender<ServiceEvent<CompositorService>>> =
     OnceCell::const_new();
 
+static COLLECT_WINDOW_CLASSES: AtomicBool = AtomicBool::new(false);
```
clotodex, line 29: I feel weird about using an atomic boolean to essentially communicate a config field

src/modules/workspaces.rs
```diff
@@ -31,12 +34,25 @@ pub struct UiWorkspace {
...
+    pub window_classes: Vec<String>,
+}
+
+fn resolve_workspace_icons(window_classes: &[String]) -> Vec<XdgIcon> {
+    window_classes
+        .iter()
+        .map(|class| class.to_lowercase())
+        .unique()
+        .map(|class_lower| {
+            xdg_icons::get_icon_from_name(&class_lower).unwrap_or_else(xdg_icons::fallback_icon)
```
clotodex, line 59: this swallows errors, is there a way to somehow only swallow "not found"?

src/services/xdg_icons.rs
```diff
@@ -0,0 +1,335 @@
...
+use log::debug;
+use std::{
+    collections::{BTreeSet, HashMap},
+    env, fs,
+    path::{Path, PathBuf},
+    sync::{LazyLock, Mutex},
+};
+
+static ICON_CACHE: LazyLock<Mutex<HashMap<String, Option<XdgIcon>>>> =
+    LazyLock::new(|| Mutex::new(HashMap::new()));
```
clotodex: is an RwLock not better than a mutex? few writes many reads?

src/modules/workspaces.rs
```diff
@@ -242,6 +265,9 @@ fn calculate_ui_workspaces(
 
 impl Workspaces {
     pub fn new(config: WorkspacesModuleConfig) -> Self {
+        set_collect_window_classes(
```
clotodex, line 335: as discussed with the atomic bool, this is a strong anti-pattern and a new messaging channel for config which we should not do

clotodex (changes requested):
We are getting there! Thank you for the contributions

Looking at the code this feels in many cases AI generated. This is not a problem, but makes it REALLY hard to understand the code quickly, and every change and review changes a lot of things again needing to understand the code from scratch. Not sure how to deal with this yes.

@romanstingler could you do me a favor and check out the xdg_icons.rs file, since you worked on the tray icons and icon discovery (if not, feel free to tag who did if I am mistaken)

romanstingler:
I applied my performance changes and a few other suggestions to the
XDG file. 

I think that we can rework the whole thing in future, 
and with more testing maybe get rid of some fallback strategies and maybe 
working only with the desktop files is ok. 
I have no idea, there are so many strange apps, so many themes, .....

There are things missing (also in the tray module) where we invalidate the caches if the user changes the theme of the system. 
Tbh, not a blocker for me for now. 

overall works fine 
image



```
use freedesktop_icons::lookup;
use iced::widget::{image, svg};
use linicon_theme::get_icon_theme;
use log::debug;
use std::{
    borrow::Cow,
    collections::{BTreeSet, HashMap},
    env,
    ffi::OsString,
    fs,
    path::{Path, PathBuf},
    sync::LazyLock,
};

use std::sync::RwLock;

const MAX_SIMILAR_ICON_CANDIDATES: usize = 5;

static ICON_CACHE: LazyLock<RwLock<HashMap<String, Option<XdgIcon>>>> =
    LazyLock::new(|| RwLock::new(HashMap::new()));
static SYSTEM_ICON_NAMES: LazyLock<BTreeSet<OsString>> = LazyLock::new(load_system_icon_names);
static SYSTEM_ICON_ENTRIES: LazyLock<Vec<(Cow<'static, str>, Cow<'static, str>)>> =
    LazyLock::new(|| {
        SYSTEM_ICON_NAMES
            .iter()
            .filter_map(|name| {
                let name_str = name.to_str()?;
                let normalized = normalize_icon_name(name_str);
                let normalized_cow = if normalized.as_ref() == name_str {
                    Cow::Borrowed(name_str)
                } else {
                    Cow::Owned(normalized.into_owned())
                };
                Some((Cow::Owned(name_str.to_string()), normalized_cow))
            })
            .collect()
    });
static DESKTOP_ICON_INDEX: LazyLock<HashMap<String, String>> =
    LazyLock::new(build_desktop_icon_index);

#[derive(Debug, Clone)]
pub enum XdgIcon {
    Image(image::Handle),
    Svg(svg::Handle),
    NerdFont(&'static str),
}

pub fn fallback_icon() -> XdgIcon {
    XdgIcon::NerdFont(crate::components::icons::StaticIcon::Point.get_str())
}

pub fn get_icon_from_name(icon_name: &str) -> Option<XdgIcon> {
    if icon_name.is_empty() {
        return None;
    }

    let cache = match ICON_CACHE.read() {
        Ok(c) => c,
        Err(_) => {
            return lookup_icon(icon_name);
        }
    };

    if let Some(cached) = cache.get(icon_name) {
        return cached.clone();
    }
    drop(cache); // Release read lock before write

    let mut cache = match ICON_CACHE.write() {
        Ok(c) => c,
        Err(_) => {
            return lookup_icon(icon_name);
        }
    };

    // Double-check after acquiring write lock (another thread may have populated it)
    if let Some(cached) = cache.get(icon_name) {
        return cached.clone();
    }

    let result = lookup_icon(icon_name);
    cache.insert(icon_name.to_string(), result.clone());
    result
}

fn lookup_icon(icon_name: &str) -> Option<XdgIcon> {
    if let Some(path) = find_icon_path(icon_name) {
        debug!("icon '{icon_name}': direct match at {path:?}");
        return icon_from_path(path);
    }
    debug!("icon '{icon_name}': no direct match");

    if let Some(candidates) = find_similar_icon(icon_name) {
        for candidate in candidates {
            debug!("icon '{icon_name}': similar candidate '{candidate}'");
            if let Some(path) = find_icon_path(&candidate) {
                return icon_from_path(path);
            }
        }
    }
    debug!("icon '{icon_name}': no similar match");

    if let Some(path) = find_desktop_icon(icon_name) {
        debug!("icon '{icon_name}': desktop index match at {path:?}");
        return icon_from_path(path);
    }
    debug!("icon '{icon_name}': no desktop index match");

    if let Some(prefix_candidate) = prefix_match_icon(icon_name)
        && let Some(path) = find_icon_path(&prefix_candidate)
    {
        debug!("icon '{icon_name}': prefix match '{prefix_candidate}' at {path:?}");
        return icon_from_path(path);
    }
    debug!("icon '{icon_name}': no prefix match — unresolved");

    None
}

fn icon_from_path(path: PathBuf) -> Option<XdgIcon> {
    if path.extension().is_some_and(|ext| ext == "svg") {
        debug!("svg icon found. Path: {path:?}");

        Some(XdgIcon::Svg(svg::Handle::from_path(path)))
    } else {
        debug!("raster icon found. Path: {path:?}");

        Some(XdgIcon::Image(image::Handle::from_path(path)))
    }
}

fn find_icon_path(icon_name: &str) -> Option<PathBuf> {
    let base_lookup = lookup(icon_name).with_cache();

    match get_icon_theme() {
        Some(theme) => base_lookup.with_theme(&theme).find().or_else(|| {
            let fallback_lookup = lookup(icon_name).with_cache();
            fallback_lookup.find()
        }),
        None => base_lookup.find(),
    }
}

fn find_similar_icon(icon_name: &str) -> Option<Vec<Cow<'static, str>>> {
    if SYSTEM_ICON_ENTRIES.is_empty() {
        return None;
    }

    let normalized = normalize_icon_name(icon_name);
    let normalized_no_separators = strip_icon_separators(normalized.as_ref());
    let mut matches: Vec<Cow<'static, str>> = Vec::with_capacity(MAX_SIMILAR_ICON_CANDIDATES);

    for (candidate_name, candidate_normalized) in SYSTEM_ICON_ENTRIES.iter() {
        if candidate_normalized.as_ref() == normalized.as_ref() {
            continue;
        }

        if candidate_normalized.as_ref().contains(normalized.as_ref())
            || normalized.as_ref().contains(candidate_normalized.as_ref())
            || candidate_normalized
                .as_ref()
                .contains(normalized_no_separators.as_ref())
        {
            matches.push(Cow::Borrowed(candidate_name.as_ref()));
            if matches.len() >= MAX_SIMILAR_ICON_CANDIDATES {
                break;
            }
        }
    }

    if matches.is_empty() {
        None
    } else {
        Some(matches)
    }
}

fn normalize_icon_name(name: &str) -> Cow<'_, str> {
    if name
        .chars()
        .all(|c| c.is_ascii_lowercase() || c.is_ascii_digit())
    {
        return Cow::Borrowed(name);
    }

    Cow::Owned(
        name.to_lowercase()
            .chars()
            .filter(|c| c.is_ascii_alphanumeric())
            .collect(),
    )
}

fn strip_icon_separators(name: &str) -> Cow<'_, str> {
    if name.bytes().all(|byte| byte != b'-' && byte != b'_') {
        return Cow::Borrowed(name);
    }

    Cow::Owned(name.chars().filter(|ch| *ch != '-' && *ch != '_').collect())
}

fn prefix_match_icon(icon_name: &str) -> Option<Cow<'static, str>> {
    if SYSTEM_ICON_ENTRIES.is_empty() {
        return None;
    }

    let normalized = normalize_icon_name(icon_name);

    if let Some(exact) = SYSTEM_ICON_ENTRIES
        .iter()
        .find(|(_, norm)| norm.as_ref() == normalized.as_ref())
    {
        return Some(Cow::Borrowed(exact.0.as_ref()));
    }

    let mut candidates: Vec<_> = SYSTEM_ICON_ENTRIES.iter().collect();
    for (idx, ch) in normalized.chars().enumerate() {
        candidates.retain(|(_, name)| name.chars().nth(idx) == Some(ch));

        if candidates.len() == 1 {
            return Some(Cow::Borrowed(candidates[0].0.as_ref()));
        }

        if candidates.is_empty() {
            break;
        }
    }

    candidates
        .first()
        .map(|(name, _)| Cow::Borrowed(name.as_ref()))
}

fn find_desktop_icon(icon_name: &str) -> Option<PathBuf> {
    let normalized = normalize_icon_name(icon_name);
    let normalized_str: String = normalized.into_owned();
    let Some(icon_value) = DESKTOP_ICON_INDEX.get(&normalized_str) else {
        debug!("icon '{icon_name}': normalized '{normalized_str}' not in desktop index");
        return None;
    };
    debug!("icon '{icon_name}': desktop index '{normalized_str}' → '{icon_value}'");

    if icon_value.starts_with('/') {
        let path = PathBuf::from(icon_value);
        if path.exists() {
            Some(path)
        } else {
            debug!("icon '{icon_name}': absolute path '{icon_value}' does not exist");
            None
        }
    } else {
        find_icon_path(icon_value)
    }
}

fn build_desktop_icon_index() -> HashMap<String, String> {
    let mut map = HashMap::new();

    for dir in desktop_application_dirs() {
        if !dir.is_dir() {
            continue;
        }
        if let Ok(entries) = fs::read_dir(&dir) {
            for entry in entries.flatten() {
                let path = entry.path();
                if path.extension().and_then(|e| e.to_str()) != Some("desktop") {
                    continue;
                }
                parse_desktop_file(&path, &mut map);
            }
        }
    }

    map
}

const MAX_DESKTOP_FILE_SIZE: u64 = 64 * 1024;

fn parse_desktop_file(path: &Path, map: &mut HashMap<String, String>) {
    let Ok(metadata) = fs::metadata(path) else {
        return;
    };

    if metadata.len() > MAX_DESKTOP_FILE_SIZE {
        debug!("desktop file too large: {}", path.display());
        return;
    }

    let Ok(contents) = fs::read_to_string(path) else {
        return;
    };

    let stem = path
        .file_stem()
        .and_then(|s| s.to_str())
        .unwrap_or_default();

    let mut icon_value: Option<String> = None;
    let mut wm_class: Option<String> = None;

    for line in contents.lines() {
        if let Some(val) = line.strip_prefix("Icon=") {
            icon_value = Some(val.trim().to_string());
        } else if let Some(val) = line.strip_prefix("StartupWMClass=") {
            wm_class = Some(val.trim().to_string());
        }
    }

    let Some(icon) = icon_value else {
        return;
    };

    if let Some(wm) = wm_class {
        let key = normalize_icon_name(&wm).into_owned();
        map.entry(key).or_insert_with(|| icon.clone());
    }
    let parts: Vec<&str> = stem.split('.').collect();
    for start in 0..parts.len() {
        let key = normalize_icon_name(&parts[start..].join(".")).into_owned();
        map.entry(key).or_insert_with(|| icon.clone());
    }
}

fn desktop_application_dirs() -> Vec<PathBuf> {
    let mut dirs = Vec::new();

    if let Ok(data_home) = env::var("XDG_DATA_HOME") {
        dirs.push(PathBuf::from(data_home).join("applications"));
    }

    if let Ok(home) = env::var("HOME") {
        dirs.push(PathBuf::from(home).join(".local/share/applications"));
    }

    let data_dirs =
        env::var("XDG_DATA_DIRS").unwrap_or_else(|_| "/usr/local/share:/usr/share".into());
    for dir in data_dirs.split(':') {
        if !dir.is_empty() {
            dirs.push(PathBuf::from(dir).join("applications"));
        }
    }

    dirs.push(PathBuf::from("/usr/local/share/applications"));
    dirs.push(PathBuf::from("/usr/share/applications"));

    dirs.sort();
    dirs.dedup();
    dirs
}

fn load_system_icon_names() -> BTreeSet<OsString> {
    let mut names = BTreeSet::new();

    for dir in icon_directories() {
        if !dir.is_dir() {
            continue;
        }

        collect_icon_names_recursive(&dir, &mut names);
    }

    names
}

fn collect_icon_names_recursive(dir: &Path, names: &mut BTreeSet<OsString>) {
    if let Ok(entries) = fs::read_dir(dir) {
        for entry in entries.flatten() {
            let path = entry.path();
            if let Ok(file_type) = entry.file_type() {
                if file_type.is_dir() {
                    collect_icon_names_recursive(&path, names);
                } else if file_type.is_file()
                    && let Some(stem) = path.file_stem()
                {
                    names.insert(stem.to_os_string());
                }
            }
        }
    }
}

fn icon_directories() -> Vec<PathBuf> {
    let mut dirs = Vec::new();

    if let Ok(data_home) = env::var("XDG_DATA_HOME") {
        let base = PathBuf::from(data_home);
        dirs.push(base.join("icons"));
        dirs.push(base.join("pixmaps"));
    }

    if let Ok(home) = env::var("HOME") {
        let base = PathBuf::from(home);
        dirs.push(base.join(".local/share/icons"));
        dirs.push(base.join(".local/share/pixmaps"));
    }

    let data_dirs =
        env::var("XDG_DATA_DIRS").unwrap_or_else(|_| "/usr/local/share:/usr/share".into());
    for dir in data_dirs.split(':') {
        if dir.is_empty() {
            continue;
        }
        let base = PathBuf::from(dir);
        dirs.push(base.join("icons"));
        dirs.push(base.join("pixmaps"));
    }

    dirs.push(PathBuf::from("/usr/share/icons"));
    dirs.push(PathBuf::from("/usr/share/pixmaps"));

    dirs.sort();
    dirs.dedup();
    dirs
}
```

gwiazdorrr:
> Looking at the code this feels in many cases AI generated. This is not a problem, but makes it REALLY hard to understand the code quickly, and every change and review changes a lot of things again needing to understand the code from scratch. Not sure how to deal with this yes.

This PR does contain some AI assistance. The "AI code review fatigue" is something I struggle with myself at work, sorry for bringing this upon you. Rebasing surely did not help the clarity, either. The solution is, well, higher standards on the submitter's end.

src/services/compositor/mod.rs
gwiazdorrr: Need to familiarize myself more with tokio. Or just make hyprland/niri always collect windows classes and process them in workspaces.rs. Not sure which is the right call atm.

src/modules/workspaces.rs
gwiazdorrr: I don't think there anything other than "not found" at this stage; iced images are loaded lazily and all the debug logs are really just progress, not real failures.

src/components/icons.rs
```diff
@@ -125,7 +125,7 @@ pub enum StaticIcon {
 }
 
 impl StaticIcon {
-    fn get_str(&self) -> &'static str {
+    pub fn get_str(&self) -> &'static str {
```
MalpenZibo: It's needed?

src/config.rs
```diff
@@ -532,6 +540,7 @@ pub struct Appearance {
     pub text_color: AppearanceColor,
     pub workspace_colors: Vec<AppearanceColor>,
     pub special_workspace_colors: Option<Vec<AppearanceColor>>,
+    pub active_workspace_colors: Option<Vec<AppearanceColor>>,
```
MalpenZibo: This new config? Is it related to the window icon? Could be placed in a dedicated PR?

src/services/compositor/mod.rs
MalpenZibo: So every kind of configuration should be sent in the `subscribe()` service function. In theory, it should also be used in the `Subscription::run_with(TypeId::of::<Self>()` as id, for example `run_with((TypeId::of::<Self>(), configuration_x), ....` This tells iced to recreate the subscription if the id changes.

TypeId::of::<Self>() is constant, so adding the `configuration_x` in the tuple should recreate the subscription only when the `configuration_x` value changes.

In this case, we should send an `indicator_format == WorkspaceIndicatorFormat::NameAndIcons` to know if we had to collect the class name, if I understand correctly the changes.

src/modules/workspaces.rs
MalpenZibo: This part should go in the `subscription` method of the module. We should pass `config.indicator_format == WorkspaceIndicatorFormat::NameAndIcons` as parameter to the service subscribe method

MalpenZibo (commented):
I see the new `xdg_icons.rs`. Please be sure to NOT introduce regression to the Tray module. I notice that you deleted some code from the tray module. I presume that xdg_icons is the new home of that code. Am I right, or are there some changes?

src/modules/workspaces.rs
```diff
@@ -433,48 +462,95 @@ impl Workspaces {
...
+                                            .width(Length::Shrink)
+                                            .into(),
+                                        XdgIcon::Image(handle) => Image::new(handle.clone())
+                                            .height(Length::Fixed(theme.font_size.xs as f32))
+                                            .width(Length::Shrink)
+                                            .into(),
+                                        XdgIcon::NerdFont(glyph) => text(*glyph)
+                                            .size(theme.font_size.xs)
+                                            .font(Font::with_name("Symbols Nerd Font"))
+                                            .into(),
```
MalpenZibo: We should use the icons. Check the `DynamicIcon`

src/modules/workspaces.rs
MalpenZibo:
```suggestion
        for w in normal {
```

src/modules/workspaces.rs
```diff
@@ -115,12 +133,14 @@ fn calculate_ui_workspaces(
             if let Some(vdesk) = virtual_desktops.get_mut(&vdesk_id) {
                 vdesk.windows += w.windows;
                 vdesk.active = vdesk.active || is_active;
+                vdesk.window_classes.extend(w.window_classes.clone());
```
MalpenZibo:
```suggestion
                vdesk.window_classes.extend(w.window_classes);
```

src/modules/workspaces.rs
```diff
@@ -115,12 +133,14 @@ fn calculate_ui_workspaces(
...
             } else {
                 virtual_desktops.insert(
                     vdesk_id,
                     VirtualDesktop {
                         active: is_active,
                         windows: w.windows,
+                        window_classes: w.window_classes.clone(),
```
MalpenZibo:
```suggestion
                        window_classes: w.window_classes,
```

src/services/xdg_icons.rs
```diff
@@ -0,0 +1,414 @@
...
+}
+
+fn find_desktop_icon(icon_name: &str) -> Option<PathBuf> {
+    let normalized = normalize_icon_name(icon_name);
+    let normalized_str: String = normalized.into_owned();
+    let Some(icon_value) = DESKTOP_ICON_INDEX.get(&normalized_str) else {
+        debug!("icon '{icon_name}': normalized '{normalized_str}' not in desktop index");
+        return None;
+    };
+    debug!("icon '{icon_name}': desktop index '{normalized_str}' → '{icon_value}'");
```
MalpenZibo:
```suggestion
    let normalized = normalize_icon_name(icon_name);
    let Some(icon_value) = DESKTOP_ICON_INDEX.get(normalized.as_ref()) else {
        debug!("icon '{icon_name}': normalized '{normalized}' not in desktop index");
        return None;
    };
    debug!("icon '{icon_name}': desktop index '{normalized}' → '{icon_value}'");

```

src/modules/workspaces.rs
MalpenZibo:
```suggestion
        for w in special {
```

MalpenZibo:
```suggestion
                monitor: w.monitor,
```

MalpenZibo:
```suggestion
        for w in normal {
```

MalpenZibo:
```suggestion
                    .or(Some(w.name))
```

MalpenZibo:
```suggestion
                w.name
```

MalpenZibo:
```suggestion
                monitor: w.monitor,
```

MalpenZibo (commented):
Some changes are not directly related to this PR, but let's fix them anyway 😄

src/components/icons.rs
gwiazdorrr: Yes, for the fallback icon (the dreaded dot).

EDIT: not needed anymore, reverted.

src/config.rs
gwiazdorrr: I could extract it, but: it is only needed for when icons are enabled, as then with the visual clutter you might not notice the "capsule" resizing.

src/modules/workspaces.rs
gwiazdorrr: Seems just using StaticIcon works even better.

1randomguy:
Any updates here, are there still blockers that prevent this from being merged?

At least the option to set the `active_workspace_color` to be different is something I would very much appreciate, not just for this PRs context. 
When using named workspaces on niri it becomes borderline impossible to see what workspace you are currently on if the only differenciating factor is a bit of padding to the left and right of the name.

If this PR isn't being merged soon, I'd like to pull out at least that feature to a new PR and get that merged 😅

gwiazdorrr:
The only blocker is that I've been absolutely swamped with other things. Since rebases are not getting any simpler with time, I plan to tag/archive the latest commit, reset the branch to the latest upstream and start over, without the baggage.

So to answer your question - go ahead, rip out whatever you see fit. Honestly, that feature should've been a separate PR to begin with.

MalpenZibo:
Sorry for the long delay. My bad, too many things to do with not enough time.

Anyway, if you're ok with that:
- I'm working on a rebase
- I found some performance concerns related to the .desktop icon retrieval process, and I will try to fix them here.
- I will also open a new PR to update the documentation about this feature

MalpenZibo (approved):

MalpenZibo:
image

PR #497, opened 2026-02-19
1| feat: add password visibility toggle to WiFi password dialog

romanstingler:
xdph_screenshot_10111c53
not my real pw, but would probably meet some pw policies :P


MalpenZibo:
:rocket:

PR #499, opened 2026-02-19
2| Feature: open network warning

romanstingler:
This one sounds simple but I hope we can have some discussion about that topic. 
First and foremost, be warned I am really bad at naming things and even after naming things I am unsure if they are properly named :) 

Note: This branch is on top of https://github.com/MalpenZibo/ashell/pull/497 otherwise rebasing is always a lot of work.

I noticed that there is a password prompt for open networks. 
In my opinion we should warn users of the risks of open networks and not ask for a password.

Therefore, **for now** I have put the content for open networks in the `src/password_dialog.rs`.
Here comes the first pool of questions: 
* do we want to keep that together, if so, do we want to rename that (one argument is that a pw prompt for secured networks is the eloquent to a warning for an open network)
* for now we just support open and normal secured networks, there are enterprise ones which have user,password and enterprise with user, password and cert, do we want to support these at some point? Or do we say let the user do that on the NM of their choice? 
* for now I use `network_dialog` instead of `password_dialog` in the code, depending what we want I will change that accordingly.

There is one fundamental difference between secured and open networks, 
I kept that the secured network PW prompt is not shown if it is in the known networks, 
but I changed that for open networks that they always show the warning/info 
(the same SSID could be used somewhere else and I don't want users to automatically connect). 

The current warning looks like that for now (sorry multiple bars open)

xdph_screenshot_1fc07593

one small change is that I disabled the confirm button on secured network PW prompts if the PW field is empty.

@clotodex @MalpenZibo thanks in advance for your feedback 

I hope this was related to your message in 
https://github.com/MalpenZibo/ashell/issues/91#issue-2809171717

> Sometimes, it asks for a password even if it's not required,

### DIFF
--- a/src/modules/settings/mod.rs
+++ b/src/modules/settings/mod.rs
@@ -44,11 +44,24 @@ pub struct Settings {
     bluetooth: BluetoothSettings,
     idle_inhibitor: Option<IdleInhibitorManager>,
     sub_menu: Option<SubMenu>,
-    password_dialog: Option<(String, String)>,
-    password_dialog_show_password: bool,
+    network_dialog: Option<NetworkDialogState>,
+    network_dialog_show_password: bool,
     indicators: Vec<SettingsIndicator>,
     custom_buttons: Vec<SettingsCustomButton>,
     custom_buttons_status: HashMap<String, Option<bool>>,
+}
+
+#[derive(Debug, Clone)]
+enum NetworkDialogKind {
+    Password,
+    OpenNetworkWarning,
+}
+
+#[derive(Debug, Clone)]
+struct NetworkDialogState {
+    ssid: String,
+    password: Option<String>,
+    kind: NetworkDialogKind,
 }
 
 #[derive(Debug, Clone)]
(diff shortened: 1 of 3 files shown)


clotodex:
I like the UX push
In my eyes if we put out a warning i would go the whole mile and also offer the button
"Connect only this time" or sth shorter
So that it is not saved to known networks

Other than that i looked into all the different ways to configure networks and it is a LOT
In favor of supporting as many as possible, but most providers of such networks have some copy paste instructions available. So i would not worry too much about the priority

MalpenZibo:
For me it's ok! If you want to implement the "connect only this time" behavior, let's do it otherwis,e let's track in a dedicated issue and let's merge this PR

romanstingler:
@MalpenZibo I had a look into that request. 
I created a new open issue for that as it will take some time for me to dig into that (or @clotodex can also have a look or try implementing it if he wants). 

I will try to focus my time on fixing the open PRs that are planned for the release. 
PS: thanks for reviewing.

Issue #450, opened 2026-01-26
Roadmap 0.9


MalpenZibo:
# Short-term goals

- [x] #566 A collection of components for common use cases in the codebase to simplify the reuse of common patterns. Things like:
  - common button: Solid, transparent, outline?
  - slider (the one used in the menu)
  - quicksettings button
  - status bar: Group, Buttons, Label 
  - list of selectable items (the one in the wifi sub-menu for example
- [x] Notifications service 
- [ ] #567 Refactor appearance config.
  - Move away from styles (island, solid, gradient). Add bar margin, padding, and surface background color. These should be enough to simulate the current styles.
  - Spacing option for module group and module items
  - Appearance option for the module should go in a [ModuleName].Appearance section. The global Appearance section should be used only for global stuff. For example, both `workspace_colors` and `special_workspace_colors` could be moved.
- [ ] #568 Explore how to implement some animation [WIP](https://github.com/MalpenZibo/ashell/pull/655)
- [ ] Begin the exploration on how to implement a global Wayland compositor support and how to include it with the current codebase
- [ ] #596 Refactor outputs.rs after iced 14 migration: we could open the menu surface only on request
- [x] #657 Osd service

# Long-term goals
- Command palette: I want to add a command palette to make it easier to interact with ashell without needing a mouse.
- integrated lock screen + Greetd integration

## About Iced

The new `iced_layershell` library seems to be working fine! It's not too complex to maintain, and let us use the official iced library. I think that we could live with that!

About [guido](https://github.com/MalpenZibo/guido), it was a great experiment, and probably we could reach great results with it. But that means porting all we have, resolving a lot of possible regressions, changing the codebase style, every contributor needs to learn a new library (used only by ashell), and we need to maintain an entire GUI library only for a status bar.
Honestly, it seems too much.

romanstingler:
1. Appearance 
I don't know if you want to make some predefined enum
`workspace_button_rounding: RoundingLevel`

and have it like 
```
None → 0 pixels (square corners)
Small → theme.radius.sm (4px)
Medium → theme.radius.md (8px)
Large → theme.radius.lg (16px)
Full → theme.radius.xl (32px)
```
So people could use the square corners and have a i3/sway look.

```
[appearance]
# Workspace button rounding: "none", "small", "medium", "large", "full"
workspace_button_rounding = "large"

[appearance.border]
width = 0.0
# Border color (optional, defaults to workspace color)
color = "#808080"
```
[...]

PR #500, opened 2026-02-20
MangoWC support integration

Sachin-Bhat:
Hi @MalpenZibo,

Owing to the roadmap (#450) where you intended to support more compositors I took a crack at implementing support for MangoWC's ipc system i.e `mmsg`. Mango is a tag-based compositor rather than a workspace based compositor which also supports multiple active-tags. From what I could see the assumption earlier was only one active tag/workspace at any time. I introduced `active_workspace_ids` to indicate this with a fallback to the old behavior when multi-tag isn't present. I have added the relevant docs too. This was tested on `mango 0.12.3(cea2bca)` and this seems to work. 

Would appreciate any feedback!

Thanks
Sachin

### DIFF
--- a/src/services/compositor/mangowc.rs
+++ b/src/services/compositor/mangowc.rs
@@ -1,0 +1,357 @@
+use super::types::{
+    ActiveWindow, ActiveWindowMango, CompositorCommand, CompositorEvent, CompositorMonitor,
+    CompositorService, CompositorState, CompositorWorkspace,
+};
+use crate::services::ServiceEvent;
+use anyhow::{Context, Result, anyhow};
+use std::collections::{HashMap, HashSet};
+use std::process::Command as StdCommand;
+use tokio::{
+    io::{AsyncBufReadExt, BufReader},
+    process::Command,
+    sync::broadcast,
+};
+
+#[derive(Debug, Default, Clone)]
+struct TagInfo {
+    clients: u16,
+    selected: bool,
+}
+
+#[derive(Debug, Default, Clone)]
+struct OutputTagState {
+    tags: HashMap<i32, TagInfo>,
+    selected_mask: u32,
+    occupied_mask: u32,
+    urgent_mask: u32,
+}
+
+pub fn is_available() -> bool {
+    StdCommand::new("mmsg")
+        .args(["-g", "-O"])
+        .output()
+        .map(|out| out.status.success())
+        .unwrap_or(false)
+}
+
+pub async fn execute_command(cmd: CompositorCommand) -> Result<()> {
+    match cmd {
+        CompositorCommand::FocusWorkspace(id) => {
+            run_mmsg(["-s", "-t", &id.to_string()]).await?;
+        }
+        CompositorCommand::FocusSpecialWorkspace(_) => {
+            return Err(anyhow!(
+                "Special workspaces are not supported in the MangoWC backend"
+            ));
+        }
+        CompositorCommand::ToggleSpecialWorkspace(_) => {
+            return Err(anyhow!(
+                "Special workspaces are not supported in the MangoWC backend"
+            ));
+        }
+        CompositorCommand::FocusMonitor(_) => {
+            return Err(anyhow!(
+                "FocusMonitor by ID is not supported in the MangoWC backend"
+            ));
+        }
+        CompositorCommand::ScrollWorkspace(dir) => {
+            let dispatch = if dir > 0 {
+                "viewtoleft_have_client"
+            } else {
+                "viewtoright_have_client"
+            };
+            run_mmsg(["-s", "-d", dispatch]).await?;
+        }
+        CompositorCommand::NextLayout => {
+            run_mmsg(["-s", "-d", "switch_keyboard_layout"]).await?;
+        }
+        CompositorCommand::CustomDispatch(dispatcher, args) => {
+            let full_dispatch = if args.is_empty() {
+                dispatcher
+            } else {
+                format!("{dispatcher},{args}")
+            };
+            run_mmsg(["-s", "-d", &full_dispatch]).await?;
+        }
+    }
+
+    Ok(())
+}
+
+pub async fn run_listener(tx: &broadcast::Sender<ServiceEvent<CompositorService>>) -> Result<()> {
+    send_latest_state(tx).await?;
+
+    let mut child = Command::new("mmsg")
+        .args(["-w", "-t", "-c", "-k", "-b"])
+        .stdout(std::process::Stdio::piped())
+        .spawn()
+        .context("Failed to spawn mmsg watch process")?;
+
+    let stdout = child
+        .stdout
+        .take()
+        .ok_or_else(|| anyhow!("Failed to capture mmsg stdout"))?;
+    let mut reader = BufReader::new(stdout);
+    let mut line = String::new();
+
+    loop {
+        line.clear();
+        let read = reader.read_line(&mut line).await?;
+        if read == 0 {
+            break;
+        }
+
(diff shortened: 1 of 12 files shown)


clotodex:
Happy to test the change on niri so nothing breaks

Could you share a screenshot how the multiple active ids look now?

Sachin-Bhat:
Hey @clotodex,

Thank you for pitching in! I can definitely share what I mean in a screenshot with some additional info. In mango multitag views are possible. Though I dont have it explicitly bound to a keybind in my dotfiles, I did the following:

I ran this:

```bash
mmsg -s -d view,1
mmsg -s -d toggleview,2
mmsg -g -t

eDP-1 tag 1 1 1 1
eDP-1 tag 2 1 1 0
eDP-1 tag 3 0 0 0
eDP-1 tag 4 0 0 0
eDP-1 tag 5 0 0 0
eDP-1 tag 6 0 0 0
eDP-1 tag 7 0 0 0
eDP-1 tag 8 0 0 0
eDP-1 tag 9 0 0 0
eDP-1 clients 2
eDP-1 tags 3 3 0
eDP-1 tags 000000011 000000011 000000000
```

before taking the screenshot.

`MangoWC multi-tag view (1|2): ashell correctly renders multiple active workspace IDs

Here tag 1 and 2 are active as per the terminal output and ashell correctly shows both tags/workspaces active at the same time which was not possible earlier.

Sachin-Bhat:
Do let me know if you need any clarifications. I would probably need some assistance in having this tested in Niri and Hyprland to ensure that there are no regressions.

clotodex:
I dont know how mango does multi monitor but you probably should test that as well (since the workspaces module was recently rewritten to allow for that) - should probabky just work though

Otherwise looks good :)
Havent done a code review yet though, so let me/us know when you are ready for review

Testing on niri now

clotodex:
Works on niri :)

Sachin-Bhat:
> I dont know how mango does multi monitor but you probably should test that as well (since the workspaces module was recently rewritten to allow for that) - should probabky just work though

Thanks for raising this. I currently don’t have access to a multi-monitor MangoWC setup, so this PR focuses only on correctness for the single-monitor IPC surface for now.

The backend is structured to remain compatible with future multi-monitor support, especially since MangoWC exposes output-scoped information through `mmsg -O` and tag masks per-output via `mmsg -g -t`. But without real hardware I can’t reliably implement or validate per-output tracking yet.

Given that, I’d prefer if this PR isn’t blocked on full multi-monitor support. Once I can test on an actual multi-monitor MangoWC setup, I’ll follow up with a PR that adds proper per-output handling (active/occupied/urgent masks, focused output state, layouts, etc.).

If any collaborators or maintainers do have a multi-monitor MangoWC environment, I’d really appreciate help capturing the IPC outputs.

Otherwise it would be nice to start a review :)

clotodex:
I agree with the General Sentiment of not blocking this PR unnecessarily for every single detail. But I would love to have a best effort implementation that then can be improved. I can through Nix definitely check if I can run Mango in a very simple setup and test that it generally works. But I guess that given that you've checked the IPC output so far you should be able to kind of arrive at a best effort implementation that then I can just test and then that should be fine. But every single edge case we probably won't capture, that is then to me a better next PR.

What do you think is needed for minimal multi monitor?
We just need to group that information for workspaces, everything else i dont think is affected

Sachin-Bhat:
Hey @clotodex sorry was a little occupied the past few days.

Thanks, this makes sense and I agree with the scope.

I’ll keep this PR as a best-effort implementation and avoid blocking on every edge case.  
For minimal multi-monitor correctness, I’ll focus only on workspace/tag identity and grouping, since that is the part affected by Mango’s monitor-scoped tags.

Concretely, next step in this PR:

- [ ] treat workspace identity as `(monitor_name, tag_id)` (not just `tag_id`)
- [ ] ensure active-state resolution is monitor-aware for workspace rendering

I won’t expand scope beyond that in this PR (window/title/layout/etc), and we can follow up with additional edge-case hardening in a separate PR after your nix validation.

andyrtr:
I have a tesing MangoWC setup with dual monitor. I've done two separate .conf files each with it's own "outputs". I have tag 1-5 at one monitor and 6-9 on the other. 

Both ashells show all tags and not only tags that belong to the monitor no matter what setting I choose for visibilty_mode.  Both bars also show only the active tags from one monitor (currently 6+9 form the right one). Mouse clicks are only taken on the right panel, the left one does not respond to any input.

MalpenZibo:
Thanks for this new feature. I will not merge it into the next release at this time. After the release of the new version, we can proceed with the inclusion of the mangoWC support

src/services/compositor/types.rs
```diff
@@ -74,6 +86,7 @@ pub struct CompositorState {
     pub workspaces: Vec<CompositorWorkspace>,
     pub monitors: Vec<CompositorMonitor>,
     pub active_workspace_id: Option<i32>,
+    pub active_workspace_ids: Vec<i32>,
```
MalpenZibo, line 89: Maybe we should maintain a single field here: `active_workspace_ids` should be the only field.
- hyprland/niri: will check only if it is empty or the first value in the `Vec`.
- mangoWC: should do what it does

I prefer this solution instead of this kind of thing `active_workspace_ids: active_workspace_id.into_iter().collect(),`

If we really want, we can create an `ActiveWorkspaces` wrapper struct with methods to simplify getting information, but maybe it's an overkill

src/services/compositor/mangowc.rs
```diff
@@ -0,0 +1,357 @@
...
+    urgent_mask: u32,
+}
+
+pub fn is_available() -> bool {
+    StdCommand::new("mmsg")
+        .args(["-g", "-O"])
+        .output()
+        .map(|out| out.status.success())
+        .unwrap_or(false)
+}
```
MalpenZibo, line 35: This seems fragile and could hang. There's no other solution?

src/services/compositor/mangowc.rs
```diff
@@ -0,0 +1,357 @@
...
+            .then(a.id.cmp(&b.id))
+    });
+
+    let active_window = parse_active_window(&active_client_raw);
+    let keyboard_layout = parse_keyboard_layout(&keyboard_raw);
+    let submap = parse_keymode(&keymode_raw);
+
+    let mut active_workspace_ids = active_workspace_ids.into_iter().collect::<Vec<_>>();
+    active_workspace_ids.sort_unstable();
+    active_workspace_ids.dedup();
```
MalpenZibo, line 179: Why dedup? I mean `active_workspace_ids` is a hashset, so no duplicated values. We create a vec from that hashset, so no duplicated values, then sort, and then dedup. Probably it is not needed.

src/services/compositor/mangowc.rs
```diff
@@ -0,0 +1,357 @@
...
+    )));
+    Ok(())
+}
+
+async fn fetch_full_state() -> Result<CompositorState> {
+    let tags_raw = run_mmsg(["-g", "-t"]).await?;
+    let outputs_raw = run_mmsg(["-g", "-O"]).await?;
+    let active_client_raw = run_mmsg(["-g", "-c"]).await?;
+    let keyboard_raw = run_mmsg(["-g", "-k"]).await?;
+    let keymode_raw = run_mmsg(["-g", "-b"]).await?;
```
MalpenZibo, line 125: here we could use a `tokio::try_join!` to run them concurrently

MalpenZibo:
I saved the work of this PR here https://github.com/MalpenZibo/ashell/pull/846 after the refactor of the Wayland compositor backends

Issue #478, opened 2026-02-09
[Bug] Crash trigger by playing Arma 3


Simber1:
Hi, whenever I play Arma 3 (cachy-proton, no gamescope) my ashell ends up crashing. I'm unsure as to what causes it, running ashell in terminal I noticed these logs before the crash (which occurred when closing the game in this case). The first line of the log is spammed repeatedly.  

```
ERROR [iced_winit::platform_specific::wayland::event_loop] SCTK dispatch error: underlying IO error: Argument list too long (os error 7)
Error trying to flush the wayland display: Argument list too long (os error 7)
ERROR [ashell] Panic: panicked at /home/peter/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tokio-1.48.0/src/runtime/blocking/shutdown.rs:51:21:
Cannot drop a runtime in a context where blocking is not allowed. This happens when a runtime is dropped from within an asynchronous context.
 disabled backtrace
```  
If you need any further logs or info lmk.

romanstingler:
hey, 
from your logs I can just assume the issue 

can you try building my branch and tell me if that resolves your issue


```
git clone https://github.com/romanstingler/ashell.git
cd ashell
git checkout fix/tokio-runtime-crash

cargo b --release
./target/release/ashell

**Comment by Simber1** [comment]
Yep, that branch fixes it. Thanks!

**Comment by Simber1** [comment]
I spoke too soon, relaunched the game and it broke again. Basically the same error.

```
ERROR [iced_winit::platform_specific::wayland::event_loop] SCTK dispatch error: underlying IO error: Argument list too long (os error 7)
Error trying to flush the wayland display: Argument list too long (os error 7)
ERROR [ashell] Panic: panicked at /home/peter/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tokio-1.49.0/src/runtime/blocking/shutdown.rs:51:21:
Cannot drop a runtime in a context where blocking is not allowed. This happens when a runtime is dropped from within an asynchronous context.
 disabled backtrace

ERROR [iced_winit::platform_specific::wayland::event_loop] SCTK dispatch error: underlying IO error: Argument list too long (os error 7)
Error trying to flush the wayland display: Argument list too long (os error 7)
```

romanstingler:
damn, 
can you try running that from terminal 

```RUST_BACKTRACE=full ashell```

Simber1:
[...]

PR #506, opened 2026-02-23
fix: enforce 2048 character hard limit for window titles

romanstingler:
This PR addresses the crash reported in #478 by truncating window titles to 2048 bytes before they are processed.

The Wayland protocol enforces a 4096-byte limit on single socket message. Certain applications, particularly games like Arma 3 running through Proton, frequently update their window titles with strings containing high-density information like FPS counters, server details, or mod lists. When these strings exceed the protocol's byte limit, the underlying IO returns Argument list too long (`os error 7`), causing the compositor or the application's event loop to crash.

Following the approach taken by other projects like Kitty and Alacritty, this change introduces a truncation limit of 2048 bytes. This value is intentionally conservative; it provides a safe buffer for protocol overhead while ensuring that even the most verbose game titles do not exceed the 4096-byte hardware limit.

https://github.com/kovidgoyal/kitty/commit/2e3037ce3a3f937067d9d50417aafae19a93dd88

### DIFF
--- a/src/modules/window_title.rs
+++ b/src/modules/window_title.rs
@@ -74,11 +74,14 @@ impl WindowTitle {
                     },
                 };
 
-                if self.config.truncate_title_after_length > 0 {
-                    truncate_text(raw_title, self.config.truncate_title_after_length)
+                // Apply hard limit of 2048 characters to prevent Wayland E2BIG errors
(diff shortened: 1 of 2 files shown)


MalpenZibo (approved):
🚀

PR #518, opened 2026-03-05
Better logging when search for target output fails.

SGumbles:
Ashell will run fine on first boot even if the outputs in the config file do not exist. Coming back from DPMS off, ashell will sometimes randomly disappear when wayland destroys/creates the output. This change lets a user catch this error in the log files as it silently fails before.


MalpenZibo:
Can you fix the formatting?

SGumbles:
> Can you fix the formatting?

Yes, sorry

PR #519, opened 2026-03-06
Tempo weather additional config

1randomguy:
I added a new config option `weather_indicator` to the tempo module to change what is shown in the bar when the `weather_location` is set.
The available options are  `Empty`, `Icon`, or `IconAndTemparature` to show either nothing next to the clock, only the icon, or everything, which is also the default.
This allows you to see the weather in the menu, without necessarily cluttering the bar.

~I didn't yet add it to the documentation, still have to do that.~

I also added hot-reloading capabilities of these config options to the tempo module.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -279,7 +279,7 @@ pub enum WeatherIndicator {
     None,
 }
 
-#[derive(Deserialize, Default, Clone, Debug)]
+#[derive(Deserialize, Default, Clone, Debug, PartialEq)]
 pub enum WeatherLocation {
     #[default]
     Current,
--- a/src/modules/tempo.rs
+++ b/src/modules/tempo.rs
@@ -139,6 +139,13 @@ impl Tempo {
                     self.current_timezone_index = 0;
                 }
 
+                let location_changed = self.config.weather_location != new_config.weather_location;
+
+                if location_changed {
+                    self.weather_data = None;
+                    self.location = None;
+                }
+
                 self.config = new_config;
                 Action::None
             }


clotodex:
@romanstingler would love your take on the config if it fits your redesign of the settings module

src/modules/tempo.rs
```diff
@@ -139,7 +139,17 @@ impl Tempo {
                     self.current_timezone_index = 0;
                 }
 
+                let weather_config_changed = self.config.weather_location
+                    != new_config.weather_location
+                    || self.config.weather_indicator != new_config.weather_indicator;
```
romanstingler: ```let location_changed = self.config.weather_location != new_config.weather_location;```

we can reuse the cached weather data as long as we don't change the location. 
The icon is just for display and shouldn't matter at that point.

src/config.rs
```diff
@@ -268,9 +268,17 @@ pub struct TempoModuleConfig {
...
     pub weather_location: Option<WeatherLocation>,
+    pub weather_indicator: Option<WeatherIndicator>,
 }
 
-#[derive(Deserialize, Default, Clone, Debug)]
+#[derive(Deserialize, Clone, Debug, PartialEq, Eq)]
+pub enum WeatherIndicator {
+    Empty,
+    Icon,
+    IconAndTemperature,
```
romanstingler, line 277: #[default]

I would add that

src/config.rs
```diff
@@ -284,6 +292,7 @@ impl Default for TempoModuleConfig {
             formats: vec![],
             timezones: vec![],
             weather_location: None,
+            weather_indicator: None,
```
romanstingler: `            weather_indicator: Some(WeatherIndicator::IconAndTemperature),`

romanstingler:
From the idea, I think it is valid.

I think in the future we should re-use less and define all display options for all of these modules.
There are some like Network, where I am not fully sure yet.


Idk if I like `Empty`. I guess you didn't went with `None` because it is a keyword.  
We could still use `None` but how about 
`weather_indicator = "No"` 
to be honest I don't know.


from the code hmmm I added a few comments nothing major in my pov.

MalpenZibo:
Can 'None' not be used as a value? It would be great if that were possible. Otherwise, let's use `No`

1randomguy:
> Can 'None' not be used as a value? It would be great if that were possible. Otherwise, let's use `No`

It's possible, just tried it. Just felt a bit wrong to me as it is a keyword, but since we will have to write `WeatherIndicator::None` I guess its fine :+1:

romanstingler:
yes, I was not sure for the same reason, I think we can go with `None` for now

src/config.rs
1randomguy: I guess the weatherindicator doesn't really need to be wrapped in an option in the first place :+1:

website/docs/configuration/modules/tempo.md
```diff
@@ -23,13 +23,15 @@ Tempo combines a highly configurable clock with an optional weather summary in t
 | `formats`          | `array`  | `[]`          | Multiple datetime formats that can be cycled through by right-clicking the clock. When provided, clicking cycles through each format in sequence.                                                     |
 | `timezones`        | `array`  | `[]`          | Timezone identifiers that can be cycled through by scrolling. Supports both IANA names (e.g., `"UTC"`, `"America/New_York"`) and fixed offsets (e.g., `"+00:00"`, `"-05:00"`).                        |
 | `weather_location` | `enum`   | `None`        | Determines which coordinates are queried when requesting weather data. `Current` geo-locates via IP using `ip-api.com`. Use the `City` variant to pin the module to a specific place.                 |
+| `weather_indicator`| `enum`   | `IconAndTemperature`| Determines what information about the weather is shown in the bar, valid options are `Empty`, `Icon`, and `IconAndTemperature`.           |
```
MalpenZibo:
```suggestion
| `weather_indicator`| `enum`   | `IconAndTemperature`| Determines what information about the weather is shown in the bar, valid options are `None`, `Icon`, and `IconAndTemperature`.           |
```

src/modules/tempo.rs
```diff
@@ -644,18 +666,21 @@ impl Tempo {
...
-                    "weather",
-                ),
-                channel(100, async move |mut output| {
+        let tempo_config_for_sub = self.config.clone();
+
+        let weather_sub = Subscription::run_with_id(
+            (
+                TypeId::of::<Self>(),
+                format!("{:?}", tempo_config_for_sub.weather_location),
+                format!("{:?}", tempo_config_for_sub.weather_indicator),
```
MalpenZibo: Do we really need the weather_indicator in the id of the subscription? I mean, changing the indicator doesn't change the behavior of the subscription

src/modules/tempo.rs
```diff
@@ -644,18 +666,21 @@ impl Tempo {
             }),
         );
 
-        let weather_sub = self.config.weather_location.clone().map(|location| {
-            Subscription::run_with_id(
-                (
-                    TypeId::of::<Self>(),
```
MalpenZibo, line 650: Why remove the mapping to create the subscription?

src/modules/tempo.rs
```diff
@@ -690,15 +715,14 @@ impl Tempo {
...
                         tokio::time::sleep(Duration::from_secs(60 * failed_attempt)).await;
                     }
-                }),
-            )
-        });
+                } else {
+                    // If weather_location is None, just sleep indefinitely
+                    let () = std::future::pending().await;
+                }
+            }),
```
MalpenZibo: I prefer the Optional weather subscription instead of creating a subscription with a pending

1randomguy: I'm curious, why do you think so? Is it considered the cleaner way in rust, or do you feel like it's more readable to you?
For me as someone who hasn't worked that much with rust, calling map on an option felt a bit unnatural at first, but I guess I'm starting to see the appeal.

src/modules/tempo.rs
1randomguy: reverted 🫡

src/modules/tempo.rs
MalpenZibo: In this case it's because we can completely avoid creating an iced subscription. If the subscription is none we return only the time subscription otherwise a batch with the two subscriptions

MalpenZibo (approved):
🚀

Issue #487, opened 2026-02-12
Fix laggy/jittery audio sliders


levitatingpineapple:
This is dependent on issue:  https://github.com/MalpenZibo/ashell/issues/414 and it's corresponding PR: https://github.com/MalpenZibo/ashell/pull/449

Currently the audio sliders suffer from few responsiveness problems. These issues are more pronounced when using Bluetooth devices.

### Jitter

Sometimes slider "jumps around" when adjusting volume. This is caused by it being updated both locally and via pulse-audio subscription.   Should be addressed by merging https://github.com/MalpenZibo/ashell/pull/449

### Compounding Lag

During a prolonged drag _set volume_ messages are generated faster than they can be processed. This leads to messages getting queued up somewhere. In the logs I can see that I keep getting new sinks/sources from pulse-audio up to few seconds after a long drag gesture has already ended.

This could be addressed by throttling in the following way:

1. First call is instantaneous
2. Subsequent calls are either throttled by time (100ms for example) or by waiting for feedback from sinks/sources subscription.
3. Final slider position is always delivered to pulse-audio

### General Responsiveness

For some headsets it takes like 300ms to update the volume - this makes ashell feel sluggish. There was a PR https://github.com/MalpenZibo/ashell/pull/374 from @romanstingler that addressed a similar issue but for the brightness slider.
Perhaps we could create a reusable solution for this as it has some complexity.

3bnz:
> Posting this here for future reference.

Did some research in isolation on what would be an _idiomatic_way in iced to have reusable and responsive sliders.

The general concept is to keep track of both - `requested` local value as well as (potentially delayed) `received` source of truth. When a drag starts - the slider switches to showing `requested` value to ensure instantaneous feedback. Then - after the drag finishes - we wait for some amount of time before showing the `received` value again. 

- The timeout should be longer than the maximum time (we expect) it takes to receive the remote value.
- The task should be cancelled, in case a new drag starts, while still showing `requested` value

This is how it could be abstracted without wrapping a widget or using the deprecated [`Component`](https://docs.rs/iced/latest/iced/widget/trait.Component.html):

### async_slider.rs

```rust
use iced::{Element, Task, task::Handle, widget::slider};
use num_traits::FromPrimitive;
use std::{ops::RangeInclusive, time::Duration};
use tokio::time::sleep;

#[derive(Clone)]
pub enum Message<V> {
    Request(V),
    GestureEnded,
    ShowReceived,
}

[...]

PR #522, opened 2026-03-09
Responsive Sliders

levitatingpineapple:
Hey @romanstingler,

I've done the integration of responsive sliders as proposed in #487 over the weekend.
Quite happy with how it turned out!

### Responsive Sliders

Adds instant UI updates, even when pulse audio is slow to respond.
I've added `Remote<Value>` utility.
Here's a logic diagram of how I expect it to work:

```mermaid
flowchart LR
    w[Slider or scroll overlay]
    pa[Pulse Audio]
    s[State]
    m{"Timed Select (mux). between instant local update and source of truth"}
    s --dictates--> w
    w --instant local update--> m --> s
    w --request--> pa --delayed source of truth--> m
```

- We have a unified logic for handling both sliders as well as scroll overlays (status icon too)
- Sliders emit `Request` messages during drag and `Timeout` message on release.
- Scroll overlays emit `RequestAndTimeout` messages since they don't have end event

### Pulse audio message throttling.

Implemented by adding a throttled channel and reusing `crate::services::throttle::ThrottleExt`, which is also used in `upower` module. The stream throttle extension has been updated to actually **drop stale values**. Since there are many flavors of stream throttling - this is the one I'm going for:

throttle

### Fix volume rounding error.

Adding remote value revealed a bug where volume from pulse-audio server is always one percent less than when set locally. I suspect it's the reason for this comment: 
  - Fixed by removing `u32(Volume::MUTE..=Volume::NORMAL) => f64(0.0..=1.0) => i32(0..=100) ` chain and only scaling the value to display as percentage.

### Additional Fixes

- Right click on mic button now correctly triggers `OpenSourceMore` message
- Channels are now scaled from `max()` instead of `avg()` channel volumes, as this is what `pub fn scale(&mut self, max: Volume)` expects

### DIFF
--- a/src/modules/settings/audio.rs
+++ b/src/modules/settings/audio.rs
@@ -12,7 +12,9 @@ use crate::{
 use iced::{
     Alignment, Element, Length, Subscription, Task, Theme,
     mouse::ScrollDelta,
-    widget::{Column, MouseArea, Row, Text, button, column, container, horizontal_rule, row, text},
+    widget::{
+        Column, MouseArea, Row, Text, button, column, container, horizontal_rule, row, slider, text,
+    },
     window::Id,
 };
 use libpulse_binding::volume::Volume;
@@ -25,9 +27,9 @@ pub enum Message {
     DefaultSinkChanged(String, Option<String>),
     DefaultSourceChanged(String, Option<String>),
     ToggleSinkMute,
-    SinkSlider(remote_value::Message<u32>),
+    SinkVolumeChanged(remote_value::Message<u32>),
     ToggleSourceMute,
-    SourceSlider(remote_value::Message<u32>),
+    SourceVolumeChanged(remote_value::Message<u32>),
     SinksMore(Id),
     SourcesMore(Id),
     OpenMore,
@@ -126,13 +128,16 @@ impl AudioSettings {
                 }
                 Action::None
             }
-            Message::SinkSlider(message) => {
+            Message::SinkVolumeChanged(message) => {
                 if let Some(service) = self.service.as_mut() {
                     if let Some(value) = message.value() {
                         let _ = service.command(AudioCommand::SinkVolume(value));
                     }
                     return Action::Task(
-                        service.sink_slider.update(message).map(Message::SinkSlider),
+                        service
+                            .sink_slider
+                            .update(message)
+                            .map(Message::SinkVolumeChanged),
                     );
                 }
                 Action::None
@@ -149,7 +154,7 @@ impl AudioSettings {
                 }
                 Action::None
             }
-            Message::SourceSlider(message) => {
+            Message::SourceVolumeChanged(message) => {
                 if let Some(service) = self.service.as_mut() {
                     if let Some(value) = message.value() {
                         let _ = service.command(AudioCommand::SourceVolume(value));
@@ -158,7 +163,7 @@ impl AudioSettings {
                         service
                             .source_slider
                             .update(message)
-                            .map(Message::SourceSlider),
+                            .map(Message::SourceVolumeChanged),
                     );
                 }
                 Action::None
@@ -232,13 +237,13 @@ impl AudioSettings {
                         let icon = icon(icon_type);
                         MouseArea::new(icon)
                             .on_right_press(Message::OpenMore)
-                            .on_scroll(Self::on_scroll(volume, Message::SinkSlider))
+                            .on_scroll(Self::on_scroll(volume, Message::SinkVolumeChanged))
                             .into()
                     }
                     SettingsFormat::Percentage | SettingsFormat::Time => {
                         MouseArea::new(Self::vol_text(volume))
                             .on_right_press(Message::OpenMore)
-                            .on_scroll(Self::on_scroll(volume, Message::SinkSlider))
+                            .on_scroll(Self::on_scroll(volume, Message::SinkVolumeChanged))
                             .into()
                     }
                     SettingsFormat::IconAndPercentage | SettingsFormat::IconAndTime => {
@@ -249,7 +254,7 @@ impl AudioSettings {
                                 .align_y(Alignment::Center),
                         )
                         .on_right_press(Message::OpenMore)
-                        .on_scroll(Self::on_scroll(volume, Message::SinkSlider))
+                        .on_scroll(Self::on_scroll(volume, Message::SinkVolumeChanged))
                         .into()
                     }
                 }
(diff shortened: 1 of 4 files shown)


romanstingler:
I will have a look as soon as I am less burned out

3bnz:
> I will have a look as soon as I am less burned out

No worries :)
I'm not blocked by anything in particular here - I'll continue to chip away at remaining tasks during this week.
Just wanted to show what progress is being made and give folks opportunity to give some early feedback if they feel like it.

clotodex:
Tested and working :tada:

src/utils/remote_value.rs
clotodex, line 1: i feel like this should go into the utils/ folder

src/utils/remote_value.rs
```diff
@@ -0,0 +1,96 @@
+use iced::{Element, Task, task::Handle, widget::slider};
+use num_traits::FromPrimitive;
```
clotodex: is this still needed (could not find from / into while skimming over)?
Just checking to save on extra dependencies

src/services/audio.rs
```diff
@@ -601,45 +602,56 @@ impl PulseAudioServer {
...
-                                }
-                                Some(PulseAudioCommand::SourceVolume(name, volume)) => {
-                                    let _ = server.set_source_volume(&name, &volume);
-                                }
-                                Some(PulseAudioCommand::DefaultSink(name, port)) => {
-                                    let _ = server.set_default_sink(&name, port.as_deref());
-                                }
-                                Some(PulseAudioCommand::DefaultSource(name, port)) => {
-                                    let _ = server.set_default_source(&name, port.as_deref());
+            tokio::runtime::Builder::new_current_thread()
```
clotodex, line 602: this is getting quite complex. And not sure if it needs to.
For this pr this is fine though.
But especially with throttling having an mspc with recv_timeout and an option that tracks the "pending action" overriding on new events but only training on the pending items might make this a whole lot simpler (and potentially sync not async)

But since your implementation is close to the previous i would make the "cleanup" in a separate pr

src/services/audio.rs
```diff
@@ -362,19 +358,19 @@ impl Service for AudioService {
             }
             AudioCommand::SinkVolume(volume) => {
                 if let Some(sink) = self.active_sink()
-                    && let Some(volume) = sink.volume.scaled(volume as f64 / 100.)
+                    && let Some(volume) = sink.volume.scaled(volume)
                 {
                     let _ = self
-                        .commander
+                        .commander_throttled
```
clotodex: Why do we still need the normal commander? I feel like the throttled commander handles everything the normal one is doing (implemented below (muting, etc) and otherwise unused)
So gutfeeling would be to move the entire service to throttled. Since we are not using "toggles" but explicit mute states etc this should be fine.

src/utils/remote_value.rs
3bnz: This is required by iced to initialize [slider](https://docs.iced.rs/iced/widget/struct.Slider.html#impl-Slider%3C'a,+T,+Message,+Theme%3E)'s default step value.
We already have implicit dependency on `num_traits`.

Technically I could inline the `remote_value::slider` function using turbofish `::<>` syntax, but then I'd have to make sure it's used correctly at every call-site

src/services/audio.rs
3bnz: Interesting. 

I tried to build some naive throttle in the [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=047d8b61ea0eefb81f1144358b30bc95) using the `recv_timeout` primitive. (Tapping "Send" will emit events)

However the `recv_timeout` **does not block** after the timeout has expired and instantly always returns `RecvTimeoutError` - so we end up essentially busy waiting.

What would be the pattern here - do we need to conditionally to also call `.recv` to block the loop after timeout expires?

src/utils/remote_value.rs
clotodex: All good :)

src/services/audio.rs
3bnz: I think I agree with you here.
In practice user will almost never emit two events of different kind within 100ms.
And even if something gets dropped - the UI won't be out of sync.

src/services/audio.rs
clotodex: your playground version does not run for me (gets auto-killed)
But does not matter.
The idea behind this is less the recv_timeout waiting, but rather "busy looping" while evens are being emitted until "duration" time events were not emitted again - after that it might turn into a busy loop again though which is fair and i did not think of.

I would leave it as you have it and maybe start an optimization pr on it. Wdyt?

src/modules/settings/audio.rs
```diff
@@ -7,24 +7,27 @@ use crate::{
...
+const VOL_PERCENT: u32 = Volume::NORMAL.0 / 100;
 
 #[derive(Debug, Clone)]
 pub enum Message {
     Event(ServiceEvent<AudioService>),
     DefaultSinkChanged(String, Option<String>),
     DefaultSourceChanged(String, Option<String>),
     ToggleSinkMute,
-    SinkVolumeChanged(i32),
+    SinkSlider(remote_value::Message<u32>),
```
clotodex: Should we rename this to SinkSliderChanged?

src/modules/settings/audio.rs
```diff
@@ -7,24 +7,27 @@ use crate::{
...
     ToggleSourceMute,
-    SourceVolumeChanged(i32),
+    SourceSlider(remote_value::Message<u32>),
```
clotodex: same here, SourceSliderChanged?

```suggestion
    SourceSliderChanged(remote_value::Message<u32>),
```

src/modules/settings/audio.rs
```diff
@@ -483,27 +468,18 @@ impl AudioSettings {
...
+                    SliderType::Source => Message::OpenSourceMore,
+                }),
             )
             .push(
                 MouseArea::new(
-                    slider(0..=100, volume, volume_changed)
-                        .step(1)
-                        .width(Length::Fill),
+                    volume
+                        .slider(Volume::MUTED.0..=Volume::NORMAL.0)
```
clotodex: potentially pedantic opinion:
but I feel like this helper function might make sense to stay in this file or inline. It "pollutes" remote_value.rs.
The RemoteValue is usable technically also for other areas of the project and thus makes sense to be in utils, but the slider is only used here and highly specifc to the UI

Happy to be convinced otherwise, wdyt?

src/utils/remote_value.rs
```diff
@@ -0,0 +1,96 @@
...
+        task
+    }
+}
+
+impl<Value> Remote<Value>
+where
+    Value: Copy + Send + From<u8> + PartialOrd + FromPrimitive + 'static,
+    f64: From<Value>,
+{
+    pub fn slider(&self, range: RangeInclusive<Value>) -> Element<'_, Message<Value>> {
```
clotodex: would move this to the module instead, see my other comment

src/modules/settings/audio.rs
3bnz: Technically the `Remote<Value>` is not really independent due to it using `iced::Task` and `iced::task::Handle` - which must always be returned to the iced runtime.
Perhaps I'm being too liberal with `use` statements for types that have very generic names.

However - inline slider will also remove explicit `num-traits` dependency - which you also mentioned.

So yes - I'm happy to inline the function.

src/modules/settings/audio.rs
3bnz: Let's keep the original  `SinkVolumeChanged` name then.
This is also emitted by the scroll overlay on the status icon - which is not a slider.

clotodex (approved):

romanstingler:
sorry, just found some time to have a look into that one. 
For me it looks good enough to be merged. 

I created an issue for the brightness slider to use the Responsive Sliders as well. 

for me the code works well. 
Thanks for the contribution.

Issue #531, opened 2026-03-14
Use Responsive Sliders for Brightness slider


romanstingler:
Based on https://github.com/MalpenZibo/ashell/pull/522
re-use Responsive Sliders for the Brightness slider as well.

3bnz:
Have you started with this?
If not - I'll take a look this evening.

romanstingler:
no I have not. 
I am busy creating new docker images for work...

PR #539, opened 2026-03-18
Brightness Slider

levitatingpineapple:
This PR contains fixes for brightness slider.

> **Quick context:** while the brightness value can be simply read from the `/sys/class...` filesystem, writing it requires a privileged process - so we must send a dbus message to `org.freedesktop.login1` so `systemd-logind` can handle it. This is why reading and getting and setting uses two different approaches.

- Adopt `Remote<Value>` #531
- Add throttled commander so we don't spam DBUS on every frame.
- Remove retained `ui_percentage` value in favor of computing the percent string only before displaying the `Text` label.
- Remove _refresh values on menu open_ - this must be redundant because we show the value in the status icon too => hence brightness value has to be always up-to-date anyway.
- Use `/sys/class/../brightness` instead of `/sys/class/../actual_brighteness` This is what both [`brightnessctl`](https://github.com/Hummer12007/brightnessctl/blob/e70bc55cf053caa285695ac77507e009b5508ee3/brightnessctl.c#L257) and [`cosmic-settings-daemon`](https://github.com/pop-os/cosmic-settings-daemon/blob/ee782f454a09310a28abe73653e6c82d06a79855/src/brightness_device.rs#L28) are using. On my system there is persistent difference between those values.

### DIFF
--- a/src/services/brightness.rs
+++ b/src/services/brightness.rs
@@ -1,4 +1,5 @@
 use super::{ReadOnlyService, Service, ServiceEvent};
+use crate::{services::throttle::ThrottleExt, utils::remote_value::Remote};
 use iced::{
     Subscription, Task,
     futures::{SinkExt, StreamExt, channel::mpsc::Sender, stream::pending},
@@ -8,23 +9,27 @@ use log::{debug, error, info, warn};
 use std::{
     any::TypeId,
     fs,
-    ops::Deref,
+    ops::{Deref, DerefMut},
     path::{Path, PathBuf},
+    time::Duration,
 };
-use tokio::io::{Interest, unix::AsyncFd};
+use tokio::{
+    io::{Interest, unix::AsyncFd},
+    sync::mpsc::{UnboundedReceiver, UnboundedSender},
+};
+use tokio_stream::wrappers::UnboundedReceiverStream;
 use zbus::proxy;
 
 #[derive(Debug, Clone, Default)]
 pub struct BrightnessData {
-    pub current: u32,
+    pub current: Remote<u32>,
     pub max: u32,
 }
 
 #[derive(Debug, Clone)]
 pub struct BrightnessService {
     data: BrightnessData,
-    device_path: PathBuf,
-    conn: zbus::Connection,
+    commander: UnboundedSender<BrightnessCommand>,
 }
 
 impl Deref for BrightnessService {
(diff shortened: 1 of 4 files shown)


clotodex:
Feel free to request me for review when ready

3bnz:
> Feel free to request me for review when ready

I don't think I have the power to request a review. :)
But yes - it's ready to look at.
Most of changes mimic the audio service.

One interesting catch is that we know, maximum `Volume::NORMAL` at compile-time, but `brightness_max` is retrieved at runtime - so we have to gracefully handle unexpected values - like 0 or it being less than 100 - for calculating scroll steps.

src/utils/remote_value.rs
```diff
@@ -13,6 +13,15 @@ pub struct Remote<Value> {
     timeout: Option<iced::task::Handle>,
 }
 
+impl<Value: Default> Remote<Value> {
+    pub fn new(value: Value) -> Self {
```
clotodex, line 17: do we need this? also have nothing against it though

src/services/brightness.rs
```diff
@@ -43,21 +54,17 @@ impl BrightnessService {
         Ok(max_brightness)
     }
 
-    async fn get_actual_brightness(device_path: &Path) -> anyhow::Result<u32> {
-        let actual_brightness = fs::read_to_string(device_path.join("actual_brightness"))?;
-        let actual_brightness = actual_brightness.trim().parse::<u32>()?;
-
-        Ok(actual_brightness)
+    async fn get_brightness(device_path: &Path) -> anyhow::Result<u32> {
+        let brightness = fs::read_to_string(device_path.join("brightness"))?;
```
clotodex, line 58: why did we switch from actual_brightness to brightness?

3bnz: Check the last point in the PR comment. :)

src/utils/remote_value.rs
3bnz: `fn new()` was added to initialize the `Remote` wrapper with non-default starting value.
The other option would be to  `Remote::<Value>::default()` and then call `fn receive(value)` on that instance.

romanstingler:
Works for me and looks ok

src/services/brightness.rs
clotodex: haha thank you!

PR #541, opened 2026-03-22
Add different display units for systeminfo

sulabi:
This pull request introduces different display units for system information values. The current ones I have added are:

- Memory: "Percentage" (default), "Fraction"
- Cpu: "Percentage" (default), "Frequency"
- Temperature: "Celsius" (default), "Fahrenheit"
- MemorySwap: Uses the same display value as memory

these can be changed in the config, and the default display values are how it was before.

example config:

```toml
[system_info.memory]
display_mode = "Percentage" # or "Fraction"

[system_info.cpu]
display_mode = "Percentage" # or "Frequency"

[system_info.temperature]
display_mode = "Fahrenheit" # or "Celsius"
```

### DIFF
--- a/src/modules/system_info.rs
+++ b/src/modules/system_info.rs
@@ -411,7 +411,7 @@ impl SystemInfo {
                                         format!("Disk Usage {mount_point}"),
                                         match self.config.disk.format {
                                             DiskFormat::Percentage => format!("{}%", usage.percentage),
-                                            DiskFormat::Fraction => format!("{} GiB", usage.fraction)
+                                            DiskFormat::Fraction => format!("{} GB", usage.fraction)
                                         }
                                     )
                                 })
@@ -484,7 +484,7 @@ impl SystemInfo {
                 StaticIcon::Mem,
                 match self.config.memory.format {
                     MemoryFormat::Percentage => ( self.data.memory_usage.percentage.to_string(), "%" ),
-                    MemoryFormat::Fraction => ( self.data.memory_usage.fraction.clone(), "" ),
+                    MemoryFormat::Fraction => ( self.data.memory_usage.fraction.clone(), " GiB" ),
                 },
                 match self.config.memory.format {
                     MemoryFormat::Percentage => Some((
@@ -503,7 +503,7 @@ impl SystemInfo {
                 StaticIcon::Mem,
                 match self.config.memory.format {
                     MemoryFormat::Percentage => ( self.data.memory_swap_usage.percentage.to_string(), "%" ),
-                    MemoryFormat::Fraction => ( self.data.memory_swap_usage.fraction.clone(), "" ),
+                    MemoryFormat::Fraction => ( self.data.memory_swap_usage.fraction.clone(), " GiB" ),
                 },
                 match self.config.memory.format {
                     MemoryFormat::Percentage => Some((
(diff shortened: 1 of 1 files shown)


src/modules/system_info.rs
```diff
@@ -37,25 +51,49 @@ fn get_system_info(
     temperature_sensor: &str,
 ) -> SystemInfoData {
     system.refresh_memory();
-    system.refresh_cpu_specifics(sysinfo::CpuRefreshKind::everything());
+    system.refresh_cpu_all();
```
MalpenZibo, line 87: Why this change?

src/modules/system_info.rs
```diff
@@ -20,11 +20,25 @@ struct NetworkData {
...
+}
+struct CpuUsage {
+    percentage: u32,
+    frequency: f32
+}
+
+struct Temperature {
+    celsius: Option<i32>,
+    fahrenheit: i32
+}
```
MalpenZibo, line 62: Why do we store both values? In my mind, we should store the "absolute" value and format that value to the desired format in the view method.

src/modules/system_info.rs
```diff
@@ -37,25 +51,49 @@ fn get_system_info(
...
 
     components.refresh(true);
     disks.refresh(true);
     networks.refresh(true);
 
-    let cpu_usage = system.global_cpu_usage().floor() as u32;
-    let memory_usage = ((system.total_memory() - system.available_memory()) as f32
-        / system.total_memory() as f32
-        * 100.) as u32;
+    let cpu_freq_mhz = system.cpus()[0].frequency() as f32;
```
MalpenZibo: Here, we only show the frequency of the first CPU. It's quite different from the counterpart `global_cpu_usage`.

src/config.rs
```diff
@@ -133,17 +133,18 @@ pub struct KeyboardLayoutModuleConfig {
 #[derive(Deserialize, Clone, Debug)]
 #[serde(default)]
 pub struct SystemInfoCpu {
-    #[serde(default)]
     pub warn_threshold: u32,
-    #[serde(default)]
     pub alert_threshold: u32,
+
+    pub display_mode: CpuDisplayMode
```
MalpenZibo: Here maybe it's better to use `Format` as in the rest of the codebase. So `pub format: CpuFormat`

juvilius:
please also include storage/disk usage in total (human-readable), if possible. 😀

src/modules/system_info.rs
sulabi: the docs state that this function calls the same as before:

`pub fn refresh_cpu_all(&mut self)`

Calling this method is the same as calling
`system.refresh_cpu_specifics(CpuRefreshKind::everything())`.
```rust
use sysinfo::System;

let mut s = System::new_all();
s.refresh_cpu_all();
```

src/modules/system_info.rs
sulabi: there are two seperate methods that need this information, `menu_view` as well as `view` formatting these values again would be redundant

src/modules/system_info.rs
sulabi: well yes, i should do the average frequency instead

MalpenZibo:
Don't worry, I'll do it

PR #542, opened 2026-03-23
feat: Media Player Visualizer (cava integration)

ashlcx:
Feature addition of a [cava](https://github.com/karlstav/cava) visualizer to media player. Only shows media is playing through the active player. 
Includes configuration options for the following:

| Field                    | Type    | Default    | Description                                              |
| ------------------------ | ------- | ---------- | -------------------------------------------------------- |
| `show_visualizer`        | bool    | `false`    | Enable or disable the visualizer.                        |
| `visualizer_bar_count`   | integer | `8`        | Number of frequency bars to display.                     |
| `visualizer_framerate`   | integer | `60`       | Target framerate for CAVA output.                        |
| `visualizer_padding`     | integer | `3`        | Horizontal padding around the visualizer canvas (px).    |
| `visualizer_color`       | color   | `"Text"`   | Bar colour — see Color below.                |
| `visualizer_channels`    | string  | `"Stereo"` | `"Stereo"` or `"Mono"`.                                  |
| `visualizer_mono_option` | string  | `"Average"`| Mono mix mode: `"Average"`, `"Left"`, or `"Right"`.      |

Example Config:

```toml
[media_player]
show_visualizer = true
visualizer_bar_count = 32
visualizer_color =  { low = "#a6e3a1", mid = "#f9e2af", high = "#f38ba8" }
visualizer_channels = "Mono"
visualizer_mono_option = "Average"
```

image

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -409,7 +409,6 @@ pub struct SettingsCustomButton {
     pub tooltip: Option<String>,
 }
 
-
 #[derive(Deserialize, Copy, Clone, Default, PartialEq, Eq, Debug)]
 pub enum MediaPlayerFormat {
     Icon,
--- a/src/config.rs
+++ b/src/config.rs
@@ -409,6 +409,7 @@ pub struct SettingsCustomButton {
     pub tooltip: Option<String>,
 }
 
+
 #[derive(Deserialize, Copy, Clone, Default, PartialEq, Eq, Debug)]
 pub enum MediaPlayerFormat {
     Icon,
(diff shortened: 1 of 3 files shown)


src/config.rs
```diff
@@ -417,18 +417,101 @@ pub enum MediaPlayerFormat {
...
+    Text,
+    Primary,
+    Success,
+    Danger,
+    Hex(HexColor),
+    Gradient {
+        low: HexColor,
+        mid: Option<HexColor>,
+        high: HexColor,
+    },
```
MalpenZibo, line 477: What if we declare this like optional values? I mean, the user could simply specify `gradient` and use the base palette (Success -> warn (right now we use a weak danger for this, but after the iced update, there will be a dedicated value) -> danger) and only if necessary override the specific color.

MalpenZibo:
supersided by #872

PR #546, opened 2026-03-25
fix(brightness): sync slider with actual brightness on menu open

MalpenZibo:
## Summary

- Re-add brightness refresh when settings menu is opened (removed in da2fca3 during Remote<Value> refactor)
- Store `device_path` in `BrightnessData` and add `sync_brightness()` method to re-read from sysfs
- Fix stale `current_value` in udev monitoring loop — now updated after each change detection so returning to the initial value isn't silently dropped
- Skip sysfs read errors instead of sending 0 via `unwrap_or_default()`

### DIFF
--- a/src/services/brightness.rs
+++ b/src/services/brightness.rs
@@ -191,15 +191,14 @@ impl BrightnessService {
                                                     );
                                                     if let Ok(new_value) =
                                                         Self::get_brightness(&device_path).await
+                                                        && new_value != current_value
                                                     {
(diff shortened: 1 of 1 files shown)


clotodex:
Ah damn my bad, thank you for the fix

Unsure in which situations it desyncs (changing it with keybinds and opening the menu syncs properly for me even without this fix)

MalpenZibo:
I don't know the real reason but for now I'll stick with this workaround

PR #558, opened 2026-03-27
feat(audio): implement volume change audio notifications

### LINKED ISSUE #525: [Feature request] Audio slider beep
It would be nice to have an audible signal when we click on the audio slider.
I do this to confirm which audio it is, when the name doesn't make it clear.

Image

romanstingler:
Based on the discussion in #525 and the follow-up conversation.

After we agreed to move away from spawning external processes (`pw-cat`), I first looked into `rodio` but that would pull in `libasound2-dev` as an additional system dependency, which I wanted to avoid to keep the dependency footprint small.

Instead I switched to `libpulse-simple-binding` which reuses the already required `libpulse.so`, so no new system library is needed while still giving us "real" audio playback.

What changed from the original approach:

- The beep is now a short embedded PCM sample (`assets/bell.pcm`) played directly via PulseAudio's simple API.
(I tried to synthesize this one but it is too complex and sounds really off otherwise).
- Playback runs on a single long-lived background thread holding one Pulse connection, so repeated beeps don't each pay a fresh connection handshake.
- Rate-limiting is built in: a beep only plays if the volume changed by at least 4% **and** at least 150ms passed since the last one, preventing overlapping sounds (no more PID tracking hacks).
- Triggered on slider drags, scroll wheel, mute toggles, and IPC volume keys, for both sink and source.
- `audio_feedback` is configurable in `[settings]` (default: `true`), and hot-reloadable. 
@MalpenZibo your call on the default value

Docs updated accordingly.

closes #525


MalpenZibo:
I was thinking that now that we were able to reduce the number of dependencies we could evaluate to use rodio.

We should probably need to add some dependency in the package and nix config to be able to compile.

So we could proceed by step. If this solution is acceptable let's merge and see, then if it's used and adding rodio is not a pain in the ass we can proceed adding a "real" audio support

romanstingler:
agree, we can add this one and if there is more demand we can just rework it. 
rebased

MalpenZibo:
Does this still make sense?

romanstingler:
@MalpenZibo I think people still want this 
if you would like I can re-implement this using rodio.

MalpenZibo:
@romanstingler ok! let's try

romanstingler:
@MalpenZibo ready to test

src/modules/settings/audio.rs
```diff
@@ -262,13 +275,15 @@ impl AudioSettings {
...
                     let _ = service.command(AudioCommand::ToggleSourceMute);
+                    self.audio_feedback.play_mute_toggle();
                 }
                 Action::None
             }
             Message::SourceVolumeChanged(message) => {
                 if let Some(service) = self.service.as_mut() {
                     if let Some(value) = message.value() {
                         let _ = service.command(AudioCommand::SourceVolume(value));
+                        self.audio_feedback.play(value);
```
garro95, line 286: Feedback on source volume change seems unusual.

MalpenZibo (commented):
We still have some issues to address.

I think that with the current implementation, we keep the audio device out of suspend for the whole ashell session, and we don't have any recovery if PulseAudio restarts or is not available yet.

Anyway, could we pause this feature for now? I still need to decide how to handle this

PR #573, opened 2026-03-30
Update iced

MalpenZibo:


src/services/audio.rs
```diff
@@ -572,8 +585,13 @@ impl PulseAudioServer {
 
                     loop {
                         let data = server.mainloop.iterate(true);
+                        if shutdown.load(Ordering::Relaxed) {
```
romanstingler, line 588: `iterate()` is blocking, maybe we want to put the shutdown block before that

romanstingler:
By testing your stuff I found something in the MPRIS service that I want to fix now :) 

I will wait with the review until it is ready.

PS: your roadmap 0.9 is basically UI only, which is fine(we have not yet focused on that) , 
but this is nothing I am good at or enjoy. 
I will stick around for bugfixes, support, some code path hardening and smaller cleanups/refactors (in services and non-UI code paths).

romanstingler:
I think then we can close this one

MalpenZibo:
@romanstingler the roadmap is UI only but in this phase I'm still defining what to do. There're a lot of things that cold be improved also backend side let's say. So fell free to suggest new things.

For now I'm quite busy with this iced update to start with a new base iced version for the 0.9 release. In this way we should have a lot of time to fix bug or regression.

romanstingler:
Yes, I will be here for sure, for code reviews and discussions. 
Just I am not good at shifting pixels :) 

I want to continue to work on existing stuff (no new features,  I think for now we have a good amount of features) 
and make that better.

PR #574, opened 2026-03-30
Refactor palette config

MalpenZibo:


romanstingler:
I will write here, because the other one doesn't build and this one is on top of it. 

Just an observation the new one has a white border around it. 

**new**
image
**old** 
image

I will edit this comment as I will progress with the review. 
Just leaving this so I don't lose it. 


-----

No worries :) 

I didn't want to show this one 
image

MalpenZibo:
Yeah, I know it. 
There were some changes in the iced palette + an error that I made in the ashell theme. 

Now we have a lot more background variants, and in iced, the secondary color is generated from the background and text color. I want to rectify my mistake so that some level of breaking changes could be present. 

In this PR, I plan to mitigate the problem. Sorry, because I forgot to set this and the PR that updates iced as a draft

clotodex:
The plan is still to keep parity with normal iced theming practices, correct?

MalpenZibo:
@romanstingler check this refactor branch to test regression in the UI.

cc @clotodex. I want to simplify things and remove the `secondary` palette from the config entirely. I plan to add a dedicated warning palette (like in the new iced version).

Let me know if this could work for you.

clotodex:
It looks good when skimming over it
But i just looked at the iced src and we are getting really close to an actual copy
Would it make sense to just support the ACTUAL iced palette? Having a few PaletteExtensions (or similar) for very custom things?

romanstingler:
the borders are fine, the colors look ok 



image


the right side of the sliders are still transparent

image

MalpenZibo:
I need to check with your config because in my case, it works
image

MalpenZibo:
@romanstingler ok, in your case the problem is caused by
```
[appearance.background_color]
base = "#3b4252"
weak = "#2e3440"
# strong = "#3b4252"
```

Your strong value is equal to the base one. Iced uses the strong value for the slider lighter track color

romanstingler:
thanks a lot, I will test it after work.



image


yes that was it. 



thanks and sorry :sweat_smile:

MalpenZibo:
ok. Then, if it's ok for you, I would like to proceed with the merge

src/config.rs
clotodex, line 1: I feel like this could benefit from .into and .from trait implementations. This is more intuitive to a rust dev and makes some code more clean. Instead of "what helper to use" it might just be a .into() or the functions can take an Into<..> or asrefs etc.
However, I dont think this is necessary, nor might you agree with this @MalpenZibo, so approving independent of this.

clotodex (approved):
🎉

src/config.rs
clotodex: talking about the hex converters and the pair construction

MalpenZibo: Something can be done, but HexColor and Color are foreign types, other cases take fallback params or return Optional value

Issue #471, opened 2026-02-06
Broken on NVIDIA + Niri


Tykveg:
Both 0.7.0 and current main are broken on nvidia with niri. The bar freezes instantly upon the start. Only clicking on custom modules and selecting workspaces works (though nothing updates, everything is frozen).
OS: NixOS
Nvidia drivers version: 580.126.09
For getting ashell from master using overlay:
```nix
# https://github.com/NixOS/nixpkgs/blob/nixos-unstable/pkgs/by-name/as/ashell/package.nix

final: prev: {
  ashell = prev.ashell.overrideAttrs (finalAttrs: prevAttrs: {
    src = prev.fetchFromGitHub {
      owner = "MalpenZibo";
      repo = "ashell";
      rev = "845ca24797d5acc313ec38aafe4ec2efb2c67458";
      hash = "sha256-xu3pbDo62tZ000Znl0mts2XhYe7ZMD9RuJqQzhBb2x4=";
    };
    # https://discourse.nixos.org/t/overriding-version-cant-find-new-cargohash/31502/7
    cargoDeps = prev.rustPlatform.fetchCargoVendor {
      src = finalAttrs.src;
      hash = "sha256-MkFkf/q5UqR2VNBUzPAIOawE06aI9spitjeL7xhJe9U=";
    };
    #cargoHash = "";
  });
}
```

<details><summary>Log (on main branch, 0.7.0 is pretty much the same)</summary>

```
INFO [ashell::config] Decoding config file "/home/user/.config/ashell/config.toml"
INFO [ashell::config] Config file loaded successfully
WARN [ashell::services::brightness] No backlight devices found
ERROR [ashell::services::brightness] Failed to access to brightness files: No backlight devices found
ERROR [ashell::services::brightness] Brightness service error
ERROR [ashell::services::upower] Failed to initialize upower service: org.freedesktop.DBus.Error.ServiceUnknown: The name org.freedesktop.UPower was not provided by any .service files
WARN [wgpu_hal::gles::egl] No config found!
WARN [wgpu_hal::gles::egl] EGL says it can present to the window but not natively
WARN [wgpu_hal::gles::adapter] Max vertex attribute stride unknown. Assuming it is 2048
WARN [wgpu_hal::gles::egl] Re-initializing Gles context due to Wayland window
WARN [wgpu_hal::gles::egl] No config found!
WARN [wgpu_hal::gles::egl] EGL says it can present to the window but not natively
[...]

PR #595, opened 2026-04-02
docs: update troubleshooting guide for rendering backends

romanstingler:
Updated the troubleshooting guide to better document rendering backend options and common visual issues.

## Changes

- Restructured the "NVIDIA Driver Issues" section into "Visual Issues, Freezing, or Rendering Problems"
- Documented both rendering paths: `wgpu` (with Vulkan/OpenGL backends) and `tiny-skia` (CPU renderer)
- Added table of valid `WGPU_BACKEND` values and their aliases
- Clarified backend behavior and fallback mechanism

## Rendering Backends Covered

1. **wgpu** (default) - Uses Vulkan by default, configurable via `WGPU_BACKEND` env var
2. **tiny-skia** - CPU renderer, can be forced with `ICED_BACKEND=tiny-skia`

### DIFF
--- a/website/docs/configuration/troubleshooting.md
+++ b/website/docs/configuration/troubleshooting.md
@@ -6,21 +6,42 @@ sidebar_position: 5
 
 Common issues and quick fixes for ashell.
 
-## NVIDIA Driver Issues
+## Visual Issues, Freezing, or Rendering Problems
 
-**Problem:** ashell freezes on startup when using NVIDIA graphics on Linux.
+**Problem:** ashell may freeze, show visual artifacts, or fail to render properly on some systems.
 
-**Cause:** The default Vulkan backend may have compatibility issues with NVIDIA drivers.
+**Cause:** Graphics driver compatibility issues. By default, ashell uses `wgpu` which uses `vulkan` as the default backend.
 
-**Solution:** Force the use of OpenGL by setting the `WGPU_BACKEND` environment variable:
+### Rendering Backends
 
-```bash
-WGPU_BACKEND=gl ashell
-```
+ashell supports two rendering paths:
 
-This uses EGL as the context creation API, bypassing Vulkan entirely.
+1. **wgpu** (default) - Uses Vulkan by default. You can force a specific backend using the `WGPU_BACKEND` environment variable:
+   ```bash
+   WGPU_BACKEND=vulkan ashell  # Force Vulkan
+   WGPU_BACKEND=gl ashell     # Force OpenGL/EGL
+   ```
 
-**If the issue persists:** Try different NVIDIA drivers or use a different compositor.
+2. **tiny-skia** (CPU renderer) - Falls back automatically if wgpu fails, or can be forced with:
+   ```bash
+   ICED_BACKEND=tiny-skia ashell
+   ```
+
+### Valid `WGPU_BACKEND` values
+
+You can use a single value or a comma-separated list (e.g. `WGPU_BACKEND=gl,vulkan`). Values are case-insensitive.
+
+| Value | Aliases | Description |
+|---|---|---|
+| `vulkan` | `vk` | Vulkan API |
+| `gl` | `gles`, `opengl` | OpenGL / OpenGL ES (EGL)
+
+### Backend Behavior
+
+- ashell tries `wgpu` first, then falls back to `tiny-skia` if wgpu fails
+- `tiny-skia` is a CPU renderer with less RAM consumption but potentially higher battery usage
+
+**If the issue persists:** Try forcing a different backend or using `ICED_BACKEND=tiny-skia`.
 
 ## Idle Inhibitor Issues
 


clotodex:
This is much nicer than i expected tbh

I really like the auto idea

My thoughts:
- could auto be default and if an env var is set then auto is unused (i have set some if these vara in my compositor so it "would just work" with what i configured there) - in that case config would not be needed
- the nvidia detection is "greedy". If you have a card active but dont use it for your compositor it falls to opengl for no reason. (E.g. i use the gpu for cuda, but my compositor runs on igpu only) - my best guess would be to detect it from other env vars, at least i am not aware of a wayland protocol that shares this info

Wdyt?

romanstingler:
@clotodex thanks for the feedback, 

> could auto be default

it is. auto isdefault and it only overwrites the env for  Nvidia users, for all others no env will be set, therefore WGPU keeps its internal auto detection. 

> and if an env var is set then auto is unused 

I changed that, if the env var is detected this one has ALWAYS precedence over our config

> the nvidia detection is "greedy"

yes, you are right. I changed that too and am using env vars to detect it. we can always try to improve the detection if needed.

clotodex:
Nice! Do we need the config setting at all then?
If it is auto by default and can be overriden with env vars we could drop the config as well (except if we think config is cleaner)

romanstingler:
the config setting if you want to have it permanently and don't want to change your launch settings for whatever reason. 
lets say you don't have it in autostart but you start it later manually(not my use-case). 
tbh I am fine with both ways. 
@MalpenZibo maybe you have 1 min for your opinion

MalpenZibo:
I'm not 100% sure. A config field seems less hacky then an env var. But still remains a non necessary addition.

I'm also thinking about use opengl as default and let user opt-in for vulkan.

This will let us drop all the Nvidia related shit. The performance loss for a status bar should be trivial and maybe (I need to check) we can reduce the memory usage that for many users seems an important things.

MalpenZibo:
OpenGL doesn't seem to reduce the RAM usage on my machine. Maybe ~2MB on 200 MB of RAM consumption.

From my point of view, we could maintain the Nvidia workaround to reduce the number of Nvidia-related issues. I will drop the renderer config, leaving the user to use the env var directly. 

The only thing, maybe in the Nvidia check, we could check if the user set the render in the env var to avoid overriding their setting

MalpenZibo:
To complete this PR. I think that we can:
- drop the configuration
- the user should use the env var (we can document this behavior)
- if no env var is specified and we detect the nvidia gpu -> auto switch to egl (document this behavior)

What do you think?

romanstingler:
fine for me, there was just one comment in the issues that made me unsure. And I wanted to have a look 
https://github.com/MalpenZibo/ashell/issues/471#issuecomment-4251936333


@MalpenZibo pushed the updated code

MalpenZibo:
@romanstingler  About #471. It seems that some combination makes even GL not work. In that specific case, now Vulkan works instead GL, which does not work, and falls back on tiny-skia (which is now enabled).

For this reason, we do not have a safe baseline. Even GL could not work. 

Probably the best thing that we could do is to update the documentation to specify that:
- `ashell` by default uses `wgpu`, which by default uses `vulkan`.
- Using `WGPU_BACKEND` is possible to force `vulkan` or `gl` 
- `ashell` will try first using `wgpu` and then fall back to `tiny-skia`
- Using `ICED_BACKEND=tiny-skia` will force the usage of `tiny-skia`

Tiny-skia is a CPU renderer, so less performance and less ram consumption. (more battery consumption? I don't know)

Sidenote - I had to create a small fork of `softbuffer` to make tiny-skia work with transparency

romanstingler:
Updated the docs

website/docs/configuration/troubleshooting.md
```diff
@@ -6,21 +6,42 @@ sidebar_position: 5
...
+   ```
+
+2. **tiny-skia** (CPU renderer) - Falls back automatically if wgpu fails, or can be forced with:
+   ```bash
+   ICED_BACKEND=tiny-skia ashell
+   ```
+
+### Valid `WGPU_BACKEND` values
+
+You can use a single value or a comma-separated list (e.g. `WGPU_BACKEND=gl,vulkan`). Values are case-insensitive.
```
MalpenZibo, line 32: Is this true? I never test it

MalpenZibo (commented):
Nice! Can we also add a section about the RAM consumption (wgpu -> high RAM usage, tiny-skia -> low RAM usage)

website/docs/configuration/troubleshooting.md
romanstingler: https://github.com/gfx-rs/wgpu#environment-variables

WGPU_BACKEND with a comma-separated list of the backends you want to use (vulkan, metal, dx12, or gl).

it tries to init the first one, if it fails it tries the next one. 
if the experience is bad but it initializes you are stuck with that one

mati865:
Also this part of the description is outdated now:
> Auto mode (default): automatically uses OpenGL when proprietary NVIDIA driver is detected, Vulkan for all other hardware (AMD/Intel/nouveau)

romanstingler:
> Also this part of the description is outdated now:
> 
> > Auto mode (default): automatically uses OpenGL when proprietary NVIDIA driver is detected, Vulkan for all other hardware (AMD/Intel/nouveau)

have you checked an old commit ??

MalpenZibo:
I think that the "problem" is in the PR description. Code side is ok to me!

MalpenZibo:
@romanstingler can you change the title and description? It could be useful for the changelog

PR #600, opened 2026-04-04
tray: skip empty icon name lookup

SimoneFelici:
Hi,
This fixes this issue: https://github.com/MalpenZibo/ashell/issues/598

### DIFF
--- a/src/services/tray/mod.rs
+++ b/src/services/tray/mod.rs
@@ -54,6 +54,11 @@ static SYSTEM_ICON_ENTRIES: LazyLock<Vec<(Cow<'static, str>, Cow<'static, str>)>
     });
 
 fn get_icon_from_name(icon_name: &str) -> Option<TrayIcon> {
+    let icon_name = icon_name.trim();
+    if icon_name.is_empty() {
+        return None;
+    }
+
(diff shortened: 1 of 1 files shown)


clotodex (approved):
happy to merge
wondering if stripping whitespace is necessary as well (or at least useful).
this would mainly apply to dbus based icon names i think, i dont know if dbus has auto-stripping. Or if we should care at all?
@romanstingler do you have any experience from your optimization deep dive?

romanstingler:
@clotodex D-Bus does not auto-strip whitespaces.
According to 
https://specifications.freedesktop.org/icon-naming/latest/
icon names shouldn't contain spaces, malformed applications though might send names with accidental whitespace.

So, I think it would be not required, but a defensive measure. 
Something like that should be fine.

```
fn get_icon_from_name(icon_name: &str) -> Option<TrayIcon> {
    let icon_name = icon_name.trim();
    if icon_name.is_empty() {
        return None;
    }
    ....
 ```

SimoneFelici:
>     let icon_name = icon_name.trim();
Should I update my PR to include this?

clotodex:
Yes please :)

PR #617, opened 2026-04-10
[Regression] Audio devices not shown

denqxotl:
This problem is regression of #449  

Before changes:
2026-04-10-103744_hyprshot

After changes:
2026-04-10-103755_hyprshot

### DIFF
--- a/src/modules/settings/audio.rs
+++ b/src/modules/settings/audio.rs
@@ -364,8 +364,7 @@ impl AudioSettings {
                             .port
                             .and_then(Self::port_icon)
                             .unwrap_or(StaticIcon::Speaker3),
-                        active: route.is_active()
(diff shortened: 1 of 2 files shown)


clotodex:
@MalpenZibo / @romanstingler you worked on the mentioned PR back then. could one of you check this out please?

romanstingler:
yes looks good to me. 
back then filters were not shown and when I tested with a noise filter it was properly displayed. 
I am not an audio power user and didn't see any suspicious behavior back then. 

But yeah, it just shows more devices (not on my computers) but that's fine.

MalpenZibo (approved):
I'm not an audio power user. From the code point of view, it makes sense. I cannot perform a deep test because I'm not using these kinds of configurations.

Issue #616, opened 2026-04-10
Workspace scroll direction is reversed after [e8df2804a60dd8dddc6da089d8b5b5ceba02c30f]


lkcv:
Hi, it looks commit https://github.com/MalpenZibo/ashell/commit/e8df2804a60dd8dddc6da089d8b5b5ceba02c30f reversed the scroll direction again.

Can the behavior be changed back to what it was so that scrolling up goes the previous workspace, and scrolling down goes to the next one?
If not, maybe this could be configurable?

Here's a previous issue about when it happened before and got resolved: https://github.com/MalpenZibo/ashell/issues/358

clotodex:
This should definitely be configurable then. I remember a discussion with @MalpenZibo when we merged #328 
For niri, or at least my setup, the current scroll direction is more intuitive. Thank you for tracking the issue.
@lkcv do you want to open a PR? If not I will when i get to it

lkcv:
@clotodex I can open a PR on it. It seems relatively straightforward to implement.

MalpenZibo:
I notice that they're too many different opinion about the scroll direction so probably the config option make sense.

Also now with the new iced version we could be able to enable the horizontal scroll that for the workspace module could make sense.

MalpenZibo:
https://github.com/MalpenZibo/ashell/pull/635

Here, I fixed the scroll behavior. Now respect the compositor preferences. After this change, test the solution and let me know if the flag is still needed.

Also, I notice that niri jumps around between workspaces using the scroll.

lkcv:
I tested on the latest master after that was pushed, but I have to set my scrolling in hyprland to natural (which inverts everything) in order to get the scrolling style I want. I guess I'm just used to scrolling "backwards" lol.

If you're okay with it, I'd like to add the config flag.

MalpenZibo:
ok, let's proceed with the config flag

lkcv:
Forgot to close this when (https://github.com/MalpenZibo/ashell/pull/622) got merged.

PR #622, opened 2026-04-11
Add configurable workspace scrolling direction

lkcv:
As discussed here: https://github.com/MalpenZibo/ashell/issues/616, adds a setting to configure the workspace scroll mode.

2 modes:
1. Natural, the default option, makes it so scrolling up will go to the next sequential workspace and scrolling down will go to the previous one. This is the current behavior in master.
2. Traditional is the opposite of this.

I'm not very good with naming things, so let me know if these need to change.

Tested on hyprland and niri.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -121,9 +121,8 @@ pub struct WorkspacesModuleConfig {
 pub enum InvertScrollDirection {
     #[default]
     All,
-    Mice,
-    Trackpads,
-    None,
+    Mouse,
+    Trackpad,
 }
 
 #[derive(Deserialize, Copy, Clone, Default, PartialEq, Eq, Debug)]
--- a/src/config.rs
+++ b/src/config.rs
@@ -112,6 +112,7 @@ pub struct WorkspacesModuleConfig {
     pub max_workspaces: Option<u32>,
     pub workspace_names: Vec<String>,
     pub enable_virtual_desktops: bool,
+    pub invert_workspace_scroll_direction: bool,
 }
 
 #[derive(Deserialize, Copy, Clone, Default, PartialEq, Eq, Debug)]
--- a/src/config.rs
+++ b/src/config.rs
@@ -114,7 +114,16 @@ pub struct WorkspacesModuleConfig {
     pub max_workspaces: Option<u32>,
     pub workspace_names: Vec<String>,
     pub enable_virtual_desktops: bool,
-    pub invert_workspace_scroll_direction: bool,
+    pub invert_scroll_direction: Option<InvertScrollDirection>,
+}
+
+#[derive(Deserialize, Copy, Clone, Default, PartialEq, Eq, Debug)]
+pub enum InvertScrollDirection {
+    #[default]
+    All,
+    Mice,
+    Trackpads,
+    None,
 }
 
 #[derive(Deserialize, Copy, Clone, Default, PartialEq, Eq, Debug)]
--- a/src/modules/workspaces.rs
+++ b/src/modules/workspaces.rs
@@ -482,14 +482,22 @@ impl Workspaces {
         )
         .on_scroll(move |direction| match direction {
             iced::mouse::ScrollDelta::Lines { y, .. } => {
-                match self.config.invert_scroll_direction {
-                    Some(InvertScrollDirection::All | InvertScrollDirection::Mice) => {
-                        Message::Scroll(-y as i32)
+                if y.is_sign_positive() {
+                    match self.config.invert_scroll_direction {
+                        Some(InvertScrollDirection::All | InvertScrollDirection::Mouse) => {
+                            Message::Scroll(-1)
+                        }
+                        Some(InvertScrollDirection::Trackpad) => Message::Scroll(1),
+                        None => Message::Scroll(1),
                     }
-                    Some(InvertScrollDirection::Trackpads | InvertScrollDirection::None) => {
-                        Message::Scroll(y as i32)
+                } else {
+                    match self.config.invert_scroll_direction {
+                        Some(InvertScrollDirection::All | InvertScrollDirection::Mouse) => {
+                            Message::Scroll(1)
+                        }
+                        Some(InvertScrollDirection::Trackpad) => Message::Scroll(-1),
+                        None => Message::Scroll(-1),
                     }
-                    None => Message::Scroll(y as i32),
                 }
             }
             iced::mouse::ScrollDelta::Pixels { y, .. } => {
@@ -497,15 +505,21 @@ impl Workspaces {
 
                 if self.scroll_accumulator.abs() < sensibility {
                     Message::ScrollAccumulator(y)
+                } else if self.scroll_accumulator.is_sign_positive() {
+                    match self.config.invert_scroll_direction {
+                        Some(InvertScrollDirection::All | InvertScrollDirection::Trackpad) => {
+                            Message::Scroll(1)
+                        }
+                        Some(InvertScrollDirection::Mouse) => Message::Scroll(-1),
+                        None => Message::Scroll(-1),
+                    }
                 } else {
                     match self.config.invert_scroll_direction {
-                        Some(InvertScrollDirection::All | InvertScrollDirection::Trackpads) => {
-                            Message::Scroll(-y as i32)
+                        Some(InvertScrollDirection::All | InvertScrollDirection::Trackpad) => {
+                            Message::Scroll(-1)
                         }
-                        Some(InvertScrollDirection::Mice | InvertScrollDirection::None) => {
-                            Message::Scroll(y as i32)
-                        }
-                        None => Message::Scroll(y as i32),
+                        Some(InvertScrollDirection::Mouse) => Message::Scroll(1),
+                        None => Message::Scroll(1),
                     }
                 }
             }
--- a/src/modules/workspaces.rs
+++ b/src/modules/workspaces.rs
@@ -508,18 +508,18 @@ impl Workspaces {
                 } else if self.scroll_accumulator.is_sign_positive() {
                     match self.config.invert_scroll_direction {
                         Some(InvertScrollDirection::All | InvertScrollDirection::Trackpad) => {
-                            Message::Scroll(1)
-                        }
-                        Some(InvertScrollDirection::Mouse) => Message::Scroll(-1),
-                        None => Message::Scroll(-1),
-                    }
-                } else {
-                    match self.config.invert_scroll_direction {
-                        Some(InvertScrollDirection::All | InvertScrollDirection::Trackpad) => {
                             Message::Scroll(-1)
                         }
                         Some(InvertScrollDirection::Mouse) => Message::Scroll(1),
                         None => Message::Scroll(1),
+                    }
+                } else {
+                    match self.config.invert_scroll_direction {
+                        Some(InvertScrollDirection::All | InvertScrollDirection::Trackpad) => {
+                            Message::Scroll(1)
+                        }
+                        Some(InvertScrollDirection::Mouse) => Message::Scroll(-1),
+                        None => Message::Scroll(-1),
                     }
                 }
             }
--- a/src/modules/workspaces.rs
+++ b/src/modules/workspaces.rs
@@ -1,5 +1,5 @@
 use crate::{
-    config::{WorkspaceVisibilityMode, WorkspacesModuleConfig},
+    config::{InvertScrollDirection, WorkspaceVisibilityMode, WorkspacesModuleConfig},
     outputs::Outputs,
     services::{
         ReadOnlyService, Service, ServiceEvent,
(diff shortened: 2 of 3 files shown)


clotodex:
I would call rhe config item "scroll direction" for sure
With the variants i am not as sure
I get where you are coming from and it might be intuitive for people
I was thinking something like inverted and natural, but need to get the order right :D

Natural: scroll down → next workspace (content moves with your fingers, like a touchscreen)
Reverse/inverted: scroll down → previous workspace (traditional mouse-style, where scrolling down moves the "viewport" down)

macOS calls it natural. Other options: "Default / Inverted" or "Content direction / Scroll direction."

@MalpenZibo would love to get your take 
(On phone so unsure if the ping goes through)

lkcv:
Changed scroll mode to direction and updated docs to show that instead. 
For the variant name, we could add an alternate one for each to the enum and have multiple arms in the scroll_direction match with things like Reverse and Inverted returning the same value , but I'm not sure if that's the best approach. Too many and it would clutter the the documentation.

I'll wait to see what MalpenZibo says before adding or changing those strings.

MalpenZibo:
Ok, I understand the regression (it's probably caused by the new iced_layershell library).

About the config things. I would like to avoid this kind of configuration and rely on the compositor config. So I would check if I can fix this in iced_layershell to follow the compositor settings. It should be possible, and for example, the updates module should already do that (in theory).

In case I cannot fix this, then I will take into consideration the config flag. Do you see reason to let the user give a specific ashell config to override the compositor setting for the workspace module and the other places in the bar, like the audio volume (I suspect that the error/regression is present also there)

lkcv:
It looks like ashell already follows the compositor config (at least for hyprland).

The config I used for testing:

input {
	numlock_by_default = true
	natural_scroll = false
}

device {
	name = logitech-g502-x-plus-1
	natural_scroll = true
}

Like hyprland, ashell will look at a device config and override the global value if it's defined (false in this test) and update the scrolling direction when it's changed.

However, I have to enable natural_scrolling in hyprland to achieve my desired ashell workspace scrolling direction, and this messes up how I scroll in every other application since I'm using a mouse and not a trackpad.

That's why I think we should have config flags to override some of these. I can open more PRs to add flags to the other modules if you're okay with going that route.

lkcv:
MalpenZibo, for the variant names, do you think they should be named something else? Like clotodex pointed out, these names might not necessarily be intuitive to everyone. I could expand the enum to include more that return the same value, but I'm hesitant to clutter the list.

MalpenZibo:
So, here we still have an issue because after the recent change to the iced_layershell, we should not invert the scroll direction between the pixels and lines event. The position_button is fine instead. So I will open a PR for that.

Then, about the config field name. If it's still necessary, I think that a simple `invert_workspace_scroll_direction: bool` should be enough.

MalpenZibo:
https://github.com/MalpenZibo/ashell/pull/654

lkcv:
I rebuilt from master again after pulling that, but scrolling up goes the next workspace instead of the previous one. I'm using a mouse, not a trackpad, so that might be why. I don't have one to test that change properly.

I'll work on changing it to a bool instead of an enum.

lkcv:
Changed it to a bool.

src/config.rs
```diff
@@ -114,6 +112,7 @@ pub struct WorkspacesModuleConfig {
     pub max_workspaces: Option<u32>,
     pub workspace_names: Vec<String>,
     pub enable_virtual_desktops: bool,
+    pub invert_workspace_scroll_direction: bool,
```
MalpenZibo: Because it's already in the workspace module section, maybe we could simply call this field `invert_scroll_direction`

MalpenZibo:
> I rebuilt from master again after pulling that, but scrolling up goes the next workspace instead of the previous one. I'm using a mouse, not a trackpad, so that might be why. I don't have one to test that change properly.
> 
> I'll work on changing it to a bool instead of an enum.

I'm going mad with this stuff ahaha. Now the behavior between the mouse and the trackpad is coherent, but as you notice is the opposite of what I expect.

Can you fix this in this PR? We need to flip the `default` direction for both lines and pixels scroll

lkcv:
😅 sorry. I'll get this added to the PR in a bit.

Just making sure I understand, I'll make the pixels scrolling read from the bool that I added for lines scrolling, and we'll keep the default -1, right? Makes sense since that's consistent with volume scrolling and the others from what I checked.

MalpenZibo:
Let's focus only on the workspace for now.

I would like to:
- Invert the default direction for both lines and pixels case: right now they work consistently but are reversed
- on top of that, let's add a config field `invert_scroll_direction` that is an Option<All | Lines | Pixels> (you can change the name of the possible values if you have somethign better)
  - None -> no changes
  - All -> invert both pixels and lines direction
  - Lines -> invert only the lines case
  - Pixels -> invert only the pixels case

lkcv:
Okay, so I reversed the scrolling directions as requested so that scrolling and swiping up on the trackpad goes to the previous workspace by default. They're still consistent with each other.

also, I changed the bool to an Option enum with an explicit None (yeah, None shows up twice in the match since it's wrapped in Some). And I changed Lines to "Mice" and Pixels to "Trackpads" because it's easier to remember what it's for. Let me know if I need to change those back.

If you're good with these changes, I'll update the docs and then squash these commits.

clotodex:
The double none makes no sense, right?

Option:
- None = no changes
- some(all) = both change
- some(mouse) = mouse changes
- some(trackpad) = trackpad changes

(Would also use singular naming)

Ao no none in the enum

Or am i missing something?

src/config.rs
```diff
@@ -114,6 +114,16 @@ pub struct WorkspacesModuleConfig {
...
+}
+
+#[derive(Deserialize, Copy, Clone, Default, PartialEq, Eq, Debug)]
+pub enum InvertScrollDirection {
+    #[default]
+    All,
+    Mice,
+    Trackpads,
+    None,
 }
```
MalpenZibo, line 126: The field default should be  `None`
```suggestion
pub enum InvertScrollDirection {
    #[default]
    All,
    Mouse,
    Trackpad,
```

src/modules/workspaces.rs
```diff
@@ -482,21 +482,31 @@ impl Workspaces {
...
-                    Message::Scroll(1)
+                    match self.config.invert_scroll_direction {
+                        Some(InvertScrollDirection::All | InvertScrollDirection::Trackpads) => {
+                            Message::Scroll(-y as i32)
+                        }
+                        Some(InvertScrollDirection::Mice | InvertScrollDirection::None) => {
+                            Message::Scroll(y as i32)
+                        }
+                        None => Message::Scroll(y as i32),
+                    }
```
MalpenZibo: We cannot use -y here. In general, also for the other case let's maintain the the fixed `-1` or `1` value.

Here it's a bug: `y` could be a large value, and I don't think that the user wants to scroll so much 😅 In the Lines cases, in theory the event should be called line by line, but just to be sure it's better to use fixed `-1 | 1` values

MalpenZibo (changes requested):

lkcv:
> The double none makes no sense, right?
> 
> Option:
> 
>     * None = no changes
> 
>     * some(all) = both change
> 
>     * some(mouse) = mouse changes
> 
>     * some(trackpad) = trackpad changes
> 
> 
> (Would also use singular naming)
> 
> Ao no none in the enum
> 
> Or am i missing something?

I'll get rid of the explicit 'None' since it's kind of pointless and makes this more confusing.

The other 2 will be 'Mouse' and 'Trackpad'

src/config.rs
lkcv: Whoops. To be clear, we don't want an explicit "None" or "Disabled", right?

src/modules/workspaces.rs
lkcv: Changing ```-y``` to ```-1``` works. But when I change ```y``` to ```1```, I can no longer scroll down. I'll try to get this working.

src/config.rs
MalpenZibo: Right!

lkcv: Sorry for not catching that at first. It makes sense since we have to handle "None" anyway. Won't be as cluttered.

src/modules/workspaces.rs
lkcv: I got this working in the latest commit. It's slightly verbose, but easy to read in my opinion. 
Let me know what you think. 

I had to to take the sign of ```y``` and ```scroll_accumulator``` into account since I wasn't just passing ```y``` anymore. 😅

src/config.rs
lkcv: Done

src/modules/workspaces.rs
```diff
@@ -494,9 +506,21 @@ impl Workspaces {
                 if self.scroll_accumulator.abs() < sensibility {
                     Message::ScrollAccumulator(y)
                 } else if self.scroll_accumulator.is_sign_positive() {
-                    Message::Scroll(-1)
+                    match self.config.invert_scroll_direction {
+                        Some(InvertScrollDirection::All | InvertScrollDirection::Trackpad) => {
+                            Message::Scroll(1)
+                        }
+                        Some(InvertScrollDirection::Mouse) => Message::Scroll(-1),
+                        None => Message::Scroll(-1),
```
MalpenZibo: This is not the inverse of what we want? In the other case for y > 0 we have:
- inverted => -1
- normal => 1

src/modules/workspaces.rs
```diff
@@ -494,9 +506,21 @@ impl Workspaces {
...
+                    }
                 } else {
-                    Message::Scroll(1)
+                    match self.config.invert_scroll_direction {
+                        Some(InvertScrollDirection::All | InvertScrollDirection::Trackpad) => {
+                            Message::Scroll(-1)
+                        }
+                        Some(InvertScrollDirection::Mouse) => Message::Scroll(1),
+                        None => Message::Scroll(1),
+                    }
```
MalpenZibo, line 523: same as before

MalpenZibo (changes requested):

src/modules/workspaces.rs
MalpenZibo: wait, I'm quite confuse. Before changing this let me test

src/modules/workspaces.rs
lkcv: when i had the values flipped to match the lines match block, it scrolled the opposite direction as the mouse (tested on a ps4 controller touchpad).

this was without the config option present, so no direction flipping should have been happening.

src/modules/workspaces.rs
lkcv: Okay, i'll wait

MalpenZibo (changes requested):
This is the correct behavior for the pixel case
```rust
            iced::mouse::ScrollDelta::Pixels { y, .. } => {
                let sensibility = 3.;

                if self.scroll_accumulator.abs() < sensibility {
                    Message::ScrollAccumulator(y)
                } else if self.scroll_accumulator.is_sign_positive() {
                    match self.config.invert_scroll_direction {
                        Some(InvertScrollDirection::All | InvertScrollDirection::Trackpad) => {
                            Message::Scroll(-1)
                        }
                        Some(InvertScrollDirection::Mouse) => Message::Scroll(1),
                        None => Message::Scroll(1),
                    }
                } else {
                    match self.config.invert_scroll_direction {
                        Some(InvertScrollDirection::All | InvertScrollDirection::Trackpad) => {
                            Message::Scroll(1)
                        }
                        Some(InvertScrollDirection::Mouse) => Message::Scroll(-1),
                        None => Message::Scroll(-1),
                    }
                }
            }
```

The complete code for reference
```rust
        .on_scroll(move |direction| match direction {
            iced::mouse::ScrollDelta::Lines { y, .. } => {
                if y.is_sign_positive() {
                    match self.config.invert_scroll_direction {
                        Some(InvertScrollDirection::All | InvertScrollDirection::Mouse) => {
                            Message::Scroll(-1)
                        }
                        Some(InvertScrollDirection::Trackpad) => Message::Scroll(1),
                        None => Message::Scroll(1),
                    }
                } else {
                    match self.config.invert_scroll_direction {
                        Some(InvertScrollDirection::All | InvertScrollDirection::Mouse) => {
                            Message::Scroll(1)
                        }
                        Some(InvertScrollDirection::Trackpad) => Message::Scroll(-1),
                        None => Message::Scroll(-1),
                    }
                }
            }
            iced::mouse::ScrollDelta::Pixels { y, .. } => {
                let sensibility = 3.;

                if self.scroll_accumulator.abs() < sensibility {
                    Message::ScrollAccumulator(y)
                } else if self.scroll_accumulator.is_sign_positive() {
                    match self.config.invert_scroll_direction {
                        Some(InvertScrollDirection::All | InvertScrollDirection::Trackpad) => {
                            Message::Scroll(-1)
                        }
                        Some(InvertScrollDirection::Mouse) => Message::Scroll(1),
                        None => Message::Scroll(1),
                    }
                } else {
                    match self.config.invert_scroll_direction {
                        Some(InvertScrollDirection::All | InvertScrollDirection::Trackpad) => {
                            Message::Scroll(1)
                        }
                        Some(InvertScrollDirection::Mouse) => Message::Scroll(-1),
                        None => Message::Scroll(-1),
                    }
                }
            }
        })
```

Sorry for the confusion

lkcv:
No worries. The funny thing is that I had that initially during my testing, but my PS4 touchpad scrolls oppositely from my mouse. If that's the intention (to swipe up on a touchpad to go to previous workspace), I'll update the docs, flip these values like in your suggestion and then push.

lkcv:
Should be ready now

MalpenZibo (approved):
🥳

PR #625, opened 2026-04-14
(Chore) Notification refactor

MalpenZibo:
## Config refactor

Removed some configuration to simplify things:
- max notifications
- toast_width
- toast_summary_line_budget
- toast_body_line_budget
- empty_state_height

Renamed other config

## UI Rework
I changed a little the UI to simplify the code and align with the rest of the application. 

I moved the icon logic to the service and refactored that part.

## Iced Layer changes
Introduced the `input_region`, we can create a full height surface and set the input region only on a specific region. This lets us have a dynamic list of toast and let the user interact with the windows behind the ashel toast surface.

## Missing pieces
- notification timeout progressibar?
- stop timeout on mouseover
- critical level (change the card border color)

### DIFF
--- a/src/modules/notifications.rs
+++ b/src/modules/notifications.rs
@@ -15,8 +15,9 @@ use chrono::{DateTime, Local};
 use freedesktop_icons::lookup;
 use iced::{
     Alignment, Background, Border, Color, Column, Element, Length, Row, Subscription, Task, Theme,
-    widget::{Space, button, column, container, image, row, rule, scrollable, svg, text},
+    widget::{Space, button, column, container, image, row, scrollable, svg, text},
 };
+use itertools::Itertools;
 use linicon_theme::get_icon_theme;
 use log::error;
 use std::{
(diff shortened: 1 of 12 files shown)


src/modules/notifications.rs
```diff
@@ -382,17 +208,13 @@ impl Notifications {
 
         match update_event {
             NotificationEvent::Received(notification) => {
-                let was_empty = self.toasts.is_empty();
-                while self.toasts.len() >= self.config.toast_max_visible {
+                while self.toasts.len() >= self.config.toast_limit {
```
romanstingler, line 218: I tested setting `toast_limit` to 0 in config and then 
```
 notify-send -i "🔥" "Fire" "Something hot"
  notify-send -i "⚠️" "Warning" "Pay attention"
  notify-send -i "✅" "Done" "Task completed"
```
hangs the bar.

We probably want something like 

```
        match update_event {
            NotificationEvent::Received(notification) => {
                if self.config.toast_limit == 0 {
                    self.toasts.clear();
                    return Action::None;
                }

                while self.toasts.len() >= self.config.toast_limit {
                    self.toasts.pop_front();
                }
....
```

Colors and size look OK. Tested it works fine other than the case above.

PR #634, opened 2026-04-15
fix(workspaces): stabilize monitor group order on Niri

MalpenZibo:


clotodex (approved):
🚀 

Not tested yet but code changes look fine

PR #636, opened 2026-04-15
fix(workspaces): scroll in visual order on Niri

MalpenZibo:


src/modules/workspaces.rs
```diff
@@ -360,14 +363,14 @@ impl Workspaces {
                     self.ui_workspaces
                         .iter()
                         .filter(|w| in_current_group(w))
-                        .filter(|w| w.id < current_workspace_id)
-                        .max_by_key(|w| w.id)
+                        .filter(|w| (w.index, w.id) < current_key)
```
clotodex: why do we filter by the tuple comparison and not index?
Either this is stable on id and index comparison should be enough, or the id is unstable and the comparison makes no sense.
Might be my post-work brain not getting it, but please do explain :)
Also goes for the next max_by_key line

src/modules/workspaces.rs
```diff
@@ -360,14 +363,14 @@ impl Workspaces {
...
+                        .max_by_key(|w| (w.index, w.id))
                 } else {
                     self.ui_workspaces
                         .iter()
                         .filter(|w| in_current_group(w))
-                        .filter(|w| w.id > current_workspace_id)
-                        .min_by_key(|w| w.id)
+                        .filter(|w| (w.index, w.id) > current_key)
```
clotodex: same here

clotodex (approved):
Very nice cleanup :)
Pre-approved pending explanation.

src/modules/workspaces.rs
MalpenZibo: So currently I'm using 
```
[workspaces]
visibility_mode = "All"
enable_workspace_filling = true
group_by_monitor = true
```
and ashell only on my laptop settings. Apart from the fact that for niri the conceptually correct thing to do is to have ashell on each monitor and visibility_mode MonitorSpecific in my case, the problem is that the id is unique but not in visual order; instead, the index is in visual order but not unique. This means that with an external monitor and 3 workspaces for each monitor, I can have 1, 1, 2, 2, 3, 3. 

This will "resolve" the issue using the id to avoid jumping between the same index on different monitors, but now my pre-work brain is telling me that's quite a hacky solution and probably doesn't cover all possible cases, but only my specific configuration.

MalpenZibo:
@clotodex, this solution looks better.

clotodex (approved):
lgtm :)

PR #637, opened 2026-04-15
Notifications blacklist

denqxotl:
The prev notification daemon that I've used had this feature for blacklisting, it was convenient for me, so I decided to add this functionality to ashell

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -412,8 +412,7 @@ pub struct NotificationsModuleConfig {
     pub toast_timeout: u64,
     pub toast_limit: usize,
     pub toast_max_height: u32,
-    #[serde(default)]
-    pub blacklist: Vec<String>,
+    pub blocklist: Vec<RegexCfg>,
 }
 impl Default for NotificationsModuleConfig {
     fn default() -> Self {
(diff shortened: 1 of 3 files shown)


src/config.rs
```diff
@@ -412,6 +412,8 @@ pub struct NotificationsModuleConfig {
     pub toast_timeout: u64,
     pub toast_limit: usize,
     pub toast_max_height: u32,
+    #[serde(default)]
+    pub blacklist: Vec<String>,
```
MalpenZibo: Can we check the regex directly here instead of relying on `compile_blacklist`? Should be something similar in the `CustomModuleDef`

MalpenZibo (commented):
Quick thought on the naming. 

I'm personally pretty neutral on the whole blacklist/blocklist debate. Honestly, when I see terms like "blacklist" or "whitelist" my brain just goes straight to the technical meaning, I don't really associate them with any cultural connotation. So I don't have a strong stance on this.

That said, since I don't really care either way, terms like "blocklist" or "denylist" feel like reasonable alternatives. They're becoming more of a standard, and the meaning is clear enough. Kind of like how I switched from master to main on git: not because I had strong feelings about it, but the default changed, and I just went with it because... why not.

What do you think? Do you have a preference? @romanstingler @clotodex, what do you think?. 
Curious if anyone feels strongly either way or if we should go with the path of least friction.

src/config.rs
denqxotl: sure, I'll fix that.

denqxotl:
blacklist is replaced with blocklist.
updated code & docs

MalpenZibo (approved):

PR #641, opened 2026-04-16
(chore) Refactor/shared components

### LINKED ISSUE #566: Custom components
Create reusable components to simplify and standardize the codebase, minimizing code duplication across similar patterns. 

Possible components

#### Bar Indicator
```
A format-aware bar indicator: icon and/or text label.
The `format` prop controls what is shown:
- `Icon` → icon only
- `Percentage` / `Time` → label only
- `IconAndPercentage` / `IconAndTime` → icon + label
```

#### Buttons and Icon Button
```
Reusable button with hierarchy-based colors and optional icon

A simple Icon button
```

#### Expandable Panel
The one used in the update menu (maybe only there?)

#### Icons
Already present, just check if it needs a refactor to align with the style of the other components

#### Module Group and Module Items
The component responsible for displaying the modules inside a group, and each module item

#### Quick Settings
More or less, it's already present as a simple function in the settings module. Check if it can be improved and moved to the components module 

#### Selectable item
A list item: clickable transparent button when not selected, success-colored label when selected. Used for device lists (sinks, sources, etc.).

With an optional cta content (e.g. a remove button).

#### Slider
The one used in the menu. As for the Quick Settings, these are more or less already components in the form of a simple function, just need a refactor and to be moved to the components module

## Ashell Theme
Right now we pass around the ashell theme. It's the fastest solution possible from a performance point of view, but it's a pain in the ass. Could make sense to lose some performance and switch to a global access?

MalpenZibo:
Closes https://github.com/MalpenZibo/ashell/issues/566

This is a huge refactor!

I tried to simplify things by creating shared components (especially buttons). This should help new contributors to better follow the ashell style.

@clotodex @romanstingler there's no need for a full review, but just a check. Let me know what you think


src/components/quick_setting_button.rs
```diff
@@ -0,0 +1,77 @@
...
+                .style(theme.quick_settings_submenu_button_style(active))
+            }))
+            .spacing(theme.space.xxs)
+            .align_y(Alignment::Center)
+            .height(Length::Fill),
+    )
+    .padding([theme.space.xxs, theme.space.xs])
+    .on_press(on_press)
+    .height(Length::Fill)
+    .width(Length::Fill)
```
romanstingler: these 2 are immediately overwritten (duplicate)

src/components/icons.rs
```diff
@@ -311,12 +365,6 @@ pub fn icon_mono<'a>(icon: impl Icon) -> Text<'a> {
     icon.to_text_mono()
 }
 
-pub enum IconButtonSize {
-    Small,
-    Medium,
-    Large,
-}
-
 enum OnPress<'a, Message> {
```
romanstingler: there is the same `OnPress` enum in button.rs. 
do we want to use that one?

romanstingler (commented):
At the first glance, it didn't kill my laptop :) 

just a question, is this intentional 

src/components/format_indicator.rs:69-70
```
IndicatorState::Warning => theme.extended_palette().danger.weak.color,
```
and not 
```
IndicatorState::Warning => theme.extended_palette().warning.weak.color,
```

src/app.rs
```diff
@@ -412,7 +412,7 @@ impl App {
                 modules::notifications::Action::Task(task) => task.map(Message::Notifications),
                 modules::notifications::Action::Show(task) => {
                     let position = self.notifications.toast_position();
-                    let width = crate::menu::MenuSize::Medium.size() as u32;
+                    let width = crate::components::menu::MenuSize::Medium.size() as u32;
```
clotodex: MenuSize should likely be a direct import at this point. Wondering if we should do reexports for components

src/modules/updates.rs
```diff
@@ -258,19 +258,15 @@ impl Updates {
...
                     .width(Length::Fill),
             );
         }
 
         buttons
             .push(
-                button(row!(text("Check now").width(Length::Fill)))
-                    .style(theme.ghost_button_style())
-                    .padding(theme.space.xs)
+                styled_button(theme, "Check now")
```
clotodex, line 268: just spitballing here, but since the general button is already a builder pattern (.on_press etc) I could imagine keeping the same "interface"
button("Check now").style(theme).on_press
this could be "forced" so button() creates an UnstyledButton that needs to either be followed by .style(theme) or .unstyled() or so
There might even be a nicer pattern.
but the mixing of both and the leading theme instead of "Check now" feels of. At least switch the order of the args

src/modules/tray.rs
```diff
@@ -205,7 +208,10 @@ impl TrayModule {
                                     Message::ToggleMenu(item.name.to_owned(), id, button_ui_ref)
                                 })
                                 .padding(theme.space.xxs)
-                                .style(theme.ghost_button_style())
+                                .style(theme.button_style(
```
clotodex, line 211: so we are supporting styled_button() and manual styling?

clotodex: i think i would prefer forcing our custom component instead

src/modules/tempo.rs
```diff
@@ -416,46 +423,47 @@ impl Tempo {
...
+                        .size(theme.font_size.md),
                     )
-                    .size(theme.font_size.md),
-                )
-                .spacing(theme.space.xs)
+                    .spacing(theme.space.xs),
+                ),
             )
-            .padding([theme.space.sm, theme.space.lg])
+            .size(ButtonSize::Large)
```
clotodex, line 458: isnt this part of the theme? feels a bit weird to "throw after"

clotodex: at least could be a component. LargeButton
Outlined should likely be part of theme. also in our companies design system outlined is not something you should ever manually set but it rather a different button kind that communicates something deliberate

clotodex (commented):
initial quick skim - realizing i am reviewing it a bit from the design-system and DX side. maybe these comments help - maybe they dont :D

src/modules/updates.rs
MalpenZibo: The theme parameter is a pain in the ass in the entire codebase. Maybe we should simply create a global resource. Anyway, for now, I prefer to have the theme as the first parameter, like in other places in the codebase.

src/modules/tray.rs
MalpenZibo: The position button in the status bar here is more like a singleton. For now, I prefer to maintain this as an exception. Maybe we can re-work this in the future

src/modules/tempo.rs
MalpenZibo: I designed the button system with size, hierarchy, and kind as independent combinable axes. Outline is a kind that communicates "secondary boundary action", and combining it with any size is intentional. 

I think it's a matter of preference. I personally prefer it like this

PR #642, opened 2026-04-16
New module to view and toggle systemd user services

fdev31:
This module lists user units (with some filtering to keep the "interesting" ones) and allows seeing their status and toggling starting/stopping them.

> [!Note]
> I am not a Rust developer (but still a seasoned software engineer...) and heavily used an LLM to create this module. I am not really able to judge on the Rust itself, but tried to stick to the project's best practices.
>
> I tried to avoid "generic" changes, but one seemed required to control the cursor appearance in commit d98b530ff5f61ea8466a1150aca0f676cadfbfd8 *fix: delegate mouse_interaction in MenuWrapper to child content*

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -16,6 +16,7 @@ use crate::{
         tempo::Tempo,
         tray::TrayModule,
         updates::Updates,
+        user_services::UserServices,
         window_title::WindowTitle,
         workspaces::Workspaces,
     },
@@ -62,6 +63,7 @@ pub struct App {
     pub settings: Settings,
     pub media_player: MediaPlayer,
     pub notifications: Notifications,
+    pub user_services: UserServices,
     pub visible: bool,
 }
 
@@ -83,6 +85,7 @@ pub enum Message {
     Settings(modules::settings::Message),
     MediaPlayer(modules::media_player::Message),
     Notifications(modules::notifications::Message),
+    UserServices(modules::user_services::Message),
     OutputEvent(OutputEvent),
     CloseAllMenus,
     ResumeFromSleep,
@@ -136,6 +139,7 @@ impl App {
                     settings: Settings::new(config.settings),
                     notifications,
                     media_player: MediaPlayer::new(config.media_player),
+                    user_services: UserServices::new(),
                     visible: true,
                 },
                 task,
@@ -392,6 +396,10 @@ impl App {
                 modules::media_player::Action::None => Task::none(),
                 modules::media_player::Action::Command(task) => task.map(Message::MediaPlayer),
             },
+            Message::UserServices(msg) => match self.user_services.update(msg) {
+                modules::user_services::Action::None => Task::none(),
+                modules::user_services::Action::Command(task) => task.map(Message::UserServices),
+            },
             Message::CloseAllMenus => {
                 if self.outputs.menu_is_open() {
                     self.outputs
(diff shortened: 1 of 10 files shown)


MalpenZibo:
Ok, first of all, thanks!

Now, about the PR scope. This adds a module to show and operate on systemd services. From my point of view, this is a little out of scope for a status bar. I mean, usually for this kind of stuff, there are dedicated apps that provide this functionality (eg: [MissionControl](https://missioncenter.io/))

Why do you need this kind of functionality in ashell?

fdev31:
This is part of what I need as a control in a "shell", at least in a lightweight form:
I first thought (and wrote) a system tray application, but then found it overkill for what it does, it was more "boilerplate" than functionality.
So I thought it could be hosted in another process and the shell seemed "obvious" to me, many users even have weather display or garbage collection information in there, so the user services seemed natural to me.

MalpenZibo:
Ok, I still see this as a "task manager" feature. 

This is limited to the user services, where other apps can show you every kind of service. User service is something that doesn't change frequently, so personally, I don't see the need to have this stuff in a status bar.

That said, this PR does not add additional dependency, and it's "only" 738 lines of code. @clotodex and @romanstingler, what's your opinion about this? Personally, I'm quite skeptical. If we merge something here, we should also maintain and fix possible bugs. This will be a module that I will not use for sure, so it would be difficult for me to maintain it.

Thinking about this, it would be nice to have a "system" that lets Ashell use "external" modules, like a plugin system or something like that. Anyway, this is out of scope for now.

fdev31:
That's kind of a point: it's the user shell, it manages user things... unlike the system units. Thank you for considering it :)

romanstingler:
IMO, this shouldn't be in a shell. 
for people who want this, we might discuss what you need to add this as a custom module. 
We could also discuss about a plugin system, but here we def. don't want to be responsible/maintain external stuff.

And what kind of user systemd are you constantly using? 
I have written some system services with timers and stuff like this. 
Why don't you manage these in a term? Instead of having a subset here and 
others are managed elsewhere?

MalpenZibo:
Just to understand if we could improve the custom module to make this happen. What do you need to transform this feature into a custom module?

clotodex:
My first thought goes to custom module as well and making that more powerful

fdev31:
> IMO, this shouldn't be in a shell. for people who want this, we might discuss what you need to add this as a custom module. We could also discuss about a plugin system, but here we def. don't want to be responsible/maintain external stuff.
> 
> And what kind of user systemd are you constantly using? I have written some system services with timers and stuff like this. Why don't you manage these in a term? Instead of having a subset here and others are managed elsewhere?

Why don't you check the clock and weather in your terminal ? Also RAM and CPU usage, network status, volume control, etc.... everything can be done in the shell and I really dont' see the point of that comment, every feature ashell natively provides is available via the shell.
I'm happy to hear you start timers, and don't undertand how the type of units I'm using makes any point here, but to satisfy the curiosity:

- personal wiki
- personal desktop assistant (needs toggling to save RAM when needed)
- pyprland (desktop extensions)
- buckle (which I like to be able to toggle)
...actually many user related services, which is the purpose of the user units...


I'm closing this issue since it's getting a bit wild with weird questions applying to everything in ashell, I feel it's not worth the effort.

MalpenZibo:
@fdev31, I'm sad that you take the question and push back in this way. I would like to apologize if I express myself in a harsh way.

The main reason behind my pushback is that at the end I'm the one in charge of the codebase. Every module and every change adds complexity to the project. It's true that everything that we show in `ashell` is available from the terminal. If I have an "infinite" amount of time, I'll be more than happy to accept every kind of module inside the codebase to show every possible piece of information. Sadly, I don't have enough time to do that. I had to choose to add widely used things to cover as many use cases as possible with minimal effort.

I would like to have a real "plugin" system where users can write `ashell` extensions using `rust` directly, and what we have here. Sadly, it's a very complex thing to do, and the only alternative is the custom modules.

I'm sorry for what happened.

fdev31:
I understand, I'm also not very patient lately, it's all fine don't worry.

PR #643, opened 2026-04-16
window_title prevent redundant state updates in recalculate_value

romanstingler:


MalpenZibo:
I'm not sure that this will prevent a re-rentering. After an update there's a view call even if the state doesn't change. To really prevent the re-rentering we should probably avoid sending the Message from the subscription

PR #653, opened 2026-04-17
Feat/ipc socket

MalpenZibo:
Add socket functionality to send a message to the running instance.

For now only the toggle-visibility message is supported


romanstingler:
I had a quick look. Works fine, looks fine to me.

PR #655, opened 2026-04-18
POC - explore animations

MalpenZibo:
- Fixed an `iced_layershell` bug that prevents the animation redraw cycle
- Add `iced_anim` as dep (not 100% sure about this decision)

- Add a simple width animation on workspace buttons
- Add a (not so) simple animation to the window title

### DIFF
--- a/src/modules/workspaces.rs
+++ b/src/modules/workspaces.rs
@@ -11,6 +11,7 @@ use iced::{
     Element, Length, Subscription, SurfaceId, alignment,
     widget::{MouseArea, Row, button, container, text},
 };
+use iced_anim::{AnimationBuilder, transition::Easing};
 use itertools::Itertools;
 use std::collections::HashMap;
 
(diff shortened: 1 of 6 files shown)


clotodex:
Do i get this right that outside of a running animation this has essentially 0 overhead?
If so this is amazing and i would add a global option for turning off animations

MalpenZibo:
In reality I recently spent some times on this and I already have a bunch of branches that implements more or less all the "main animations".

I still need to polish things up.

Everything is opt-in for now because I would like to treat the animation as an experimental feature untill the official final animation release in iced.

From a performance point of view I use the official low level iced animation API + iced_anim that's an utility library build on top of that API. Obviously there's a performance overhead because during the animation the widget requires extra draw call but it's normal in every library.

The final result seems ok, I will create some working PR in the future.

src/modules/workspaces.rs
```diff
@@ -442,36 +443,61 @@ impl Workspaces {
...
-                                .style(theme.workspace_button_style(empty, color))
-                                .padding(if w.id < 0 {
+                            {
+                                let target_width = match (w.id < 0, &w.displayed) {
+                                    (true, _) => None,
+                                    (_, Displayed::Active) => Some(theme.space.xl),
+                                    (_, Displayed::Visible) => Some(theme.space.lg),
+                                    (_, Displayed::Hidden) => Some(theme.space.md),
+                                };
+                                let name = w.name.clone();
```
romanstingler, line 453: here we can replace the `.clone()` with `.as_str()`

`iced::widget::text` accepts `&str` directly and stores it as `Fragment::Borrowed`, which outlives the closure because its lifetime `'a` matches the `Element<'a>` returned by `view<'a>`

src/modules/workspaces.rs
```diff
@@ -442,36 +443,61 @@ impl Workspaces {
...
                                 } else {
                                     Message::ToggleSpecialWorkspace(w.id)
+                                };
+                                let height = theme.space.md;
+                                let font_size = theme.font_size.xs;
+
+                                Some(match target_width {
+                                    Some(tw) => AnimationBuilder::new(tw, move |w| {
+                                        button(
+                                            container(text(name.clone()).size(font_size))
```
romanstingler, line 474: and remove the `.clone()` here as well

MalpenZibo:
I think that I will deprecate this PR. I had a final plan based on this POC that I will push in the next few days. @romanstingler makes a lot of sense. I will do that

PR #657, opened 2026-04-18
Feat/osd audio, brightness, airplane control

MalpenZibo:


website/docs/configuration/main.md
```diff
@@ -151,3 +151,37 @@ Alternatively, you can still use a `SIGUSR1` signal:
...
+```bash
+ashell msg volume-up --no-osd
+```
+
+### OSD Configuration
+
+```toml
+[osd]
+enabled = true   # Disabled by default; set to true to enable the OSD overlay
+timeout = 1500   # Auto-hide delay in milliseconds
```
denqxotl, line 186: I would add width/height and position left/right/center

CodedNil:
Looks great, glad this is being added. Would be nice to have caps lock and next/previous track too, I could work on that in a followup PR unless you want to here.

website/docs/configuration/main.md
MalpenZibo: That's for a future PR. The plan is to have an 8-point grid (the screen center doesn't seem like a good place) with a spacing option

MalpenZibo:
> Looks great, glad this is being added. Would be nice to have caps lock and next/previous track too, I could work on that in a follow-up PR unless you want to here.

What do you expect from a next/prev/play-pause feature? A simple icon that shows the action or the name ok the track also?

For the caps-lock i need to understand if exist a reliable method to get the caps lock key state

CodedNil:
> What do you expect from a next/prev/play-pause feature? A simple icon that shows the action or the name ok the track also?
image
Something like this with SwayOSD where it shows the action icon plus next track

MalpenZibo:
> Something like this with SwayOSD where it shows the action icon plus next track

ok, I will probably implement this other stuff in future PRs

romanstingler:
for the base PR, I think it looks fine. Code is ok, it works and the basics are covered. 

I would say lets merge it

PR #660, opened 2026-04-20
Fix notifications daemon

### LINKED ISSUE #659: [Notifications] Daemon stop work on config update
When I update config file in ~/.config/ashell notifications daemon stop work. It doesn't handle and display notifications

denqxotl:
Closes #659

### DIFF
--- a/src/services/notifications/dbus.rs
+++ b/src/services/notifications/dbus.rs
@@ -2,7 +2,6 @@ use super::NotificationIcon;
 use log::{debug, info, warn};
 use std::collections::HashMap;
 use std::time::SystemTime;
-use tokio::sync::OnceCell;
 use tokio::sync::broadcast;
 use zbus::{
     Connection,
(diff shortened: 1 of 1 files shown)


src/services/notifications/dbus.rs
```diff
@@ -172,10 +180,13 @@ impl NotificationDaemon {
 
         let dbus_proxy = DBusProxy::new(&connection).await?;
         let flags = RequestNameFlags::AllowReplacement.into();
-        if dbus_proxy.request_name(NAME, flags).await? == RequestNameReply::InQueue {
```
MalpenZibo: Maybe we could avoid the `OnceCell` using `ReplaceExisting`. The Tray service as a similar solution could you check and try to understand if the `ReplaceExisting` resolve the issue?

denqxotl: sure

MalpenZibo (approved):
🚀

PR #665, opened 2026-04-21
Add two more configuration options for battery display

CodedNil:
Adds battery_hide_when_full which hides the battery indicator when it is full, this is useful imo for laptops which are plugged in almost all the time. In my view it is better at least for some people to only show information when relevant to the user.
Also adds peripheral_expanded_by_default which expands the peripheral battery display when opening the full settings menu by default, instead of having nothing visible by default. Since that menu is purely informative and the others are for engaging in actions it is beneficial for me to have all the information available in one click rather than two.

I appreciate these extra configs might be seen as wasteful bloat so don't mind if it isn't merged.


whynotofu:
Would it be possible for it to use:

> /sys/class/power_supply/BAT0/status

to determine if full, as I cap my battery charge level to 60%, since it is mostly plugged in.

CodedNil:
> Would it be possible for it to use:
> 
> > /sys/class/power_supply/BAT0/status
> 
> to determine if full, as I cap my battery charge level to 60%, since it is mostly plugged in.

That makes a lot of sense, that is sorted now and I added documentation for these configs.

MalpenZibo (approved):
👍

Issue #664, opened 2026-04-21
Not enough information about the packages necessary (battery info in my case) for the modules


VongoSanDi:
Hi,

I have a really minimal Arch configuration so I only install the packages I need. While configuring my topbar, i couldn't get the battery informations. I had to ask AI and check the sources to understand I was missing the upower package.

So I was wondering if it should be included in the Ashell doc or if it was a mistake on my side to not install upower by default ?

PS:
I also can't get the cpu temperature atm ='(

romanstingler:
Hey, you are right Upower is not mentioned in the docs. 
I will update them and see where we have missed some information. 

Yes, I know about the sensor situation. 
I am working on an auto detection heuristic, but this will take until the end of week. 
The issue is that the default sensor is not the one you have. 
https://malpenzibo.github.io/ashell/docs/configuration/modules/system_info#temperature
at the end of the page there is a config example, just use your sensor. 

You will need `lm_sensors` and run ` sensors` to find yours, but usually it is 
```
acpitz temp1 - ACPI thermal zone
coretemp Package id 0 - Intel CPU temperature
k10temp Tctl - AMD Ryzen CPU temperature
```

VongoSanDi:
> Hey, you are right Upower is not mentioned in the docs. I will update them and see where we have missed some information.
> 

Good to know.

> Yes, I know about the sensor situation. I am working on an auto detection heuristic, but this will take until the end of week. The issue is that the default sensor is not the one you have. https://malpenzibo.github.io/ashell/docs/configuration/modules/system_info#temperature at the end of the page there is a config example, just use your sensor.
> 
> You will need `lm_sensors` and run ` sensors` to find yours, but usually it is
> 
> ```
> acpitz temp1 - ACPI thermal zone
> coretemp Package id 0 - Intel CPU temperature
> k10temp Tctl - AMD Ryzen CPU temperature
> ```

You were right, I wasn't using the good sensor.
I did check with the ````sensors```` command and I had:
````
ucsi_source_psy_USBC000:001-isa-0000
Adapter: ISA adapter
in0:           5.00 V  (min =  +5.00 V, max =  +5.00 V)
curr1:         0.00 A  (max =  +0.00 A)

dell_smm-isa-00de
Adapter: ISA adapter
fan1:        3419 RPM  (min =    0 RPM, max = 5100 RPM)
fan2:        3407 RPM  (min =    0 RPM, max = 5100 RPM)
temp1:        +60.0°C  
[...]

PR #666, opened 2026-04-21
document optional runtime dependencies and package requirements

romanstingler:
related to https://github.com/MalpenZibo/ashell/issues/664

It might be a bit too much, just let me know, but I am not sure if people will check the whole docs
rather then looking at the specific module they have an issue with.


MalpenZibo (approved):
Seems a good addition! Nice work!!

Issue #335, opened 2025-11-29
Settings module icon not centralized in highlighted area


hecto600:
When hovering the highlighted area of the icon it is not centralized. See the image below:
<img width="65" height="46" alt="Image" src="https://github.com/user-attachments/assets/2797e283-834f-46f3-8070-0f0b4db230a3" />

MalpenZibo:
Are you using the `Solid` bar style?

hecto600:
> Are you using the `Solid` bar style?

yes

MalpenZibo:
Sorry, I forgot. Are you using the release version 0.6, or are you using the next version (following the main branch of the repository)

hecto600:
> Sorry, I forgot. Are you using the release version 0.6, or are you using the next version (following the main branch of the repository)

Im using 0.6 version

MalpenZibo:
Ok! Because there were some changes in the icons recently. Anyway, I'm not able to replicate your issue. 

Can I ask your ashell config file to try to replicate this issue?

hecto600:
# Note
Temperature(not showing at all)  and update (not sure how to show pacman and paru ) modules not working as expected. Maybe I'm configuring it wrongly, but I will open a proper issue later.
  
```
[modules]
left = ["Workspaces"]
center = ["Clock","calendar"]
right = ["Updates","SystemInfo","Settings","Tray"]


[updates]
check_cmd = "checkupdates; paru -Qua; pacman -Qu; "
update_cmd = 'alacritty -e bash -c "paru; echo Done - Press enter to exit; read" &'

[system_info]
indicators = [ "Cpu", "Memory", "Temperature" ]

[system_info.cpu]
warn_threshold = 60
alert_threshold = 80

[system_info.memory]
warn_threshold = 70
alert_threshold = 85

[system_info.disk]
warn_threshold = 40
alert_threshold = 60

[system_info.temperature]
warn_threshold = 30
alert_threshold = 40

[[CustomModule]]
name = "calendar"
icon = ""
command = "zenity --calendar --title=\"Calendar\""


[appearance]
style = "Solid"
scale_factor = 1.5

primary_color = "#7aa2f7"
success_color = "#9ece6a"
text_color = "#a9b1d6"

workspace_colors = [ "#7aa2f7", "#9ece6a" ]

[appearance.danger_color]
base = "#f7768e"
weak = "#e0af68"

[appearance.background_color]
base = "#1a1b26"
weak = "#24273a"
[...]

PR #668, opened 2026-04-21
add typed temperature sensor configuration with auto-detection

romanstingler:
Yeah basically based on 
https://github.com/MalpenZibo/ashell/issues/664
https://github.com/MalpenZibo/ashell/issues/335

these 2 issues where people didn't had the default sensor or the temp was not was was expected, 
I thought I would just start with some heuristic auto-detect for sensors. 

Basically what I did is adding some `sensor_type` 
```
"Cpu" (default) — CPU temperature (coretemp/k10temp)
"Gpu" — GPU temperature (amdgpu/nouveau/xe)
"Acpi" — ACPI thermal zone (acpitz)
"Nvme" — NVMe SSD temperature
```
and in the config you can set nothing, which falls back to `Cpu`
or you set a `sensor_type` or you can override that with the name of the sensor (like we had it before). 

```
[system_info.temperature]
warn_threshold = 60
alert_threshold = 80
sensor_type = "Cpu"  # "Cpu", "Gpu", "Acpi", or "Nvme"
# sensor = "k10temp Tctl"  # optional: override with specific sensor
format = "Celsius"
```

No clue if you want to leave that with a capital letter or lowercase. 

For me it was something that some people were struggling with,
and this underlines the "install and run without a ton of configuring" slogan.

**What I am missing is the `sensor` output of a nvidia-open and closed-source driver.** 

I am not sure, for now if you have a sensor manually configured and it fails (lets say you swap the disk between systems) then it falls back to auto-detect ( we can discuss that, but a warning will be sent to log). 

I get something like (for now I put those into INFO)
```
INFO [ashell::modules::system_info] Auto-detected temperature sensor for Nvme: nvme Composite CT1000T705SSD3
INFO [ashell::modules::system_info] Using temperature sensor: nvme Composite CT1000T705SSD3

INFO [ashell::modules::system_info] Auto-detected temperature sensor for Gpu: amdgpu edge
INFO [ashell::modules::system_info] Using temperature sensor: amdgpu edge

INFO [ashell::modules::system_info] Auto-detected temperature sensor for Cpu: k10temp Tctl
INFO [ashell::modules::system_info] Using temperature sensor: k10temp Tctl

INFO [ashell::modules::system_info] Auto-detected temperature sensor for Acpi: acpitz temp1
INFO [ashell::modules::system_info] Using temperature sensor: acpitz temp1
```

### DIFF
--- a/src/modules/system_info.rs
+++ b/src/modules/system_info.rs
@@ -4,7 +4,7 @@ use crate::{
     components::icons::{StaticIcon, icon},
     config::{
         CpuFormat, DiskFormat, MemoryFormat, SystemInfoIndicator, SystemInfoModuleConfig,
-        TemperatureFormat,
+        SystemInfoTemperature, TemperatureFormat, TemperatureSensorType,
     },
     theme::AshellTheme,
     utils,
@@ -15,6 +15,7 @@ use iced::{
     widget::{Column, Row, column, container, row, text},
 };
 use itertools::Itertools;
+use log::{info, warn};
 use std::time::{Duration, Instant};
 use sysinfo::{Components, Disks, Networks, System};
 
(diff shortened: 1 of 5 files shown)


website/docs/configuration/modules/system_info.md
```diff
@@ -111,8 +111,17 @@ The Temperature indicator displays the current temperature from the configured s
...
+The default sensor type is `Cpu`, which will look for `k10temp` (AMD) or `coretemp` (Intel) sensors.
+
+You can configure the sensor type with `sensor_type`:
+
+- `"Cpu"` (default) — CPU temperature (coretemp/k10temp)
+- `"Gpu"` — GPU temperature (amdgpu/nouveau)
+- `"Acpi"` — ACPI thermal zone (acpitz)
+- `"Nvme"` — NVMe SSD temperature
+
+Alternatively, you can specify a exact sensor name with the `sensor` option to override auto-detection:
```
MalpenZibo:
```suggestion
Alternatively, you can specify an exact sensor name with the `sensor` option to override auto-detection:
```

src/modules/system_info.rs
```diff
@@ -80,13 +81,79 @@ struct SystemInfoData {
...
+                sensor, temperature_config.sensor_type
+            );
+            Some(sensor.to_string())
+        } else {
+            warn!(
+                "Configured temperature sensor '{}' not found in component list",
+                sensor
+            );
+            None
+        }
```
MalpenZibo: It's ok to me but in the PR description you say "for now if you have a sensor manually configured and it fails (lets say you swap the disk between systems) then it falls back to auto-detect". 

Here if I understand correctly we return None and create a warn log but there's no fallback behavior.

src/modules/system_info.rs
```diff
@@ -135,16 +202,39 @@ fn get_system_info(
         ),
     };
 
-    let temperature_cel = sensor_index
-        .and_then(|i| components.get(i))
-        .and_then(|c| c.temperature().map(|t| t as i32))
-        .or_else(|| {
+    let temperature_cel = if let Some(sensor) = temperature_config.sensor.as_deref() {
```
MalpenZibo: here. If we specify a sensor with a random string then the fallback behavior doesn't work

src/modules/system_info.rs
```diff
@@ -135,16 +202,39 @@ fn get_system_info(
...
+            reading = cached_sensor_label.as_ref().and_then(|label| {
+                components
+                    .iter()
+                    .find(|c| c.label() == label)
+                    .and_then(|c| c.temperature().map(|t| t as i32))
+            });
+        }
+
+        reading
+    };
```
MalpenZibo: This whole section it's quite confusing and cause wrong behavior. Can we simplify the logics?

MalpenZibo (changes requested):
Very nice addition!! Just some logics to be refined

src/modules/system_info.rs
romanstingler: Yes, sorry for my unclear wording. 
Basically what I wanted to ask, if we should go for now (until people want it the other way) with the auto-detect as fallback. 

The user wouldn't see the wrongly configured sensor (if he doesn't check the logs) but it would just work.
That's where I was struggling to make a decision. 

But for now I will change that to auto-detect and then lets see.

romanstingler:
I have reworked the whole PR a bit. 
I think the code is now easier to read.

and I added the fallback if the sensor label is not found 

```
WARN [ashell::modules::system_info] Configured temperature sensor 'nvme Composite' not found, falling back to auto-detection for Cpu
```

MalpenZibo:
I reworked the config a little bit and resolved the merge conflict

PR #669, opened 2026-04-21
add option to select between Celsius and Fahrenheit in Tempo module

### LINKED ISSUE #663: Temperature Units
Hey, thanks for the great status bar! :)

Is it possible to control whether Celsius or Fahrenheit are used for the CPU temperature/weather temperature? I suspect there isn't an option for this right now, but double checking I didn't miss something in the docs.

romanstingler:
@MalpenZibo 
I was not sure,
I saw that `api.open-meteo.com` supports directly fahrenheit and celsius so I added that here. 
My issue is, that I don't use weather and when I already had it, I saw that there is also wind speed in km/h. 

The question, should I rename that to `imperial` and `metric` in the config? Or any other suggestions? 

image

closes https://github.com/MalpenZibo/ashell/issues/663

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -301,7 +301,7 @@ pub enum CpuFormat {
     Frequency,
 }
 
-#[derive(Clone, Debug, Deserialize, Default)]
+#[derive(Clone, Debug, Deserialize, Default, PartialEq, Eq)]
 pub enum TemperatureFormat {
     #[default]
     Celsius,
--- a/src/modules/tempo.rs
+++ b/src/modules/tempo.rs
@@ -5,7 +5,7 @@ use crate::{
         styled_button,
     },
     config::{TempoModuleConfig, WeatherIndicator, WeatherLocation},
-    i18n::chrono_locale,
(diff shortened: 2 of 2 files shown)


MalpenZibo:
So I plan to tackle the i18n situation globally for the entire app.
For this reason, the logic part to request a specific unit of measure will be needed.

About the config. 
In my mind, we should have a global `language` and `locale` option that should be read from the system or specified in the config.
On that basis, we should localize every module. 

Since we already have places (datetime format) where we can override a specific format, maybe we should also add that here.

I prefer for now to use a global locale config field (that we should implement, I will do it as soon as possible), and on that basis request the correct unit measure (in this case metric for everyone and imperial for us (I don't know)?)

Otherwise, you can add specific config here, and then I could pipe that config to the global locale config

romanstingler:
I have rebased the branch on top of your i18n branch and reworked the code to use units based on the UnitSystem.
image
image

MalpenZibo (approved):

MalpenZibo:
:rocket:

MasonRemaley:
Nice, thanks for the quick patch! :D

PR #672, opened 2026-04-22
refactor(theme): access AshellTheme via thread_local global

MalpenZibo:
I decided to drop the AshellTheme ref parameter drilling to a thread_local variable. 

We lost just a little-little-bit of performance for a huge benefit in terms of API ergonomics


MalpenZibo:
@clotodex @romanstingler The review is not needed, but just for visibility. From my point of view this refactor makes a lot of sense

PR #676, opened 2026-04-23
Feat/i18n foundations

MalpenZibo:


src/app.rs
```diff
@@ -40,6 +41,24 @@ use std::{collections::HashMap, f32::consts::PI, path::PathBuf};
...
+    // `region` setting. Log a warning each time config is resolved so the user
+    // sees it until they migrate.
+    let region_fallback = if config.region.is_none() {
+        config.tempo.locale.map(|l| l.to_string())
+    } else {
+        None
+    };
+    if region_fallback.is_some() {
+        warn!("`tempo.locale` is deprecated; move it to the top-level `region = \"...\"` setting");
+    }
```
MalpenZibo: completely deprecate?

src/config.rs
```diff
@@ -453,8 +457,10 @@ pub struct TempoModuleConfig {
     #[serde(default)]
     pub weather_location: Option<WeatherLocation>,
     pub weather_indicator: WeatherIndicator,
-    #[serde(deserialize_with = "deserialize_locale")]
-    pub locale: Locale,
+    /// Deprecated: use the top-level `region` field instead. Still honored if
+    /// set, with a warning logged at startup and on config reload.
+    #[serde(default, deserialize_with = "deserialize_locale")]
+    pub locale: Option<Locale>,
```
MalpenZibo: Maybe we should directly move on

romanstingler:
I just skimmed the code and I have seen we have some trimming at '@' (sr@latin). 
We have  to keep in mind there are some counties like Serbia where they have latin and cyrillic if you want to allow translations later.

src/app.rs
romanstingler: yes I would remove it

PR #679, opened 2026-04-23
Disk deduplication

### LINKED ISSUE #661: Option to specify which disks are displayed
I'm on nixos with impermanence which mounts persistent directories instead of the whole disk, this leads to a huge list of disk usages when most of them are duplicates. Not sure how the customisation could work but maybe an option to make it based on real disk partitions instead of what is mounted. In this screenshot there's only one disk with two partitions.

Image

CodedNil:
Fixes #661, adds a deduplicate option to system info for disks, which if true will only have one listed disk per partition instead of mount points.

image

### DIFF
--- a/website/docs/configuration/modules/system_info.md
+++ b/website/docs/configuration/modules/system_info.md
@@ -69,6 +69,17 @@ You can change the display format using the `format` option in `[system_info.dis
 - `"Percentage"` (default) — shows disk usage as a percentage (e.g., `54%`)
 - `"Fraction"` — shows used and total disk space in GB (e.g., `256.00/512.00 GB`)
 
+#### Deduplication
+
+When multiple mount points share the same underlying block device, the Disk indicator can show a separate entry for each mount point, which may be redundant.
+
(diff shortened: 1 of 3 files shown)


website/docs/configuration/modules/system_info.md
```diff
@@ -69,6 +69,17 @@ You can change the display format using the `format` option in `[system_info.dis
 - `"Percentage"` (default) — shows disk usage as a percentage (e.g., `54%`)
 - `"Fraction"` — shows used and total disk space in GB (e.g., `256.00/512.00 GB`)
 
+#### Deduplication
+
+When multiple mount points share the same underlying block device, the Disk indicator can show a separate entry for each mount point, which may be redundant.
+
+Setting `deduplicate = true` in `[system_info.disk]` collapses these into a single entry per device, labelled by the device name (e.g., `sda`) rather than the mount path. Defaults to `false`.
```
MalpenZibo: Maybe the default in this case could be `true`

MalpenZibo (commented):
Nice!!

website/docs/configuration/modules/system_info.md
CodedNil: Done thanks

PR #681, opened 2026-04-23
feat - Add experimental animation support

MalpenZibo:
First PR that introduces a global flag to enable/disable animations. Default = false, this feature is still experimental


romanstingler:
Animations are nice, but I don't like them in general on my machines :)

MalpenZibo:
With the flag disabled, nothing should change :smile:

PR #682, opened 2026-04-23
feat: add width animations for bar modules and workspace buttons

MalpenZibo:
Add two complementary animation approaches:

- AnimatedSize widget: auto-detects and smoothly animates content width changes. 

- iced_anim::AnimationBuilder: used in workspaces for explicit button width transitions between active/visible/hidden states.

### DIFF
--- a/src/components/animated_size.rs
+++ b/src/components/animated_size.rs
@@ -250,7 +250,7 @@ where
 {
     AnimatedSize {
         content: content.into(),
-        duration: Duration::from_millis(400),
+        duration: Duration::from_millis(100),
         easing: Easing::EaseOutCubic,
     }
 }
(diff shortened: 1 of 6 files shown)


src/components/animated_size.rs
```diff
@@ -0,0 +1,256 @@
...
+pub fn animated_size<'a, Message, Theme, Renderer>(
+    content: impl Into<Element<'a, Message, Theme, Renderer>>,
+) -> AnimatedSize<'a, Message, Theme, Renderer>
+where
+    Renderer: iced::core::Renderer,
+{
+    AnimatedSize {
+        content: content.into(),
+        duration: Duration::from_millis(400),
+        easing: Easing::EaseOutCubic,
```
romanstingler, line 254: not a comment, just a question. 
Is this intentionally different from 
`AnimationBuilder` `.animation(Easing::EASE.quick()) `

MalpenZibo: The two easing functions are from different crates. I still had to tweak the animation. Probably, this is not subtle enough. The EaseOutCubic should match the feel of EASE

PR #687, opened 2026-04-24
Calendar show days without leading zeros

### LINKED ISSUE #685: Calendar days with leading zero
I prefer cleaner look of calendar without leading zeros. It is very simple change, if there is no need to preserve old behavior. Should I just change it or would it need to be configurable ?

whynotofu:
Thank you!

Closes #685


MalpenZibo (approved):
🚀

PR #692, opened 2026-04-26
feat(settings): add `remove_hibernate_btn` support in power menu

Scott-Nx:
Add support for `remove_hibernate_btn` so the Hibernate action can be hidden from the power settings menu on devices where hibernation is not usable, such as systems using zram.

## Changes
- Added `remove_hibernate_btn` to settings config
- Passed the config through settings initialization and reload handling
- Hid the Hibernate button in `PowerSettings::menu()` when enabled
- Updated the docs in the website docs

## Notes
This helps avoid showing a Hibernate action that would not work on affected systems.

---

### Before (0.8.0)
ภาพ

### After
ภาพ

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -567,6 +567,7 @@ pub struct SettingsModuleConfig {
     pub vpn_more_cmd: Option<String>,
     pub bluetooth_more_cmd: Option<String>,
     pub remove_airplane_btn: bool,
+    pub remove_hibernate_btn: bool,
     pub remove_idle_btn: bool,
     pub indicators: Vec<SettingsIndicator>,
     #[serde(rename = "CustomButton")]
@@ -598,6 +599,7 @@ impl Default for SettingsModuleConfig {
             vpn_more_cmd: Default::default(),
             bluetooth_more_cmd: Default::default(),
             remove_airplane_btn: Default::default(),
+            remove_hibernate_btn: Default::default(),
             remove_idle_btn: Default::default(),
             indicators: vec![
                 SettingsIndicator::IdleInhibitor,
--- a/src/modules/settings/mod.rs
+++ b/src/modules/settings/mod.rs
@@ -184,6 +184,7 @@ impl Settings {
                 config.peripheral_indicators,
                 config.peripheral_battery_format,
                 config.peripheral_expanded_by_default,
+                config.remove_hibernate_btn,
             )),
             audio: AudioSettings::new(AudioSettingsConfig::new(
                 config.audio_sinks_more_cmd,
@@ -497,6 +498,7 @@ impl Settings {
                         config.peripheral_indicators,
                         config.peripheral_battery_format,
                         config.peripheral_expanded_by_default,
+                        config.remove_hibernate_btn,
                     )));
                 self.audio
                     .update(audio::Message::ConfigReloaded(AudioSettingsConfig::new(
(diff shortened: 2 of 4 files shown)


src/config.rs
```diff
@@ -567,6 +567,7 @@ pub struct SettingsModuleConfig {
     pub vpn_more_cmd: Option<String>,
     pub bluetooth_more_cmd: Option<String>,
     pub remove_airplane_btn: bool,
+    pub remove_hibernate_btn: bool,
```
MalpenZibo, line 560: Here, maybe we could simply make the `hibernate_cmd` Optional and hide the button for the `None` value.

The only problem is that right now we use `systemctl hibernate` as the default value, and I'm not sure if it's possible to use toml to specify None as a value. 

The only way is to make this feature opt-in, letting the user set the command and switch to None as a default value.

@clotodex @romanstingler, what do you think?

MalpenZibo (commented):
This is something that I needed 😄

src/config.rs
Scott-Nx: I followed the existing `remove_airplane_btn` pattern to keep the config consistent.

It ensures everything works **out of the box** with sensible defaults, without forcing users to set a command just to see the button.

It matches how we handle other buttons. Using a different logic just for hibernate might be confusing for users.

btw, keeping UI visibility (boolean) and the command (string) separate avoids **hidden logic** where an empty string dictates UI behavior.

So I think keeping them separate keeps the architecture cleaner, but let me know what you think!

MalpenZibo: Yep, but here the difference is that we already have a hibernate button config (the one for the command to use). So this case is more similar to the `wifi_more_cmd`, where we have an Option<String> that, if it's None, disables the button. The `remove_airplane_btn` is the only one where there isn't a command to specify.

The main difference is that the other buttons are opt-in, this one instead, is opt-out at the moment

Scott-Nx: I’ve tested setting`hibernate_cmd` as `Option<String>` on a separate branch: [Scott-Nx:feat/none-hibernate_cmd](https://github.com/Scott-Nx/AShell_Fork/tree/feat/none-hibernate_cmd).

Current Behavior:

- No `hibernate_cmd` in config: The button is Hidden (Default).
- `hibernate_cmd` is set (e.g., "" or a command): The button is Shown.

So now the Hibernate button is hide by default.

If this looks good, I’ll keep this version.

---

#### Default (Disable)
2569-04-27-125702_hyprshot

#### Enable
2569-04-27-125716_hyprshot

MalpenZibo: @sudo-Tiz implemented this feature.  From my point of view we can switch to a None default value hiding the button even if it's a breaking change

Scott-Nx: Just to clarify, I'm Scott-Nx (I think you might have tagged the wrong person earlier!).

Glad you agree with the Option<String> approach. Since this version is already tested and working on my feat/none-hibernate_cmd branch, I'll close this PR and open a new one from that branch to keep everything clean and consistent.

Thanks for the feedback, see you in the new PR!

MalpenZibo: Nono, the person is correct, the comment is misleading. I mean that Sudo-Tiz implemented the hibernate button functionality.

Anyway, let's proceed. I will place the breaking label to the PR

PR #693, opened 2026-04-26
fix(tray): drop malformed icon pixmaps to avoid panic

Lykathia:
As an example, the PX struct in fyne.io/systray can initialize to an empty pixmap, which was resulting in a core dump from iced.

Filter those entries before picking the icon.

Additionally, refactor the conversion so both the initial quert and the property change subscription applies the same validation.

### DIFF
--- a/src/services/tray/mod.rs
+++ b/src/services/tray/mod.rs
@@ -267,6 +267,55 @@ pub enum TrayIcon {
     Svg(svg::Handle),
 }
 
+fn pixmap_to_icon(icons: Vec<dbus::Icon>) -> Option<TrayIcon> {
+    icons
+        .into_iter()
+        .filter(|i| {
+            // SNI clients sometimes return entries with zero dimensions or a
+            // bytes payload that doesn't match width*height*4 (e.g. when only
+            // IconName is populated). Feeding those to iced's atlas uploader
+            // panics in the padding loop, so drop them up front.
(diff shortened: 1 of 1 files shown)


src/services/tray/mod.rs
```diff
@@ -267,6 +267,39 @@ pub enum TrayIcon {
...
+            // bytes payload that doesn't match width*height*4 (e.g. when only
+            // IconName is populated). Feeding those to iced's atlas uploader
+            // panics in the padding loop, so drop them up front.
+            if i.width <= 0 || i.height <= 0 {
+                return false;
+            }
+            let expected = (i.width as usize)
+                .checked_mul(i.height as usize)
+                .and_then(|v| v.checked_mul(4));
+            expected == Some(i.bytes.len())
```
MalpenZibo: Can we add a log here? Maybe a `debug!` one just to know when we filter some invalid icons?

MalpenZibo (commented):
Nice! just a minor comment

src/services/tray/mod.rs
Lykathia: Yeah, of course. Will update shortly.

Lykathia: Updated - added a log for invalid bytes as well.

MalpenZibo (approved):

PR #696, opened 2026-04-26
feat - i18n Updates module

MalpenZibo:


MalpenZibo:
@romanstingler @clotodex, if you are convinced by this approach, I will open the other PRs to complete the refactor of the rest of the codebase without explicit request reviews.

romanstingler:
Hopefully not too many strings :) 

fyi you can do something like that in fluent 
```
updates-available =
    { $count ->
        [one] { $count } Update available
        *[other] { $count } Updates available
    }
 ```
 
 yeah I guess the flat names are easier to handle than with attributes.
 
 even though, you still could add something like 
 
 ```
 
 updates =
    .up-to-date = Up to date ;)
    .available =
        { $count ->
            [one] { $count } Update available
           *[other] { $count } Updates available
        }
    .button-update = Update
    .button-check-now = Check now



 fn get_by_path<'a>(
    bundle: &'a fluent_bundle::FluentBundle,
    path: &str,
) -> Option<&'a fluent_bundle::FluentAttribute> {
    let (msg_id, attr) = path.split_once('.')?;
    bundle.get_message(msg_id)?.get_attribute(attr)
}

let attr = get_by_path(&bundle, "updates.available").unwrap();
```

and use dot notation. 

To be honest, I don't think this makes a big of a difference. 
Either way is fine by me.

clotodex:
nothing to add here and not a strong preference either way. I do like @romanstingler's feedback though. Also like the grouping for scalability. But very very weak preference here, since we could just as well just group them visually in the file and that could save some code. Yeah, no opinion here, go ahead with what you feel most comfortable with

MalpenZibo:
I fixed the plural form. For the grouping, for now, we shouldn't have too many strings, so I prefer to avoid any form of overcomplication. In the future, we will see.

PR #697, opened 2026-04-26
fix(tray): keep menu open when toggling a checkbox item

Lykathia:
Toggling a checkmark item dispatched the same MenuSelected message as a regular click, so the handler in app.rs closed all menus immediately.

This change routes checkbox toggles through a dedicated MenuToggled message and that runs the SNI command without closing the menu, matching how desktop tray menus typically behave.

ashell

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -414,6 +414,7 @@ impl App {
                         .close_all_menus(self.general_config.enable_esc_key),
                     task.map(Message::Tray),
                 ]),
+                modules::tray::Action::TrayMenuCommandKeepOpen(task) => task.map(Message::Tray),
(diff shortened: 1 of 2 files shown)


Lykathia:
As an aside/ using the toggler widget for checkbox I find a bit visually confusing. I've updated my own local copy of ashell to use checkmarks instead -- I think I'd prefer no checkbox for empty state as well compared to how I have it now.

Would there be interest in this UX change, if I clean it up?

image
image

MalpenZibo (approved):

MalpenZibo:
I found the ui change cleaner but a little bit less intuitive. 

Could you open an issue with the screenshot to discuss this?

PR #702, opened 2026-04-26
Make airplane mode and bluetooth icons match their state

whynotofu:
Matches icon with state:
qirplane_mode_off
airplane_mode_on
screenshot


MalpenZibo (approved):
👍

PR #704, opened 2026-04-27
feat(i18n): respect LC_MEASUREMENT for unit system

clotodex:
I ran ashell today and realized everything is in imperial metrics.
To respect the system setup, locale exposes LC_MEASUREMENT

ashell currently derives the unit system solely from LC_TIME (or the region config key), mapping en_US → Imperial and everything else → Metric. This conflates two independent glibc locale categories: users who want US date formatting but metric units (or vice versa) have no way to express that without lying about one of the two locales or manually configuring.

This change reads LC_MEASUREMENT (with the standard LC_ALL → LC_MEASUREMENT → LANG fallback chain) and uses a CLDR-derived allowlist of imperial regions (en_US, en_LR, my_MM). I am not an expert on that and this was an AI recommendation. The three options DO use imperial though. If someone has more insight, happy to adjust.

When LC_MEASUREMENT is unset, falls back to the existing LC_TIME-based derivation — so no behavior change for existing users.

### DIFF
--- a/src/i18n.rs
+++ b/src/i18n.rs
@@ -143,8 +143,9 @@ fn env_chain(config: Option<&str>, category_var: &str) -> Option<String> {
 }
 
 fn derive_units(c: Locale) -> UnitSystem {
+    // `en_LR` omitted: not a chrono Locale variant.
     match c {
-        Locale::en_US => UnitSystem::Imperial,
+        Locale::en_US | Locale::my_MM => UnitSystem::Imperial,
         _ => UnitSystem::Metric,
     }
 }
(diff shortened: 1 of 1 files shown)


MalpenZibo (commented):
Maybe let's fix the `derive_unit` to match the new logics
```
fn derive_units(c: Locale) -> UnitSystem {
     match c {
         Locale::en_US | Locale::en_LR | Locale::my_MM => UnitSystem::Imperial,
         _ => UnitSystem::Metric,
     }
  }
}
```

Maybe en_LR is not a valid chrono locale, in that case we can remove and use the fallback

MalpenZibo (approved):
👍

PR #706, opened 2026-04-28
feat(settings)!: optional hibernate_cmd with empty string handling

Scott-Nx:
Makes `hibernate_cmd` optional in the power settings module, improving flexibility for systems that do not support hibernation. 

Also normalizes empty and whitespace-only optional command values to `None`, so unset and blank command fields behave consistently.

## Changes

- Changed `hibernate_cmd` from `String` to `Option<String>` in both `SettingsModuleConfig` and `PowerSettingsConfig`, and updated the default to `None`.
- Guarded hibernate execution so it only runs when a command is configured.
- Hid the Hibernate button from the power menu when `hibernate_cmd` is not set.
- Added config deserialization normalization so empty or whitespace-only optional command strings are treated as `None`.

## Documentation

- Updated documentation to clarify that `hibernate_cmd` is optional, and the Hibernate button only appears when it is set to a non-empty value.
- Added notes that empty or whitespace-only optional command fields are treated as unset.
- Updated related configuration docs for optional command-style fields to reflect this behavior.

## Behavior

- **Opt-in by Default:** If `hibernate_cmd` is omitted in TOML, the Hibernate button is hidden.
- **Blank Values Are Unset:** If `hibernate_cmd = ""` or contains only whitespace, it is treated as `None`, so the button remains hidden.
- **Dynamic Visibility:** The button renders only when a non-empty command is configured.
- **Flexible Execution:** When defined, the power module executes the configured command on click.

## Additional Consistency Change

The empty-string normalization was applied consistently to other optional command-like config fields as well, including:

- `lock_cmd`
- `audio_sinks_more_cmd`
- `audio_sources_more_cmd`
- `wifi_more_cmd`
- `vpn_more_cmd`
- `bluetooth_more_cmd`
- custom button
- custom module

## Testing Checklist

- [x] **Default State:** Verified button is hidden when `hibernate_cmd` is missing.
- [x] **Explicit Command:** Verified button is shown and works with `"systemctl hibernate"`.
- [x] **Custom Command:** Verified button executes custom strings like `"echo test"`.
- [x] **Empty String:** Verified `hibernate_cmd = ""` hides the button.
- [x] **Whitespace String:** Verified whitespace-only values are treated as unset.
- [x] **Build Validation:** `cargo fmt` and `cargo check` pass.

### Case 1: `hibernate_cmd` not set

screenshot_2569-04-28_08-03-57

### Case 2: `hibernate_cmd = ""` Button remains hidden.

screenshot_2569-04-28_08-05-33

### Case 3: `hibernate_cmd = "systemctl hibernate"` Button is shown (test with notify-send).

screenshot_2569-04-28_08-06-02


Scott-Nx:
Continuing the work from #692 with a refined approach (hiding the button by default and handling empty strings).

MalpenZibo (approved):

Issue #673, opened 2026-04-22
Few more ipc, osds and configs


whynotofu:
These would be really nice!

```
ashell msg microphone-up
ashell msg microphone-down
ashell msg microphone-toggle-mute
ashell msg cycle-keyboard-backlight
ashell msg toggle-idle-inhibitor
```

And accompanying OSDs and configurable steps on volume, mic and screen brightness. (I use screen brightness step of 4%, but I assume there could be people with sensitive ears as my eyes are. max_step could still be 5%)

PR #708, opened 2026-04-28
feature: OSD on indicators RFC

whynotofu:
1. feature: add audio_step and brightness_step to config
2. feature: show OSD on volume and microphone indicator scrolling
3. feature: add sink/source mute on indicator right click if *_more_cmd not set.
4. feature: show OSD on brightness indicator scrolling

For feature No.1: There are three places where step is used:

1. Sliders on_scroll: Currently 5. In this PR I have fixed it to 1.
2. Indicators on_scroll: Currently 5. In this PR I have fixed it to 1.
3. IPC: In this PR it's read from config.

Part of #673

Sorry for being in the same PR but I just kept typing ...

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -13,7 +13,7 @@ use crate::{
         media_player::MediaPlayer,
         notifications::Notifications,
         privacy::Privacy,
-        settings::{self, Settings, audio},
+        settings::{self, EventSource, Settings, audio},
         system_info::SystemInfo,
         tempo::Tempo,
         tray::TrayModule,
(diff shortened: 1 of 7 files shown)


MalpenZibo:
Not sure about this. Three main things that I don't like: 
- not sure if make sense to have a step config on scroll event
- I prefer to move the "decoding" logic during the config deserialization process (so the 1 to 5 check)
- Not sure about the range 1 to 5

Personally, I would like to have a step parameter on the msg first. For now, I don't think it's necessary to have a step config to use on the scroll value.

whynotofu:
What range would you suggest ? 1 - 10 ?

MalpenZibo:
ok, so maybe 1 to 10 is better. 

My only doubt is, if you use 1, then do we have a real use case for these changes? I mean, I don't use the scrolling behavior, you use 1, that's what we already have, so there's actually someone who will use this feature?

whynotofu:
Regarding IPC value vs config: In ashell/config.toml it is less code and it does not need to go through from/to/from string and validation on each invocation. So from efficiency standpoint it is more efficient.

Point number 3 resolved.

whynotofu:
So to clarify: There are three places when step is used:

1. Sliders on_scroll: Currently 5, it should be 1 fixed.
2. Indicators on_scroll: Currently 5, I think it should configurable. it would be only 8 extra one word lines to add it. And two more words in docs.
3. IPC: this PR.

I think best place for it is in ashell/config.toml. IPC I do not know why bother, it makes sense for linux commands that interact directly with kernel and audio server. But for shell it really should be in shell.

Issue #327, opened 2025-11-23
Log grew to 15G filling up /tmp


clotodex:
I switched from hyprland to niri to try it out - and the workspace module freaked out :D
Within the first few seconds it filled up my tmpfs.

While this is an issue I can resolve by not using that module, I find it suprising that the logger can explode like this.

Does anyone seeing this know of a "deduplicate" or logcompression best practice for rust?

Otherwise amateur googling led me to flexi_loggers logrotation that can be configured to happen on size and clean up with compression.
https://docs.rs/flexi_logger/latest/flexi_logger/code_examples/index.html#rotate-the-log-file

I would set it up as 
- max size (e.g. 1 gig)
- rotate it automatically with compression (optionally keep a maximum number)

Happy to push a PR for this @MalpenZibo if you agree

The deduplication issue or "shutting up" a module that errors out is still an open concern though

clotodex:
While not resolving the logging issue, https://github.com/MalpenZibo/ashell/pull/328 does put an end to the hyprland error spam for me :)

MalpenZibo:
I saw the PR with niri support! very appreciated! I had an idea to resolve the log bug, but maybe it's not needed after #328

clotodex:
My general gut feeling is that an "erroring module" might need some first class support - showing an error in the bar, being unable to "load". That feels like the cleanest (and hardest, and most maintenance effort, and complex).
The logrotation is easy to build if you think this is worth it (can submit a pr).
But as you said, with #328 I (personally) am not running into the issue anymore

chazfg:
I'm on hyprland and I get this behavior if pipewire isn't started and I start ashell. My log fills with `ERROR [ashell::services::privacy] Pipewire listener exited`.

chazfg:
It seems like my logs are filling from here: 

https://github.com/MalpenZibo/ashell/blob/a8549c9e37ccb9e5479e1b2b09561a5c648f77f7/src/services/privacy.rs#L189

Specifically:
```
                select! {
                    value = pipewire.recv().fuse() => {
                        match value {
                            Some(event) => {
                                let _ = output.send(ServiceEvent::Update(event)).await;
                            }
                            None => {
                                error!("Pipewire listener exited");
                            }
                        }
                    },

```

[...]

PR #709, opened 2026-04-29
fix(logging): added file size of 10MB as an additional criterion for …

MustafaAamir:
…log file 

`/tmp/` reaches capacity due to insufficient criterion for log file rotation. This can be observed by running `watch -n2 'du -sh /tmp/ashell/*.log 2>/dev/null | sort -rh | head -5'` with the default config.


clotodex:
#327 as reference

MalpenZibo (approved):
👍

MustafaAamir:
Should the constant TMP_FILE_SIZE be configurable?

MalpenZibo:
I think that for now it is not necessary

MustafaAamir:
#327 mentions a 1 GiB file size -- which I think is a tad too much.

PR #722, opened 2026-05-01
Replace generic `unwrap()` with `expect()` in network backends

romanstingler:
And providing meaningful error messages for hardcoded D-Bus object paths.


MalpenZibo (approved):
👌

PR #723, opened 2026-05-01
fix: replace unwrap on connection id lookup with safe chaining

romanstingler:
The `find_connection` function used `s[connection].get(id).unwrap()` which panics if either the 'connection' section or 'id' key is missing. Replaced with `s.get().and_then()` pattern consistent with existing code. Connections without an id are skipped rather than crashing.


MalpenZibo (approved):
👌

PR #728, opened 2026-05-04
russian language support and clipboard history module

JS2006:
I added russian language support and clipboard history module


romanstingler:
Clipboard module has been removed in 0.8.0. 
you can build it with the custom module.

MalpenZibo:
Two things:
- Thanks for the russian language translation, but can you open a dedicated PR for that? Let's try not to mix things (translation + an entire new module)
- For the clipboard module, we removed a simple version in favor of the usage of a custom module. This implementation is way more complex and feature-rich. We could also evaluate this addition, but first, please open an issue, and let's have a discussion about that. I notice that the PR is probably AI-written, it's not a problem, but as stated in the README `You are responsible for the code you submit`. This PR has several problems; the most obvious is the `eprintln` usage instead of the default log system.

So, to proceed the best way is to open a dedicated PR for the translation, then open an issue to discuss about the clipboard module and eventually how to implement it, then we can proceed with a PR.

PR #729, opened 2026-05-05
feat(tray): left/right click behavior

### LINKED ISSUE #492: Tray Click Behaviour
When I left-click tray icons, it shows the menu associated with the application.
Right-click and middle-click do nothing.

From reading the documentation and using other status bars, my expectation would be that left-click either opens/closes the application, or does some action that is associated with this program.
For example the nextcloud icon would show/hide the sync overview, or the fcitx-5 icon would change the keyboard layout.

I'm using the latest git version as a flake on nixos with niri.

alexandre-abrioux:
Closes https://github.com/MalpenZibo/ashell/issues/492

## Change description

This PR adds support for configuring the left- and right-click behavior for tray items. We can now either:
- show the context menu (like before)
- open the application (new)

## To consider (deprecated section, [see conclusion](https://github.com/MalpenZibo/ashell/pull/729#issuecomment-4423174918))

⚠️ I took the liberty of changing the default for left-clicking tray items. It now opens the application instead of opening the context menu. The context menu is now bound to right-click by default.

I felt it was an OK change to make because:
- Left-clicking workspace icons already switches the workspace, and so switching to an app with left-clicking tray icons feels similar.
- Context menus often open on right-click.
- I've introduced new config properties so users can specify their preference.
- The documentation includes an example on how to switch the buttons.

However, I understand this is mostly subjective and likely a matter of personal preference. I'm happy to revert that part. Please let me know if you consider this unwanted and would like to revert the default left-click behavior to its previous state.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -439,11 +439,9 @@ impl Default for NotificationsModuleConfig {
     }
 }
 
-#[derive(Deserialize, Default, Clone, Debug, PartialEq, Eq)]
+#[derive(Deserialize, Clone, Debug, PartialEq, Eq)]
 #[serde(rename_all = "snake_case")]
 pub enum TrayClickAction {
-    #[default]
-    None,
     Open,
     Menu,
 }
@@ -452,7 +450,7 @@ pub enum TrayClickAction {
 #[serde(default)]
 pub struct TrayModuleConfig {
     pub blocklist: Vec<RegexCfg>,
-    pub right_click: TrayClickAction,
+    pub right_click: Option<TrayClickAction>,
 }
 
 #[derive(Deserialize, Clone, Debug)]
--- a/src/config.rs
+++ b/src/config.rs
@@ -439,9 +439,20 @@ impl Default for NotificationsModuleConfig {
     }
 }
 
+#[derive(Deserialize, Default, Clone, Debug, PartialEq, Eq)]
+#[serde(rename_all = "snake_case")]
+pub enum TrayClickAction {
+    #[default]
+    None,
+    Open,
+    Menu,
+}
+
 #[derive(Deserialize, Clone, Debug, Default)]
+#[serde(default)]
 pub struct TrayModuleConfig {
     pub blocklist: Vec<RegexCfg>,
+    pub right_click: TrayClickAction,
 }
 
 #[derive(Deserialize, Clone, Debug)]
--- a/src/services/tray/mod.rs
+++ b/src/services/tray/mod.rs
@@ -690,6 +690,7 @@ impl ReadOnlyService for TrayService {
 #[derive(Debug, Clone)]
 pub enum TrayCommand {
     MenuSelected(String, i32),
+    Activate(String),
 }
 
 impl Service for TrayService {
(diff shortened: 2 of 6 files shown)


MalpenZibo:
I'm not sure about this change. 

The default config behavior should remain the old one. Left-clicking the icon opens the menu. The problem with opening the app is more about the scope of the module. This is a tray module and not a Docker bar. I do not expect to be able to open an application from there directly, maybe only from some tray menu option if present.

Anyway, it's a minimal change, so we could also include it. Not sure about the configuration, in this way it's possible to set both clicks on the same action, which doesn't make much sense. We could place a simple bool to invert the action.
- field not present or `false` -> left click open the menu, right open the app
- field `true` -> left click open the app, right open the menu

src/services/tray/mod.rs
```diff
@@ -722,6 +723,23 @@ impl Service for TrayService {
                     Task::none()
                 }
             }
+            TrayCommand::Activate(name) => {
```
clotodex: does this mean the user gets no feedback if the dbus does not expose an activate item?
While I dont have a quick fix, I feel like that will make it feel "broken" for some tray elements.
Any ideas? Otherwise we can keep it as is for now

alexandre-abrioux:
@MalpenZibo Thanks for the review :pray:

To keep the exact behavior we have today, what do you think about extending the enum with a third values: `none`. The default config would be:
```toml
[tray]
left_click = "menu"
right_click = "none"
```

Users could configure the following if they want:
 ```toml
[tray]
left_click = "open"
right_click = "menu"
```

**Comment by MalpenZibo** [comment]
I don't like to have an explicit non-value.
Let's add only a `right_click` field as Option<ActionEnumName>.
- None -> no action on right click -> default
- Open -> open the app, and automatically, the left click opens the menu
- Menu -> open the menu, and automatically, the left click opens the app

**Comment by alexandre-abrioux** [comment]
@MalpenZibo Sounds good! I've updated the PR, feel free to review 🙂

File: src/config.rs [hunk 2]
```diff
@@ -439,9 +439,20 @@ impl Default for NotificationsModuleConfig {
...
+    None,
+    Open,
+    Menu,
+}
+
 #[derive(Deserialize, Clone, Debug, Default)]
+#[serde(default)]
 pub struct TrayModuleConfig {
     pub blocklist: Vec<RegexCfg>,
+    pub right_click: TrayClickAction,
```
MalpenZibo:
```suggestion
    pub right_click: Option<TrayClickAction>,
```

src/config.rs
```diff
@@ -439,9 +439,20 @@ impl Default for NotificationsModuleConfig {
...
 
+#[derive(Deserialize, Default, Clone, Debug, PartialEq, Eq)]
+#[serde(rename_all = "snake_case")]
+pub enum TrayClickAction {
+    #[default]
+    None,
+    Open,
+    Menu,
+}
+
```
MalpenZibo, line 448:
```suggestion
pub enum TrayClickAction {
    Open,
    Menu,
}

```

src/config.rs
alexandre-abrioux: Addressed in https://github.com/MalpenZibo/ashell/pull/729/changes/2844f38b2030e1ec0fe0afd963869311ac274435

src/config.rs
alexandre-abrioux: Addressed in https://github.com/MalpenZibo/ashell/pull/729/changes/2844f38b2030e1ec0fe0afd963869311ac274435

alexandre-abrioux:
@MalpenZibo, thanks for the review. I understand you advised removing the `None` variant from the enum and making the absence of a specific config result in the default behavior. I've made that change and updated the documentation accordingly 🙏

MalpenZibo (approved):
👌

PR #731, opened 2026-05-06
Fix Tray: prevent icon duplication

SimoneFelici:
Hi,
this PR fixes my issue: https://github.com/MalpenZibo/ashell/issues/593
feel free to ask some question about the issue,
I've tested this fix with various tray icons on my system and everything works.

### DIFF
--- a/src/services/tray/dbus.rs
+++ b/src/services/tray/dbus.rs
@@ -126,11 +126,7 @@ impl StatusNotifierWatcher {
                 let mut watcher = interface.get_mut().await;
                 let emitter = SignalEmitter::new(conn, OBJECT_PATH).unwrap();
                 watcher
-                    .register_status_notifier_item_manual(
-                        name_str,
-                        sender.into_inner(),
-                        &emitter,
-                    )
+                    .register_status_notifier_item_manual(name_str, sender.into_inner(), &emitter)
                     .await;
             }
         }
--- a/src/services/tray/dbus.rs
+++ b/src/services/tray/dbus.rs
@@ -126,7 +126,11 @@ impl StatusNotifierWatcher {
                 let mut watcher = interface.get_mut().await;
                 let emitter = SignalEmitter::new(conn, OBJECT_PATH).unwrap();
                 watcher
-                    .register_status_notifier_item_manual(name_str, sender.into_inner(), &emitter)
+                    .register_status_notifier_item_manual(
+                        "/StatusNotifierItem",
+                        sender.into_inner(),
+                        &emitter,
+                    )
                     .await;
             }
         }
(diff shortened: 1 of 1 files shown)


romanstingler:
thanks for the contribution
please fix the formatting
just run `make check` which runs 

```
❯ make check
cargo fmt
cargo check
   Compiling ashell v0.8.0 (/srv/opensource/ashell)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1.92s
cargo clippy -- -D warnings
    Checking ashell v0.8.0 (/srv/opensource/ashell)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 2.35s
```
@clotodex are you using tray apps lately ??

SimoneFelici:
> thanks for the contribution please fix the formatting just run `make check` which runs

Ops, sorry.
I've run make check.

src/services/tray/dbus.rs
MalpenZibo, line 156: Let's add this check
```suggestion
) {
    if self.items.iter().any(|(s, _)| s == &sender) {
        return;
    }

    let service = if service.starts_with('/') {
        format!("{sender}{service}")
    } else {
        service.to_string()
    };
```

MalpenZibo: Let's remove this check
```suggestion
```

src/services/tray/dbus.rs
```diff
@@ -126,11 +126,7 @@ impl StatusNotifierWatcher {
                 let mut watcher = interface.get_mut().await;
                 let emitter = SignalEmitter::new(conn, OBJECT_PATH).unwrap();
                 watcher
-                    .register_status_notifier_item_manual(
-                        "/StatusNotifierItem",
-                        sender.into_inner(),
-                        &emitter,
-                    )
+                    .register_status_notifier_item_manual(name_str, sender.into_inner(), &emitter)
```
MalpenZibo: Instead of this change, can you try modifying the check in the `register_status_notifier_item_manual`?

MalpenZibo (commented):
Can you try this other solution? I suspect that we're moving the bug instead of resolving it.

SimoneFelici:
> Can you try this other solution? I suspect that we're moving the bug instead of resolving it.

Thank you, this also work,
Should I commit this newer version?

MalpenZibo:
Yes, if it works, I prefer my approach

SimoneFelici:
> Yes, if it works, I prefer my approach

https://github.com/MalpenZibo/ashell/pull/731/changes/74e274a66e611d1cf72af4d4f771477fa56a5f4a

Feel free to merge

MalpenZibo (approved):

Issue #727, opened 2026-05-04
[Bug] Battery info is not shown in Settings module


ammarasyad:
<img width="626" height="68" alt="Image" src="https://github.com/user-attachments/assets/a0335a47-c95c-43c1-9d00-95d95d6cb440" />
Currently on latest ashell version (0.8.0) on CachyOS with Hyprland. Removing and re-adding the Settings module with ashell open crashes ashell (launched from my terminal) with the message:

```fish: Job 1, 'ashell &' terminated by signal SIGSEGV (Address boundary error)```

I apologize if this was reported beforehand.

Logs doesn't show anything off:
```
INFO [ashell::config] Decoding config file "/home/user/.config/ashell/config.toml"
INFO [ashell::config] Config file loaded successfully
WARN [ashell::modules::clock] Clock module is deprecated and will be removed in a future release. Please migrate to the Tempo module.
WARN [wgpu_hal::gles::adapter] Detected skylake derivative running on mesa i915. Clears to srgb textures will use manual shader clears.
WARN [wgpu_hal::gles::egl] Re-initializing Gles context due to Wayland window
```

config.yaml:
```
[modules]
left = [ "Workspaces", "SystemInfo", "Power Usage", "Tray" ]
center = [ "WindowTitle" ]
right = [ "Sway", [ "Tempo", "Privacy", "Settings" ] ]

[[CustomModule]]
name = "Sway"
type = "Button"
icon = "" # default icon
icons.'dnd.*' = "" # icon if regex of "alt" is true, multiple such options can be given
alert = ".*notification" # to show a red dot next to it
listen_cmd = "swaync-client -swb"
command = "swaync-client -t -sw"

[[CustomModule]]
name = "Power Usage"
type = "Text"
listen_cmd = "sh -c 'while true; do cat /tmp/power_usage; sleep 1; done'"

[settings]
lock_cmd = "hyprlock &"

[system_info]
indicators = [ "Cpu", "Memory", "Temperature", "DownloadSpeed", "UploadSpeed" ]

[system_info.temperature]
sensor = "coretemp Package id 0"

[window_title]
truncate_title_after_length=75

[appearance]
opacity = 0.4

primary_color = "#7aa2f7"
success_color = "#9ece6a"
text_color = "#a9b1d6"

workspace_colors = [ "#7aa2f7", "#9ece6a" ]

[appearance.menu]
opacity = 0.7
backdrop = 0.3

[appearance.danger_color]
base = "#f7768e"
[...]

PR #732, opened 2026-05-06
Fix double-subscription crash when Settings module is enabled

romanstingler:
There was a bug in `src/app.rs` in the `subscription()` method. 
When Settings was in the module list, the code would subscribe to Settings' underlying D-Bus services (UPower, PulseAudio, NetworkManager) twice: 
once through `self.modules_subscriptions()`, and again explicitly at the end of `Subscription::batch()` for OSD support. 
This double-subscription could corrupt D-Bus proxy state and cause segmentation faults. The fix checks if Settings is in any module list and only adds the explicit subscription when it's not present, eliminating the duplicate while maintaining OSD functionality.

Related to #727.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -757,6 +757,25 @@ impl App {
     }
 
     pub fn subscription(&self) -> Subscription<Message> {
+        use crate::config::{ModuleDef, ModuleName};
+
+        let mut all_modules = self
+            .general_config
+            .modules
+            .left
+            .iter()
+            .chain(&self.general_config.modules.center)
+            .chain(&self.general_config.modules.right);
+
+        let settings_in_module_list = all_modules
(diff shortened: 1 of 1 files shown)


MalpenZibo (changes requested):
We have some issues here:
- we had to check everytime in the subscribe fn (minor issue)
- if the settings module is inside a ModuleDef::Group, this logic doesn't work

I think that we can use the osd enabled flag to understand if we need the root setting subscription, and in the `get_module_subscription`, using the OSD-enabled field, disable the setting subscription.

In this way, if the OSD service is enabled, we always use the settings subscription and never attach the service subscription in the `get_module_subscription`. Otherwise, we do not use the settings subscription in the root, but only inside the `get_module_subscription`

MalpenZibo:
Is this PR really necessary? The original issue seems resolved, and in theory, iced shouldn't create two subscriptions if the id is the same

Issue #324, opened 2025-11-22
More mouse events support for CustomModule


abrushworkriot:
I just decided to check out what's new in the Hyprland ecosystem and saw this project, decided to test it out and found out it's much more elegant and requires less tinkering than my ugly waybar setup. Great job! I'd love to see this project grow.

Currently I was able to more or less replicate all functionality (and more!) of my very basic waybar setup, but I've been used to the volume module in waybar, which turns volume up/down when you scroll over it with your mouse. It would be very easy to replicate if more mouse events were supported by custom modules (predefined modules as well) like right/left/middle mouse click, scroll up/down etc. I think it shouldn't be too hard, but I'm not sure if I'll be able to implement it myself, having very limited Rust experience. I may try anyway.

Additionally, it would be great to be able to customise the label (or icon) depending on the output of some command, so we can e.g. poll for current volume/brightness etc. and display it on the bar.

So yeah, making the custom modules more customisable would be great in general, even with opening custom menus or such, but complexity can be increased ad absurdum and that's probably not the point, but I think more mouse actions would be reasonable for start.

clotodex:
hey :)
To your first point - happy to guide / review.
Check out the custom_module.rs code to check out how things are consumed and triggered and then work yourself backwards. The entire bar is written in a message-passing style.
Reading some entry level docs on using iced-rs also helped me.
Don't worry about your rust experience. Most patterns have been already implemented in the code-base.
If you get something working in a hacky way that is great - through the PR we can iterate towards something mergable.
I did not dive super deep into this - but to achieve what you want you might need to use something like https://docs.rs/iced/latest/iced/widget/struct.MouseArea.html#method.on_scroll
You will likely need to change modules/mod.rs (search for button)
Use iced to capture the mouse events - create new messages that will be passed - collect those messages in the custom module and do something with it :)

To the custom modules - dynamic icon changes ARE supported (in a rudimentary way)
```[[CustomModule]]
name = "CustomNotifications"
icon = "<your_default_icon>"
command = "swaync-client -t -sw"
listen_cmd = "swaync-client -swb"
icons.'dnd.*' = "<your_icon_if_dnd_in_output>"
alert = ".*notification" # adding a badge if notification in output
``` 
You can have multiple regex to configure which icon to show
this follows the waybar format - check out the docs: https://malpenzibo.github.io/ashell/docs/configuration/modules/custom_module

[...]

PR #735, opened 2026-05-06
Add configurable mouse event handlers to CustomModule

romanstingler:
This PR adds more ways to interact with custom modules in ashell. 
You can now set different commands for right-click, middle-click, and scrolling up or down directly on the module button itself. 

related to https://github.com/MalpenZibo/ashell/issues/324 even if the solution for this issue was a different one.

+ fixed/updated a bit of the docs (before release we will have to go through them again and check)

### DIFF
--- a/docs/src/modules/module-registry.md
+++ b/docs/src/modules/module-registry.md
@@ -31,30 +31,41 @@ The return type is `Option<(Element, Option<OnModulePress>)>`:
 
 ## OnModulePress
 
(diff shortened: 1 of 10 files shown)


MalpenZibo (approved):
Nice! We just have to resolve the conflict

romanstingler:
I will have some time in the evening to go through everything

romanstingler:
@MalpenZibo rebased

PR #738, opened 2026-05-07
feature: add audio_step and brightness_step to config

whynotofu:
Changes:
1. Sliders `on_scroll` changed from 5 to 1.
2. Indicators on_`scroll` changed from 5 to 1.
3. IPC (audio/brightness) `step` is read from config.

### DIFF
--- a/src/modules/settings/brightness.rs
+++ b/src/modules/settings/brightness.rs
@@ -19,7 +19,7 @@ pub enum Message {
     Event(ServiceEvent<BrightnessService>),
     Changed(remote_value::Message<u32>),
     MenuOpened,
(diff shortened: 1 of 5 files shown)


src/modules/settings/brightness.rs
```diff
@@ -44,15 +59,11 @@ impl BrightnessSettings {
         self.service.as_ref().map(|s| (s.current.value(), s.max))
     }
 
-    fn step(max: u32) -> u32 {
-        (5 * max / 100).max(1)
-    }
```
MalpenZibo, line 49: We can maintain this function by adding a `&self` and replacing the `5` with `self.config.step`

PR #740, opened 2026-05-07
workspaces: make button width scale to fit names

LukeCarrier:
I bind a keyboard shortcut <kbd>Super</kbd> + <kbd>R</kbd> to set the name of each space, and I generally set the names to those of the projects I'm working on. In current ashell releases this results in overlapping text as soon as names are over 1-2 characters in length.

Make the workspace buttons width shrink to fit the workspace names, and differentiate between them with colours rather than borders.

image

### DIFF
--- a/src/modules/workspaces.rs
+++ b/src/modules/workspaces.rs
@@ -444,27 +444,14 @@ impl Workspaces {
                                         .align_y(alignment::Vertical::Center),
                                 )
                                 .style(theme.workspace_button_style(empty, color))
-                                .padding(if w.id < 0 {
-                                    match w.displayed {
-                                        Displayed::Active => [0.0, theme.space.md],
-                                        Displayed::Visible => [0.0, theme.space.sm],
-                                        Displayed::Hidden => [0.0, theme.space.xs],
-                                    }
-                                } else {
-                                    [0.0, 0.0]
-                                })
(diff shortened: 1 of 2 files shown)


MalpenZibo (commented):
I recently added the workspace animation that works with the fixed width of the button so I need to rethink the solution to address this bug

LukeCarrier:
Would it make sense to use colour in place of item width to differentiate between active and inactive workspaces, and those which have contents and those which don't? That's what my second commit was doing here, but I wasn't really sure if you'd want changes to the theme.

No rush, I am happy to run this patch in the meantime. ashell is an awesome piece of software and is a major upgrade over my janky Waybar configuration -- thank you for maintaining it!

MalpenZibo:
Let's try to resolve the issue before the next release.

I will try to implement in a new PR a way to correctly display workspaces with a name, maintaining the new workspace width animation. Then we can discuss other changes

MalpenZibo:
fixed here https://github.com/MalpenZibo/ashell/pull/790

MalpenZibo:
For this, could you open a dedicated issue?

> Would it make sense to use colour in place of item width to differentiate between active and inactive workspaces, and those which have contents and those which don't? That's what my second commit was doing here, but I wasn't really sure if you'd want changes to the theme.

PR #741, opened 2026-05-08
feature: OSD for indicators

whynotofu:
1. Adds OSD for volume, microphone and brightness indicator on_scroll.
2. Adds mute on right click on volume and microphone indicators.
3. Refactors OSD so mute is not used for everything.
4. Fix for bug in AudioService where it misrepresents actual volume as 0 when muted.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -13,7 +13,7 @@ use crate::{
         media_player::MediaPlayer,
         notifications::Notifications,
         privacy::Privacy,
-        settings::{self, Settings, audio},
+        settings::{self, Settings},
         system_info::SystemInfo,
         tempo::Tempo,
         tray::TrayModule,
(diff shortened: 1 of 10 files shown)


MalpenZibo:
Wait!! Too many things in a single PR. 

> Adds OSD for volume, microphone, and brightness indicator on_scroll.

Why? I don't want to show the OSD during a user interaction with ashell. Doesn't make much sense IMO.

> Adds mute on right click on volume and microphone indicators.

In this way, the user cannot decide not to have an action on the right click

There are changes mixed with the audio/brightness step changes.

Please open dedicated PRs and discuss the changes before the implementation

whynotofu:
> There are changes mixed with the audio/brightness step changes.

Once audio/brightness step changes are merged I will rebase it.

> Why? I don't want to show the OSD during a user interaction with ashell. Doesn't make much sense IMO.

Try it, you gonna like!

>  In this way, the user cannot decide not to have an action on the right click

They can **not** right click. :smile: Really powerful! :smile:

Issue #317, opened 2025-11-17
Use ext-workspace-v1 protocol


jazzpi:
There is a [wayland-protocol for handling workspaces](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/blob/main/staging/ext-workspace/ext-workspace-v1.xml). This is [supported by Hyprland](https://github.com/hyprwm/Hyprland/blob/4695f85829b73610975d04e694d46c04481e1b79/src/protocols/ExtWorkspace.hpp).

Using this instead of Hyprland's IPC/socket2 should make the workspaces module compatible with other compositors, while having the added benefit of exposing the `urgent` state (as far as I can tell, this is not available via the IPC at all and socket2 only exposes a "window requested `urgent` state" event, nothing for "window lost `urgent` state", let alone workspace-level `urgent` events).

PR #742, opened 2026-05-11
feat(workspaces): highlight workspaces with urgent windows

kiryl:
Plumbs a has_urgent flag through CompositorWorkspace and UiWorkspace and renders urgent workspaces with a danger-colored border (solid danger background when empty). Hidden urgent workspaces are widened so the highlight stays visible. Niri populates the flag from its workspace and window urgency state; Hyprland is left as a no-op for now (its IPC has no per-client urgent field, so it would need separate event tracking).

### DIFF
--- a/src/theme.rs
+++ b/src/theme.rs
@@ -624,6 +624,7 @@ impl AshellTheme {
     pub fn workspace_button_style(
         &self,
         is_empty: bool,
+        is_urgent: bool,
         colors: Option<Option<AppearanceColor>>,
     ) -> impl Fn(&Theme, Status) -> button::Style + use<> {
         let radius_lg = self.radius.lg;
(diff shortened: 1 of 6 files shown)


romanstingler:
just for tracking reasons
related to https://github.com/MalpenZibo/ashell/issues/317

src/theme.rs
```diff
@@ -654,18 +655,23 @@ impl AshellTheme {
...
+                    urgent_color
+                } else if is_empty {
                     theme.extended_palette().background.weak.color
                 } else {
                     bg_color
                 })),
                 border: Border {
-                    width: if is_empty { 1.0 } else { 0.0 },
-                    color: bg_color,
+                    width: if is_urgent || is_empty { 2.0 } else { 0.0 },
```
MalpenZibo: Why change the border width? Can you provide a screen with the two borders to make a comparison?

kiryl: 2026-05-28T22:33:34

It looks better to my eye. But, yeah, it is matter of taste. Want me to drop this? Or move to a separate commit?

MalpenZibo: If you can revert only the border size changes, I can merge the PR for the next release.

kiryl: Done.

PR #753, opened 2026-05-13
fix(tempo): swap temperature order to max/min in daily forecast

### LINKED ISSUE #737: Tempo order of temperature
@MalpenZibo 

All weather sites I know have the max temp first and the low temp second 

Image

we have it the other way round

Image

Question should we also do it like that ?

romanstingler:
Standard weather convention displays high temperature first
(max) followed by low temperature (min). The tempo module
was showing min/max order which is inconsistent with other
weather services.

closes https://github.com/MalpenZibo/ashell/issues/737

image


MalpenZibo (approved):
🚀

PR #754, opened 2026-05-13
Feat/Media Player Toolbar Cover

Gamanius:
Added support for displaying album cover art in the media player toolbar indicator.
Cover
CoverAndTitle
The following values have been added to the media player configuration field: 
| Value          | Description                                                                                                  |
| -------------- | ------------------------------------------------------------------------------------------------------------ |
| `Cover`        | Displays cover art when available; otherwise falls back to the media icon.                                   |
| `CoverAndTitle`| Displays cover art followed by the current title; if cover art is unavailable, falls back to icon and title. |

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -627,6 +627,8 @@ pub enum MediaPlayerFormat {
     Icon,
     #[default]
     IconAndTitle,
+    Cover,
+    CoverAndTitle
 }
 
 #[derive(Deserialize, Clone, Debug)]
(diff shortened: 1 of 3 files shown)


MalpenZibo:
Looks like a very small image O.o I will try this feature after the next release to see how it works

MalpenZibo:
I decided to block this PR for now because with the current design, the cover art is way too small to make sense

PR #756, opened 2026-05-16
feat(i18n): add French (fr-FR) translation

noirbizarre:
## Summary

- Adds a complete French translation catalog at `i18n/fr-FR/ashell.ftl`, mirroring every key and section from `i18n/en-US/ashell.ftl` (67 keys across 14 sections).
- Registers `fr-FR` in the `CATALOGS` slice in `src/i18n.rs` so the new locale is picked up by the runtime when `language = "fr-FR"` (or `LC_MESSAGES=fr_FR.UTF-8`).
- Documents the translation-adding workflow in `docs/src/contributing/common-tasks.md` — previously undocumented.

## Notes

- All Fluent placeholders (`{ \$count }`, `{ \$ssid }`, `{ \$duration }`, …) and selector keys (`[one]/*[other]`, `[on]/*[off]`) are preserved unchanged.
- French uses the same two CLDR plural categories as English; phrasings work for both 0 and 1 under `[one]`.
- `derive_units` in `src/i18n.rs` already falls through to Metric for `fr-FR` — no unit-system change needed.

## Verification

- `make check` passes (fmt + check + clippy with `-D warnings`).
- Tested locally with `language = "fr-FR"`: settings panels, updates module, notifications, system info, and weather conditions all render in French.
- Fallback still works for unknown locales (warns and uses `en-US`).

Split into two commits:
- `docs(i18n): document the translation workflow`
- `feat(i18n): add French (fr-FR) translation`

### DIFF
--- a/docs/src/contributing/common-tasks.md
+++ b/docs/src/contributing/common-tasks.md
@@ -67,6 +67,38 @@ pub fn my_button_style(&self) -> impl Fn(&Theme, Status) -> button::Style {
 }
 ```
 
+## Adding a Translation
+
+ashell uses [Fluent](https://projectfluent.org/) via the `i18n-embed` crate. 
+Catalogs live in `i18n/<lang-tag>/ashell.ftl` and are baked into the binary at compile time with `include_str!`.
+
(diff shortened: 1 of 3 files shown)


MalpenZibo (approved):
🚀

PR #757, opened 2026-05-17
fix(hyprland): support Lua dispatch protocol on 0.55+ with Lua config

### LINKED ISSUE #751: workspace switch on hyprland doesn't do anything
Using `0.8.0` release, since hyprland's `0.55.0` release clicking on workspace buttons on bar doesn't do anything. 
Probably because of hyprland's recent changes

romanstingler:
Hyprland 0.55.0 switched to Lua config by default and changed the
dispatch protocol: `hyprctl dispatch` now expects Lua dispatcher calls
instead of the old `hyprlang format`. This broke workspace switching
and all other dispatches when a Lua config is active.

Detect the config provider via 'hyprctl status' (configProvider: lua
vs configProvider: hyprlang) and choose the appropriate dispatch
method:
- Lua config: use 'hyprctl eval' with `hl.dsp.*` Lua API
- Hyprlang config: use the original `Dispatch::call()` socket protocol

This maintains backward compatibility with Hyprland <=0.54.x and
0.55.x users still running hyprlang configs.

closes https://github.com/MalpenZibo/ashell/issues/751

### DIFF
--- a/src/services/compositor/hyprland.rs
+++ b/src/services/compositor/hyprland.rs
@@ -14,7 +14,22 @@ use itertools::Itertools;
 use std::sync::{Arc, RwLock};
 use tokio::sync::broadcast;
 
-pub async fn execute_command(cmd: CompositorCommand) -> Result<()> {
+/// Detect whether Hyprland is using Lua or hyprlang config.
+/// Checks `hyprctl status` for the `configProvider` field.
+/// Falls back to hyprlang (false) if detection fails.
+fn is_lua_config() -> bool {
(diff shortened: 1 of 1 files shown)


MalpenZibo (approved):
Code it's ok! I didn't test it because right now I don't have a machine with Hyprland 0.55. Does it work well?

romanstingler:
Yesterday, I switched from `.conf` to `.lua` config and using this branch works fine on my side.

```
❯ hyprctl status

configProvider: lua
backend: drm
```

PR #760, opened 2026-05-19
fix: Use XDG_RUNTIME_DIR to store logs

francoisfreitag:
Fixes a crash when I open a session as another user. When ashell starts as user alpha, it creates /tmp/ashell/, owned by alpha. Then, logging in as user beta, ashell attempts to create /tmp/ashell for user beta, but that directory exists and is owned by user alpha.

In order to allow multiple users on the same system to use ashell, store the logs under a location specific to that user: `$XDG_RUNTIME_DIR/ashell`.
https://specifications.freedesktop.org/basedir/latest/#variables:~:text=actions%20history%20%28logs%2C%20history%2C%20recently%20used%20files%2C%20%E2%80%A6%29

## Notes

- <s>Added the xdg dependency. Multiple places implement XDG_* logic in ashell, the dependency may help simplify some locations. It's easy to remove the dependency and implement the specification if the dependency is an issue.</s> => [no](https://github.com/MalpenZibo/ashell/pull/760#discussion_r3272184244)
- Not sure if a changelog / upgrade doc is needed?

<details>
<summary>
Output from steps I took debugging the issue, it may help understanding it better.
</summary>
<pre>
```
$ RUST_BACKTRACE=1 ashell
INFO [ashell::config] Decoding config file "/home/pdi/.config/ashell/config.toml"
ERROR [ashell] Panic: panicked at /home/freitafr/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/flexi_logger-0.31.8/src/parameters/file_spec.rs:392:14:
called `Result::unwrap()` on an `Err` value: Os { code: 13, kind: PermissionDenied, message: "Permission denied" }
    0: <unknown>
   1: <unknown>
   2: <unknown>
   3: <unknown>
   4: <unknown>
   5: <unknown>
   6: <unknown>
   7: <unknown>
   8: <unknown>
   9: <unknown>
  10: <unknown>
  11: <unknown>
  12: <unknown>
  13: <unknown>
  14: <unknown>
  15: <unknown>
  16: <unknown>
  17: <unknown>
  18: <unknown>
  19: <unknown>
  20: <unknown>
  21: <unknown>
  22: <unknown>
  23: <unknown>
  24: <unknown>
  25: __libc_start_main
  26: <unknown>
```

```shell
$ strace ashell
# ...
write(1, "INFO [ashell::config] Decoding c"..., 82INFO [ashell::config] Decoding config file "/home/pdi/.config/ashell/config.toml"
) = 82
statx(AT_FDCWD, "/tmp/ashell/ashell_rCURRENT.log", AT_STATX_SYNC_AS_STAT, STATX_ALL, 0x7ffdba1601e0) = -1 EACCES (Permission denied)
statx(AT_FDCWD, "/tmp/ashell/ashell_rCURRENT.log", AT_STATX_SYNC_AS_STAT, STATX_ALL, 0x7ffdba1601e0) = -1 EACCES (Permission denied)
openat(AT_FDCWD, "/tmp/ashell", O_RDONLY|O_NONBLOCK|O_CLOEXEC|O_DIRECTORY) = -1 EACCES (Permission denied)
write(1, "\33[38;5;196mERROR\33[0m [ashell] \33["..., 324ERROR [ashell] Panic: panicked at /home/freitafr/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/flexi_logger-0.31.8/src/parameters/file_spec.rs:392:14:
called `Result::unwrap()` on an `Err` value: Os { code: 13, kind: PermissionDenied, message: "Permission denied" }
 disabled backtrace
) = 324
```

```shell
$ ls -l /tmp/ | rg ashell
drwx------ 2 freitafr freitafr  60 May 19 14:31 ashell
$ whoami
pdi
```
</details>

### DIFF
--- a/docs/src/contributing/testing-and-debugging.md
+++ b/docs/src/contributing/testing-and-debugging.md
@@ -6,10 +6,10 @@ ashell does not currently have an automated test suite. Testing is done manually
 
 ## Debugging with Logs
 
-ashell writes logs to `/tmp/ashell/`. To watch logs in real time:
+ashell writes logs to `$XDG_RUNTIME_DIR/ashell/`. To watch logs in real time:
 
 ```bash
-tail -f /tmp/ashell/*.log
+tail -f $XDG_RUNTIME_DIR/ashell/*.log
 ```
 
 ### Adjusting Log Level
@@ -39,7 +39,7 @@ In debug builds (`cargo build` without `--release`), all logs are also printed t
 Check logs for initialization errors:
 
 ```bash
-grep -i "error\|failed\|panic" /tmp/ashell/*.log
+grep -i "error\|failed\|panic" $XDG_RUNTIME_DIR/ashell/*.log
 ```
 
 ### D-Bus Issues
--- a/src/ipc.rs
+++ b/src/ipc.rs
@@ -140,8 +140,10 @@ pub fn socket_path() -> PathBuf {
         None => [
             std::env::temp_dir(),
             PathBuf::from(format!("ashell-{uid}.sock")),
-        ]
-    }.iter().collect()
+        ],
+    }
+    .iter()
+    .collect()
 }
 
 // ---------------------------------------------------------------------------
--- a/src/ipc.rs
+++ b/src/ipc.rs
@@ -13,6 +13,8 @@ use anyhow::{Context, Result, anyhow};
 use clap::Subcommand;
 use iced::Subscription;
 
+use crate::xdg;
+
 /// Maximum bytes to read from a client connection.
 const MAX_REQUEST_LEN: u64 = 4096;
 
@@ -132,11 +134,14 @@ impl FromStr for IpcCommand {
 }
 
 pub fn socket_path() -> PathBuf {
-    if let Some(dir) = std::env::var_os("XDG_RUNTIME_DIR") {
-        return PathBuf::from(dir).join("ashell.sock");
-    }
     let uid = unsafe { libc::getuid() };
-    PathBuf::from(format!("/tmp/ashell-{uid}.sock"))
+    match xdg::get_runtime_dir() {
+        Some(dir) => [dir, PathBuf::from("ashell.sock")],
+        None => [
+            std::env::temp_dir(),
+            PathBuf::from(format!("ashell-{uid}.sock")),
+        ]
+    }.iter().collect()
 }
 
 // ---------------------------------------------------------------------------
--- a/src/ipc.rs
+++ b/src/ipc.rs
@@ -131,12 +131,12 @@ impl FromStr for IpcCommand {
     }
 }
 
-pub fn socket_path() -> Result<PathBuf> {
+pub fn socket_path() -> PathBuf {
     if let Some(dir) = std::env::var_os("XDG_RUNTIME_DIR") {
-        return Ok(PathBuf::from(dir).join("ashell.sock"));
+        return PathBuf::from(dir).join("ashell.sock");
     }
     let uid = unsafe { libc::getuid() };
-    Ok(PathBuf::from(format!("/tmp/ashell-{uid}.sock")))
+    PathBuf::from(format!("/tmp/ashell-{uid}.sock"))
 }
 
 // ---------------------------------------------------------------------------
@@ -145,7 +145,7 @@ pub fn socket_path() -> Result<PathBuf> {
 
 /// Run the IPC client: connect to the daemon, send a command, print the response.
 pub fn run_client(cmd: &IpcCommand) -> Result<()> {
-    let path = socket_path()?;
+    let path = socket_path();
     let mut stream = UnixStream::connect(&path)
         .with_context(|| format!("connect to {} — is ashell running?", path.display()))?;
 
@@ -187,7 +187,7 @@ enum ListenerError {
 /// remove the file or bind a new listener — otherwise we'd orphan the
 /// primary's fd and break `ashell msg` until it's restarted.
 fn create_listener() -> std::result::Result<UnixListener, ListenerError> {
-    let path = socket_path().map_err(ListenerError::Other)?;
+    let path = socket_path();
 
     match UnixStream::connect(&path) {
         Ok(_) => return Err(ListenerError::AlreadyRunning),
--- a/src/main.rs
+++ b/src/main.rs
@@ -80,9 +80,8 @@ fn main() -> iced::Result {
 
     debug!("args: {args:?}");
 
-    let logdir = xdg::get_runtime_dir().unwrap_or_else(
-        || [env::temp_dir(), PathBuf::from("ashell")].iter().collect()
-    );
+    let logdir = xdg::get_runtime_dir()
+        .unwrap_or_else(|| [env::temp_dir(), PathBuf::from("ashell")].iter().collect());
     let logger = Logger::with(
         LogSpecBuilder::new()
             .default(log::LevelFilter::Info)
--- a/src/main.rs
+++ b/src/main.rs
@@ -8,6 +8,7 @@ use flexi_logger::{
 use iced::{Anchor, Font, KeyboardInteractivity, Layer, LayerShellSettings};
 use log::{debug, error, warn};
 use std::backtrace::Backtrace;
+use std::env;
 use std::panic;
 use std::path::PathBuf;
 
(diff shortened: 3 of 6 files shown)


GonuDvc (commented):
The happy path is clean. What happens if the third-party API returns 429 here? Might be worth handling that case explicitly.

Cargo.toml
```diff
@@ -95,6 +95,7 @@ i18n-embed = { version = "0.16", default-features = false, features = [
 i18n-embed-fl = "0.10"
 rust-embed = "8"
 unic-langid = "0.9"
+xdg = "2.5.2"
```
MalpenZibo: Can we avoid the new dependency? It's a very small library, not very updated, and I prefer to avoid this dependency

docs/src/contributing/testing-and-debugging.md
```diff
@@ -76,7 +76,7 @@ WGPU_BACKEND=gl ashell
 You can run ashell with a custom config for testing:
 
 ```bash
-ashell --config-path /tmp/test-config.toml
+ashell --config-path $XDG_STATE_HOME/test-config.toml
```
MalpenZibo: This should be reverted. A tmp config for testing should be placed in the tmp folder

src/main.rs
```diff
@@ -78,12 +80,16 @@ fn main() -> iced::Result {
 
     debug!("args: {args:?}");
 
+    let logdir = match BaseDirectories::with_prefix("ashell") {
+        Ok(dirs) => dirs.get_state_home(),
+        Err(_) => [env::temp_dir(), PathBuf::from("ashell")].iter().collect(),
+    };
```
MalpenZibo: Projects like hyprland/noctalia use $XDG_RUNTIME_DIR.
The main difference is the persistence and disk access. $XDG_RUNTIME_DIR is on RAM $XDG_STATE_HOME is on DISK.

For `ashell` we have runtime logs primarily, things for debug, without any log-lived information messages, that's the main point of using the `/tmp` folder. We should decide if make sense to persist this log on disk.

IMO we could use the $XDG_RUNTIME_DIR, I don't think that persisting `ashell` logs makes much sense.

@clotodex @romanstingler, what do you think?

```rust
  let logdir = std::env::var_os("XDG_RUNTIME_DIR")
      .map(|dir| PathBuf::from(dir).join("ashell"))
      .unwrap_or_else(|| {
          let uid = unsafe { libc::getuid() };
          PathBuf::from(format!("/tmp/ashell-{uid}"))
      });
```

francoisfreitag:
https://github.com/MalpenZibo/ashell/pull/760#pullrequestreview-4320224494 does not make sense to me, looks like bot activity.

src/main.rs
romanstingler: yeah 
`XDG_RUNTIME_DIR=/run/user/1000`
`tmpfs on /run/user/1000 type tmpfs (rw,nosuid,nodev,relatime,size=3214072k,nr_inodes=803518,mode=700,uid=1000,gid=1000,inode64,huge=within_size)` 
so this is RAM and `/tmp` is RAM (on my machines) so yeah that's fine

francoisfreitag: The specification mentions:

> If `$XDG_RUNTIME_DIR` is not set applications should fall back to a replacement directory with similar capabilities and print a warning message. Applications should use this directory for communication and synchronization purposes and should not place larger files in it, since it might reside in runtime memory and cannot necessarily be swapped out to disk. 

As you mentioned, other projects don’t seem to follow the spec closely.
I doubt the RAMFS performance is needed for the less than ~10 lines that ashell outputs (on my machine, with the default config).

If you’re feeling strongly about this, I’ll update to use `XDG_RUNTIME_DIR`, but that doesn’t seem like the best choice to me.

Another option is to log to the system journal (syslog, supported by flexi_logger https://docs.rs/flexi_logger/latest/flexi_logger/writers/struct.SyslogWriter.html). That allows rotation and centralization, plus all the management capabilities that syslog offers.

docs/src/contributing/testing-and-debugging.md
francoisfreitag: Oopsie, thanks! :+1:

Cargo.toml
francoisfreitag: Sure thing!

src/main.rs
romanstingler: I dont care as long as I have an option to use my ram for that. 
I often test stuff and have to use debug and so on and there it writes hundreds of MB to GB in matter of minutes, depending on what is debugged and what is tested.

francoisfreitag: I was thinking to suggest something like:
```shell
$ XDG_STATE_HOME=/tmp/ashell cargo run
```

But that will impact cargo. Also, using tools like [direnv](https://direnv.net/) to set it automatically will have side effects on all commands run from the ashell clone directory (e.g. `nvim`).

So, if RAMFS is a requirement, maybe `XDG_RUNTIME_DIR` is a better option.

francoisfreitag:
Should be updated to use `XDG_RUNTIME_DIR` per https://github.com/MalpenZibo/ashell/pull/760#discussion_r3272332931

francoisfreitag:
I may attempt to write some tests for the new function `get_runtime_dir` if we agree on the approach.

src/ipc.rs
```diff
@@ -132,11 +134,19 @@ impl FromStr for IpcCommand {
...
+        PathBuf::from(format!("ashell-{uid}.sock")),
+    ].iter().collect();
+    let runtime_dir = match xdg::get_runtime_dir() {
+        Ok(dir) => match dir {
+            Some(dir) => dir,
+            None => default_runtime_dir,
+        },
+        Err(_) => default_runtime_dir,
+    };
+    Ok(PathBuf::from(runtime_dir).join("ashell.sock"))
```
MalpenZibo: The fallback case produces `/tmp/ashell-1000.sock/ashell.sock` that's not good, it should be `/tmp/ashell-1000.sock`

src/xdg.rs
```diff
@@ -0,0 +1,32 @@
...
+    }
+    let metadata = runtime_dir.metadata()?;
+    if !metadata.is_dir() {
+        return Ok(None);
+    }
+    let uid = unsafe { libc::geteuid() };
+    if fs::metadata(&runtime_dir)?.st_uid() != uid {
+        return Ok(None);
+    }
+    if metadata.permissions().mode() != 0o700 {
```
MalpenZibo: This check always fails on my machine

src/xdg.rs
```diff
@@ -0,0 +1,32 @@
...
+use std::io;
+use std::os::linux::fs::MetadataExt;
+use std::os::unix::fs::PermissionsExt;
+use std::path;
+
+// $XDG_RUNTIME_DIR defines the base directory relative to which user-specific non-essential runtime
+// files and other file objects (such as sockets, named pipes, ...) should be stored.
+// The directory MUST be owned by the user, and they MUST be the only one having read and write
+// access to it. Its Unix access mode MUST be 0700.
+pub fn get_runtime_dir() -> Result<Option<path::PathBuf>, io::Error> {
```
MalpenZibo: This signature is strange... Why a `Result<Option>`? We always use the default value in case of error or none result, so why the double encapsulation?

Can I suggest
```rust
pub fn get_runtime_dir() -> Option<PathBuf> {
    let runtime_dir = PathBuf::from(env::var_os("XDG_RUNTIME_DIR")?);
    let metadata = runtime_dir.metadata().ok()?;
    let uid = unsafe { libc::geteuid() };
    (runtime_dir.is_absolute()
        && metadata.is_dir()
        && metadata.st_uid() == uid
        && metadata.permissions().mode() & 0o777 == 0o700)
        .then_some(runtime_dir)
}
```

MalpenZibo (changes requested):

src/xdg.rs
francoisfreitag: I wanted to distinguish between:
- `XDG_RUNTIME_DIR` is set but invalid
- `XDG_RUNTIME_DIR` could not be checked (e.g. getting the metadata failed)

But that’s unnecessary for the current usage, so might as well simplify. :+1:

I’m not (yet) familiar with Rust. Your version is much more readable, thanks!

MalpenZibo (approved):
🚀

src/xdg.rs
francoisfreitag: Thanks for merging! I assume this check stopped failing on your machine?

MalpenZibo: The trick was `metadata.permissions().mode() & 0o777 == 0o700`

francoisfreitag: :facepalm: Thanks!

PR #766, opened 2026-05-21
fix: replace panicking code paths with graceful error handling

dacrab:
## Summary

Replaces `unwrap()`, `expect()`, and `panic!()` calls in runtime code paths with proper error handling. The bar will now log warnings and degrade gracefully instead of crashing when D-Bus connections fail, PulseAudio restarts, or system calls error.

Also converts blocking `std::process::Command` calls to async `tokio::process::Command` to avoid blocking the tokio runtime.

## Changes

| File | Fix |
|------|-----|
| `src/services/network/dbus.rs` | 4 `.unwrap()` in D-Bus signal handlers → `match` + `warn!` with fallback |
| `src/services/tray/dbus.rs` | 5 `.unwrap()` in signal emitters → `match` + `warn!` with early return |
| `src/services/audio.rs` | `panic!()` in PulseAudio init → `return Err(...)` |
| `src/utils/launcher.rs` | `std::process::Command` → `tokio::process::Command` |
| `src/services/compositor/hyprland.rs` | `is_lua_config`/`dispatch_lua` → async with `tokio::process::Command`; `std::sync::RwLock` → `tokio::sync::RwLock` |
| `src/config.rs` | Hot-reload file read wrapped in `spawn_blocking` |
| `src/services/idle_inhibitor.rs` | Check `ftruncate` return value |
| `src/modules/custom_module.rs` | `.expect()` on child wait → `match` + `warn!` |
| `src/app.rs` | Signal stream `.expect()` → graceful fallback with empty stream |

## Testing

- `cargo check` ✓
- `cargo clippy -- -D warnings` ✓ (zero warnings)

## Motivation

If NetworkManager restarts, PulseAudio crashes, or a D-Bus signal emission fails, the bar previously panicked and crashed. These are recoverable conditions — the bar should log the error and continue operating with degraded functionality until the service recovers.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,7 +21,7 @@ jobs:
     steps:
       - uses: actions/checkout@v6
 
-      - uses: dorny/paths-filter@v3
+      - uses: dorny/paths-filter@v4
         id: filter
         with:
           filters: |
--- a/.github/workflows/nix-ci.yml
+++ b/.github/workflows/nix-ci.yml
@@ -21,7 +21,7 @@ jobs:
     steps:
       - uses: actions/checkout@v6
 
-      - uses: dorny/paths-filter@v3
+      - uses: dorny/paths-filter@v4
         id: filter
         with:
           filters: |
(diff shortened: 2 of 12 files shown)


MalpenZibo (approved):
🚀

PR #773, opened 2026-05-23
Force cpu temp to always be metric

TheBlueOompaLoompa:
For the most part, temps for computer components are always in Celsius. Even Task Manager on Windows uses Celsius when set to an English US region. I don't think I've personally ever seen anyone in the US use Fahrenheit to talk about computers.

This PR just locks the `SystemInfo` module to always give CPU temps in Celsius. Although, this could be a config option if anyone says they still want Fahrenheit.

### DIFF
--- a/docs/src/reference/config-reference.md
+++ b/docs/src/reference/config-reference.md
@@ -152,7 +152,7 @@ alert_threshold = 80
 ```
 
 **Dependencies:**
-- Temperature monitoring reads the kernel `hwmon` sysfs interface directly (no extra package required). The `sensor` option is either a type keyword (`"Cpu"`, `"Gpu"`, `"Acpi"`, `"Nvme"`) for auto-detection or an exact hwmon label (e.g. `"acpitz temp1"`) — run `sensors` (from `lm_sensors`) to find the right name. The displayed unit follows the locale / unit system, not a per-module option.
+- Temperature monitoring reads the kernel `hwmon` sysfs interface directly (no extra package required). The `sensor` option is either a type keyword (`"Cpu"`, `"Gpu"`, `"Acpi"`, `"Nvme"`) for auto-detection or an exact hwmon label (e.g. `"acpitz temp1"`) — run `sensors` (from `lm_sensors`) to find the right name. The displayed unit follows the locale / unit system unless `units` (`"Celsius"` / `"Fahrenheit"`) overrides it.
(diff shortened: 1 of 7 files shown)


dubyatp:
Any particular reason why this was closed so suddenly? I just updated and the F cpu temperature is bothering me, haha. I too primarily use F for most things except hardware temps. Seems like there's no other way around except for code changes

I would probably support it being toggleable, but in most cases, C should be the default when it comes to hw temps imho.

TheBlueOompaLoompa:
There was work on an option that lets you configure it directly that I didn't find. I'm not sure of the state of that but I think it'll be on the next release.

TheBlueOompaLoompa:
Found it https://github.com/MalpenZibo/ashell/pull/668

TheBlueOompaLoompa:
I did some more investigation, apparently they removed the temperature configuration from that PR, and that PR is where they made it based on unit system. I'll re-open this.

MalpenZibo:
So we're talking about the CPU temp in the system info module. 

The work on the i18n of the measure unit started, I think here https://github.com/MalpenZibo/ashell/issues/663, where a user specific request the `fahrenheit` for the system info module.

That said, I don't know much about the imperial system and how it's utilized in the US (from my point of view, the entire imperial system doesn't make any sense, but this is another story :sweat_smile:).

Now you're asking the exact opposite of the original issue. So, from your point of view, is the Fahrenheit scale used for weather and the Celsius one for computers?

logannc:
> That said, I don't know much about the imperial system and how it's utilized in the US (from my point of view, the entire imperial system doesn't make any sense, but this is another story 😅).

No one not having some banter disagrees, but it's still what we're used to. :) 

That said, I think `region` controlling everything implicitly is too coarse grained. For example, I'd want US date formats, Fahrenheit for weather, but Celsius for CPU temp - we typically use Celsius for science or technical things. Also the range of CPU temps approximately cap out around 100C so its a more convenient range, but that's getting into personal preference.

MalpenZibo:
Ok, but let's make a config to override the regional setting for this case.

Something like:
```
pub struct SystemInfoTemperature {
    warn_threshold: Option<i32>,
    alert_threshold: Option<i32>,
    pub sensor: TemperatureSensor,
    units: Option<celsius | fahrenheit>
}
```

By default, we should follow the global region settings. If this field is populated (by default should be none), then we use this field to override the global region setting.

Remember to update the documentation

MalpenZibo (approved):

PR #774, opened 2026-05-23
feat(audio): configurable volume_step and max_volume with overdrive i…

sudo-Tiz:
### Features

- **`volume_step`** (default: 5) — configurable step size for IPC volume up/down commands
- **`max_volume`** (default: 100, max: 200) — allows volume to exceed 100% hardware level

### Overdrive visual feedback (when volume > 100%)

- **Panel icon** switches to `SpeakerOverdrive` (󰝝)
- **Settings slider** rail turns red (`danger` color)
- **OSD progress bar** turns red (`progress_bar::danger`)
- **OSD percentage** (opt-in via `show_volume_percentage`) displays the actual value

### Config

```toml
[settings]
volume_step = 5      # 1..=50
max_volume = 150     # 1..=200, default 100

[osd]
show_volume_percentage = true
show_brightness_percentage = false
```

### Design choices

- Overdrive is **derived, never stored** — computed from `value > 1.0` in views, no extra field propagated through messages
- OSD value is normalized to `NORMAL_VOLUME` (not `vol_max`), so it can exceed 1.0 naturally
- Panel icon normalization uses `NORMAL_VOLUME` (100% hardware) so all 3 speaker levels remain reachable below 100%
- Microphone is **not** affected — stays capped at 100%
- Config values are clamped with a `warn!` log if out of range


MalpenZibo (commented):
<del>Nice! Can you fix the docs conflict?</del>

Done :smile:

MalpenZibo (approved):
🚀

PR #775, opened 2026-05-24
ci: upgrade to Node.js 24 and fix GitHub Pages deployment

dacrab:
## Changes

This PR fixes GitHub Pages deployment issues and upgrades workflows to Node.js 24 to avoid deprecation warnings.

### Fixes:
- Upgraded to Node.js 24 (Node 20 will be deprecated on June 2nd, 2026)
- Updated `deploy-pages` action from v4 to v5
- Added `FORCE_JAVASCRIPT_ACTIONS_TO_NODE24` environment variable for actions that don't yet support Node 24 natively

### Context:
Node.js 20 actions are deprecated and will be removed from GitHub runners on September 16th, 2026. This PR ensures the workflows continue to work without warnings.

Tested in my fork and deployment works correctly.

PR #777, opened 2026-05-25
ci: upgrade dorny/paths-filter v3 → v4 (node24)

dacrab:
## Changes

Upgrade `dorny/paths-filter` from `v3` to `v4` in `ci.yml` and `nix-ci.yml`.

### Why

`v3` runs on `node20`, which is deprecated and will be forcibly removed from GitHub runners on **September 16, 2026**. `v4` (released March 2026) upgrades the runtime to `node24`.

All other actions in the repo are already on node24-compatible versions. `peaceiris/actions-mdbook` is the only remaining node20 action but has no node24 release yet — it is handled separately via `FORCE_JAVASCRIPT_ACTIONS_TO_NODE24` in the gh-pages workflow (PR #775).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,7 +21,7 @@ jobs:
     steps:
       - uses: actions/checkout@v6
 
-      - uses: dorny/paths-filter@v3
+      - uses: dorny/paths-filter@v4
         id: filter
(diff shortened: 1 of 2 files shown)


MalpenZibo (approved):
👌

PR #779, opened 2026-05-26
feat(power): add UPower charge limit toggle to quick settings

Scott-Nx:
## Summary

- Add UPower-based battery charge-limit detection and toggling.
- Show a Charge Limit quick settings button when supported/readable.
- Reflect active charge-limit state on the battery pill/top bar only while not discharging.
- Use `󱈏` (`f120f`) in quick settings and `󱈑` (`f1211`) in the compact battery pill/top bar.

## Details

- Uses UPower D-Bus only:
    - `ChargeThresholdSupported`
    - `ChargeThresholdEnabled`
    - `EnableChargeThreshold(boolean)`
- Hides the Charge Limit control when unsupported or unreadable.
- Keeps normal battery icons/styles while UPower reports discharging.
- Uses an optimistic state fallback if refresh fails after a successful toggle.
- Guards empty charge-limit event streams with `pending()`.

## Validation

- `make check`
- `cargo run` (manual real system test)

> [!NOTE]
>
> No sysfs fallback or percentage threshold editing is implemented.

---

## Demo

https://github.com/user-attachments/assets/b36909cc-f65b-4275-a95e-06ae185de6a3


MalpenZibo (commented):
preliminary review. Can we user a quick settings button insead of the current one to enable/disable the limit? Also could we use this icon 󱈏 (`f120f`)?

Scott-Nx:
> preliminary review. Can we user a quick settings button insead of the current one to enable/disable the limit? Also could we use this icon 󱈏 (`f120f`)?

Sure. I’ll move the charge-limit toggle to a quick settings button and switch the active icon to 󱈏 (f120f).

Scott-Nx:
> preliminary review. Can we user a quick settings button insead of the current one to enable/disable the limit? Also could we use this icon 󱈏 (`f120f`)?

I switched the quick settings icon to `󱈏` (`f120f`) as requested, and it looks good in the expanded quick settings button.

One thing I noticed is that using the same icon in the top bar looks less clear than the previous `󱈑` (`f1211`) icon. Would you prefer using `󱈏` (`f120f`) everywhere for consistency, or using `󱈏` (`f120f`) for the quick settings button while keeping `󱈑` (`f1211`) for the top bar?

ภาพ

MalpenZibo:
Using 󱈏 (f120f) for the quick settings button while keeping 󱈑 (f1211) for the top bar!

Scott-Nx:
> Using 󱈏 (f120f) for the quick settings button while keeping 󱈑 (f1211) for the top bar!

Updated as requested. The Charge Limit quick settings button now uses `󱈏` (`f120f`), while the compact battery pill/top bar keeps `󱈑` (`f1211`).

MalpenZibo (approved):
👌

PR #782, opened 2026-05-27
tempo: fix hourly forecast and current weather time to use location timezone

### LINKED ISSUE #781: Forecast times shown based on UTC not local time
It is around 01:00 local time here at time of writing, and the sky is overcast. The weather widget seems to be basing its times on UTC, so while the correct current conditions are correct, the hourly forecast is offset by 6 hours. My configuration for the Tempo widget is as follows:

```toml
[tempo]
clock_format = "%a %F %T"
weather_location = { City = "Calgary" }
weather_indicator = "Icon"
```

Image

romanstingler:
Previously, the hourly forecast time was not changing according to the selected TZ.

Old 
image

---
New 
image

  
image

  closes https://github.com/MalpenZibo/ashell/issues/781

### DIFF
--- a/src/modules/tempo.rs
+++ b/src/modules/tempo.rs
@@ -494,6 +494,8 @@ impl Tempo {
             .zip(self.location.as_ref())
             .map(|(data, location)| {
                 let inner_element: Element<'a, Message> = if location_visible {
+                    let display_time =
+                        self.time_str("%R", self.current_timezone_index, Some(data.current.time));
                     text(format!(
                         "{}{} - {}",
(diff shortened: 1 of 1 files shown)


goose121:
As a note I think it might be more consistent to use the system timezone (or the clock's specified timezone, I'm not sure if it can be different) instead of the weather location timezone so there aren't ever multiple different timezones without an indication in case the weather location is different from the clock's timezone, but that's probably a matter of opinion

MalpenZibo (approved):
🚀

PR #785, opened 2026-05-29
Chore/upower cleanup

romanstingler:
**dbus.rs:**
- Changed `&self` to `self` for `is_peripheral()`, `is_power_source()`, and `description()` methods (the enum is `Copy`, so passing by value is more efficient)

**mod.rs:**
- Changed `&self` to `self` for `get_icon()` and `get_battery_icon()` methods (enum is `Copy`)
- Changed `&Option<Vec<...>>` to `Option<&Vec<...>>` for `system_battery_devices` parameter (more idiomatic)
- Removed trailing comma in error message
- Removed unused imports (`ThrottleExt`, `stream_select`)
</assistant>

### DIFF
--- a/src/services/upower/mod.rs
+++ b/src/services/upower/mod.rs
@@ -172,7 +172,7 @@ impl fmt::Display for PeripheralDeviceKind {
 }
 
 impl PeripheralDeviceKind {
-    pub fn get_icon(&self) -> StaticIcon {
+    pub fn get_icon(self) -> StaticIcon {
         match self {
             PeripheralDeviceKind::Keyboard => StaticIcon::Keyboard,
             PeripheralDeviceKind::Mouse => StaticIcon::Mouse,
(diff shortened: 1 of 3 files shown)


megabyte6 (changes requested):
~~What's the reason for removing the `&` in `&self`? Doesn't this make the function less flexible?~~

Edit: Sorry, I missed your explanation. Please ignore this.

megabyte6 (approved):
Looks good to me 👍

src/services/upower/mod.rs
```diff
@@ -520,6 +521,7 @@ impl UPowerService {
             for device_path in peripheral_paths {
                 let device = upower.get_device(device_path).await?;
 
+                let conn = conn.clone();
```
MalpenZibo: why all these new `conn.clone()`?

romanstingler: damn thanks, fixed
missed that 

image

I tested a few clippy lints to see what they are suggesting and this one 
https://rust-lang.github.io/rust-clippy/master/index.html#needless_continue

would prefer `select_all` and I was curious why it would go for this and what that would mean in our case. 
In the end I reverted that one and missed the 2 lines. 

Thanks for being so thorough. just shows me that sometimes when I already get tired shouldn't push through but stop and continue with a fresh head.

MalpenZibo: Sometimes it's a miracle that I'm able to spot something! Usually, I do some reviews after my 8 hours of work :sweat_smile:

MalpenZibo (approved):
👌

src/services/upower/mod.rs
romanstingler: You have super powers from all the spaghetti from nonna  :)

PR #786, opened 2026-05-30
feat(media_player): support configurable indicator fields

mengdehong:
## Summary

This PR adds configurable display fields to the media player indicator, mainly to reduce the space used by long artist names or song titles.

## Changes

- Updated `indicator_format` to support:
  - `Text`: displays text only
  - `IconAndText`: displays both the icon and text
  - `Icon`: displays the icon only
- Added `indicator_fields` to configure which MPRIS metadata fields are displayed:
  - `Artist`
  - `Title`
  - `Album`
- Metadata fields are displayed in their configured order and joined with ` - `.
- Updated the media player documentation and configuration reference.
- Note: `IconAndTitle` remains supported as an alias for `IconAndText` for backward compatibility.

## Default Behavior

The default configuration preserves the existing indicator appearance:

```toml
[media_player]
indicator_format = "IconAndText"
indicator_fields = ["Artist", "Title"]
```

## Example

Display only the song title in the status bar:

```toml
[media_player]
indicator_format = "Text"
indicator_fields = ["Title"]
```

## Screenshot

With `indicator_fields = ["Title"]`, the indicator no longer displays the artist name.
Media player indicator displaying only the song title

### DIFF
--- a/docs/src/reference/config-reference.md
+++ b/docs/src/reference/config-reference.md
@@ -190,7 +190,9 @@ on_click = "vpn-toggle"
 
 ```toml
 [media_player]
-format = "{artist} - {title}"
+max_title_length = 100
+indicator_format = "IconAndText" # "Text", "IconAndText", or "Icon"
+indicator_fields = ["Artist", "Title"]
 ```
 
 **Dependencies:** Any MPRIS-compatible media player (e.g., Spotify, Firefox, VLC, Strawberry). No extra system package is needed.
--- a/src/config.rs
+++ b/src/config.rs
@@ -721,8 +721,18 @@ pub struct SettingsCustomButton {
 #[derive(Deserialize, Copy, Clone, Default, PartialEq, Eq, Debug)]
 pub enum MediaPlayerFormat {
     Icon,
+    #[serde(alias = "Title")]
+    Text,
+    #[serde(alias = "IconAndTitle")]
     #[default]
-    IconAndTitle,
+    IconAndText,
+}
+
+#[derive(Deserialize, Copy, Clone, PartialEq, Eq, Debug)]
+pub enum MediaPlayerTextField {
+    Artist,
+    Title,
+    Album,
 }
 
 #[derive(Deserialize, Clone, Debug)]
@@ -731,6 +741,7 @@ pub struct MediaPlayerModuleConfig {
     pub max_title_length: u32,
     pub indicator_format: MediaPlayerFormat,
     pub show_visualizer: bool,
+    pub indicator_fields: Vec<MediaPlayerTextField>,
 }
 
 impl Default for MediaPlayerModuleConfig {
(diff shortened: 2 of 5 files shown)


MalpenZibo (changes requested):
**NOTE**: I rebased the PR

Very nice!

Only one minor thing. Can we completely disable the menu if `menu_fields` is set to an empty array?

we can add a `menu_enabled` method that returns `!self.config.menu_fields.is_empty()`

mengdehong:
Completed, `menu_fields = []` now disables the media player menu while keeping the bar indicator visible.

MalpenZibo (changes requested):
Some doubt.

In the menu, we are emphasizing the first element in the list for the `menu_field_text` keys. Before this change, it was always the Title.

The menu_fields list doesn't respect a real order; it's more like an enable/disable list in the menu.

So we have 3 choices:
- Make the Title always the one with the emphasis regarding the position (if it's not present, then no emphasis). It's strange to have something with emphasis in the second or third row
- Add to the documentation that order matters, but only for the `menu_field_text`. Meh solution, I mean it's strange to have something that respects an order and other entries that do not.
- Change the config: one to specify the description metadata, the `menu_field_text`, and three other options to enable/disable the cover, button controls, and volume indicator. If the description metadata doesn't contain any key, disable the menu because the other components could not be present (a player with no volume or control and no way to get a cover)

So we have some issues here.

From my point of view, the indicator part 👍 and can be merged. Maybe we should change the `indicator_format` values in `Icon`, `IconAndText`, `Text`.

The menu part doesn't convince me for a couple of reasons:
- gives the user too many possibilities. We need to test more stuff, and in general, `ashell` as its design is not like other projects where you can customize almost everything. So here it's a borderline field.
- Is it really so important to customize the description metadata, controls and stuff in the menu? I mean, maybe someone prefers not to lead the cover image, but the rest? it really matter? It's something that appears only in a dedicated menu, so I don't know. From my point of view, the pros are not enough
- We still have some issues in some configurations where the risk is to open the menu just to see empty cards

My idea is that we could maintain only the stuff for the indicator and open an issue for the rest.

@clotodex @romanstingler WDYT?

mengdehong:
That makes sense.

My original motivation for this PR was only to solve the issue where some long song titles or artist names take too much space in the bar. The menu customization was more of a side change while working on the same area.

  If that part adds more design concerns or maintenance burden, I’m happy to remove the menu-related changes from this PR. Sorry for adding extra scope there.

MalpenZibo:
No problem. Without the menu change from my point of view the PR is ok!

mengdehong:
I have removed the menu customization changes and restored the original menu behavior. I also rebased and squashed the PR history into a single commit for clarity.

MalpenZibo (approved):
👍

PR #787, opened 2026-05-30
feat(tray): support checkbox submenus

Lykathia:
A menu item can be a checkbox and a submenu parent at the same time.

This change checks the properties independently instead of the previous mutually exclusive matches.

Two side effects of expanding submenus were observed and included:
    - A menu with many items could grow past the screen edge. Added a height cap and scrolling.
    - Checkbox rows were lacking a hover highlight.


Lykathia:
Yes... we have too many menu items 😅

tailscale

MalpenZibo (approved):
🚀

PR #788, opened 2026-06-01
add configurable wind speed unit

### LINKED ISSUE #784: Feature request: Tempo - show m/s
Hello!

In my home country Sweden, wind speed is standardized to show in m/s, or [meter per second](https://en.wikipedia.org/wiki/Metre_per_second).

Currently, tempo allows to toggle between Imperial (miles per hour), or Metric (kilometer per hour).

For us confused swedes, it would be very useful to be able to configure the tempo module to show wind speed in this unit as km/h is kind of arcane unit around here. Local weather apps, news reporting and newspapers all use m/s.

romanstingler:
@MalpenZibo 

I added the functionality to override the `wind_speed_unit` in the config.

I was not sure if you want to re-calculate the km/h to m/s internally or if you want to use the OpenMeteo's API params like I did here.

Basically, we still default to the locale-based temperature units if this is not set otherwise we override that one.
Also no clue if you want to have that in lowercase or like this.

```
[tempo]
wind_speed_unit = "Ms"
```

closes https://github.com/MalpenZibo/ashell/issues/784

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -1,5 +1,5 @@
 use crate::app::Message;
-use crate::i18n::UnitSystem;
+use crate::i18n::{UnitSystem, unit_system};
 use crate::services::upower::PeripheralDeviceKind;
 use crate::utils::celsius_to_fahrenheit;
 use hex_color::HexColor;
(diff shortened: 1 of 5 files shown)


MalpenZibo (approved):
It's ok like this!! nice job 🚀

PR #794, opened 2026-06-05
chore - Update dependencies 

MalpenZibo:


romanstingler:
https://github.com/MalpenZibo/iced_layershell/pull/21

romanstingler:
@MalpenZibo 

Just documenting 

zbus 5.14 ->  5.16 
introduced 

```
WARN [zbus::proxy] Failed to populate properties cache via GetAll: org.freedesktop.DBus.Error.InvalidArgs: No such interface 'org.bluez.Battery1'. Property change streams will not produce values.
```

```
❯ git bisect bad
d8c53637f429cafc90c48a9cf5cdd23a806d118f is the first bad commit
commit d8c53637f429cafc90c48a9cf5cdd23a806d118f (HEAD)
Author: Simone Camito
Date:   Fri Jun 5 11:32:11 2026 +0200

    upd deps

 Cargo.lock | 1474 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++---------------------------------------------------------------------------------------
 Cargo.toml |   10 +-
 2 files changed, 790 insertions(+), 694 deletions(-)

```

The new version eagerly calls `GetAll` on proxy construction to populate a properties cache, which produces these warnings.

related issues

Issue #1325 — "Log errors from `GetAll` method call": https://github.com/dbus2/zbus/issues/1325 This is the issue that led to the `WARN` log being added. The request was to stop silently swallowing `GetAll` errors.

PR #1389 — "Make PropertyStream yield changes if service starts later": https://github.com/dbus2/zbus/pull/1389 This is the fix that changed `PropertiesCache::init()` to not fail on `GetAll` errors (so property streams work even when the service isn't running yet). But the warnings are still emitted.

PR #796, opened 2026-06-06
feat(settings): fallback text in empty tooltip popups

MalpenZibo:


romanstingler:
for me that looks fine, 
If you are ok with it that we don't have anything for brightness (I had no idea if/what to show there)

romanstingler (approved):

MalpenZibo:
Neither I. For now, it's ok like it is.

PR #797, opened 2026-06-06
fix: suppress zbus proxy warnings for optional interfaces

romanstingler:
- Add `has_battery` field to BluetoothDevice to avoid creating `BatteryProxy` for devices that don't support `org.bluez.Battery1`
- Use `CacheProperties::No` on `BatteryProxy` and `PowerProfilesProxy` to skip eager `GetAll` calls that produce warnings when services/interfaces are not available

related to https://github.com/MalpenZibo/ashell/pull/794#issuecomment-4640326723

### DIFF
--- a/src/services/bluetooth/mod.rs
+++ b/src/services/bluetooth/mod.rs
@@ -29,6 +29,7 @@ pub struct BluetoothDevice {
     pub path: OwnedObjectPath,
     pub connected: bool,
     pub paired: bool,
+    pub has_battery: bool,
 }
 
 #[derive(Debug, Clone)]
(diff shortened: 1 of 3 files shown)


src/services/bluetooth/mod.rs
```diff
@@ -120,17 +121,20 @@ impl BluetoothService {
...
-                        battery
-                            .receive_percentage_changed()
-                            .await
-                            .map(|_| {})
-                            .boxed(),
-                    );
+                    if device.has_battery {
+                        let battery = BatteryProxy::builder(conn)
+                            .path(device.path.clone())?
+                            .cache_properties(zbus::proxy::CacheProperties::No)
```
MalpenZibo, line 127: hum... I suspect that this cacheProperties::No could break the event stream.

From the zbus docs 
> The stream will yield the current value first, then wait for the value changes. If caching is not enabled on this proxy, the resulting stream will not return any events.

https://docs.rs/zbus/latest/zbus/proxy/struct.Proxy.html#method.receive_property_changed

src/services/upower/mod.rs
```diff
@@ -678,7 +681,10 @@ impl UPowerService {
             })
             .boxed();
 
-        let powerprofiles = PowerProfilesProxy::new(conn).await?;
+        let powerprofiles = PowerProfilesProxy::builder(conn)
+            .cache_properties(zbus::proxy::CacheProperties::No)
```
MalpenZibo, line 685: same problem here

Issue #615, opened 2026-04-09
Replace style config. Feedback request


MalpenZibo:
Remove the style config, adding 3 new config values:
- status bar margin
- status bar padding
- status bar background color

I'm planning to remove the Gradient option. Not sure if it's really used.

MalpenZibo:
# My idea:

## Current PR: status bar layout and background:
- Margin (outer spacing via Wayland layer-shell) and padding (inner content spacing)
- Background color:
  - Transparent: no bar background, modules get individual containers (islands look)
  - Solid: uses the background color from the palette. I'd like to keep the easy-to-set-up philosophy. Allowing a fully custom background color also means the user would need to set up a matching foreground color, which goes against the palette system. I'm skeptical about adding that.

## Future PR: status bar border:
- Roundness: custom value, or predefined values from the theme?
- Color: same concern as background: maybe it's better to always derive from the palette (e.g., a darker background variant, secondary, or primary)?
- Width: custom value, or the one from the theme?

## Future PR: module styling:
- Border?
- Background color?
- Roundness?

# General concern:

More options mean more complexity, more bugs, and more effort to maintain the codebase. Giving the user too many choices could also go against the project's original philosophy.

My original idea with the style enum (Islands/Solid/Gradient) was to offer a preset of curated style solutions. Maybe that's still the right approach instead of exposing individual properties. But I found the presets too limiting.

We should find a good balance. Any contribution to this discussion is welcome.

clotodex:
I dont have much opinion on the "how" but agree with the limited out of the box options

I am not sure if a "style override" could not be supported anyways.
After all we are exposing the different values for rhe theme options anyways so i would give the option to fiddle around with them

I would keep the theme optimized for what we are supporting and potentially have a "collection" issue that helps with collecting all custom requests etc so we know what we want or dont want to support

Finally - why did you find the original config limiting? To your own customization or to the users'?

MalpenZibo:
Personally the project start to create my ideal status bar. For this reason I created the island style (at the time it wasn't a style and the whole concept of style doesn't exist).

Then in a fork an user create a status bar with a background color.

After that I decided to add these two style in the official repo + the gradient style because why not (bad idea).

Then the opacity and backdrop config as been added.

Now from my point of view we can remain in this situation I personally don't need other configuration. 

I would like to understand what an user would like to be able to change and comes up with some new settings that replace the style system letting the user more freedom without making the entire configuration too complex.

[...]

PR #798, opened 2026-06-07
feat: expose radius to config

magnus-rattlehead:
Wanted to expose the radius options to config, personally I prefer no rounded corners

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -888,6 +888,26 @@ impl Default for MenuAppearance {
     }
 }
 
+#[derive(Deserialize, Clone, Copy, Debug)]
+#[serde(default)]
+pub struct Radius {
+    pub sm: f32,
+    pub md: f32,
+    pub lg: f32,
+    pub xl: f32,
+}
+
+impl Default for Radius {
+    fn default() -> Self {
+        Self {
+            sm: 4.0,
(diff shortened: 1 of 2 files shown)


romanstingler:
let's wait for
@MalpenZibo 's opinion 

from a technical standpoint, add validation rejecting negative radius values via a custom deserializer like opacity_deserializer. 
Add `[appearance.radius]` to the example config in `docs/src/core/config-system.md` (pre-existing).

```
[appearance.radius]
sm = 4.0
md = 8.0
lg = 16.0
xl = 32.0
```

also @MalpenZibo stylisitc question
Field-level `#[serde(default)]` could be used instead of struct-level. Struct-level `#[serde(default)]` is fine and more concise, but the rest of `Appearance` uses field-level attrs (e.g. `pub menu: MenuAppearance` relies on `MenuAppearance`'s own `Default`). Either style is acceptable here since `Radius` has a `Default` impl

MalpenZibo:
Yes, I need to work on this since the last release. I had to decide the right direction. I prefer to work on this stuff personally, so I'm not sure if I will merge this PR. For sure, I will take this into consideration when defining the new option that I plan to add. 

https://github.com/MalpenZibo/ashell/issues/615

PR #801, opened 2026-06-08
fix(power): handle pending battery states

Scott-Nx:
## Summary

Fixes a follow-up from #779 where batteries held by charge limit could still show as `Discharging` with `Calculating...` in the power tooltip.

UPower may report this held state as `PendingCharge` / `PendingDischarge`. These states are now handled explicitly as `NotCharging` instead of falling back to fake `Discharging(0)`.

## Improvements

- Shows `Not charging` instead of misleading `Discharging` for charge-limit held batteries.
- Removes fake `Calculating...` ETA for states that are not actively charging or discharging.
- Adds explicit handling for `PendingCharge`, `PendingDischarge`, `Unknown`, and `Empty`.
- Keeps low-battery danger styling limited to real discharging batteries.
- Keeps #779 top-bar charge-limit icon behavior unchanged.
- Adds regression tests for pending, unknown, timed, and icon fallback behavior.

## Behavior changes

| Scenario | Before | After |
|---|---|---|
| Charge-limit held battery | `Discharging` + `Calculating...` | `Not charging`, no ETA |
| `PendingCharge` / `PendingDischarge` | `Discharging(0)` | `NotCharging` |
| `Unknown` / `Empty` | `Discharging(0)` | `Unknown` |
| Low-battery danger | Could apply to fallback discharging states | Only applies to real `Discharging` |
| Charging / Discharging / Full | Existing behavior | Unchanged |
| #779 top-bar charge-limit icon | Existing behavior | Unchanged |

## Validation

- [x] `cargo fmt`
- [x] `cargo test`
- [x] `cargo clippy --all-targets --all-features -- -D warnings`
- [x] `make check`

### Before
screenshot_2569-06-08_15-40-26

### After
screenshot_2569-06-08_16-27-16

### DIFF
--- a/i18n/en-US/ashell.ftl
+++ b/i18n/en-US/ashell.ftl
@@ -87,6 +87,8 @@ settings-power-profile-power-saver = Power Saver
 
 settings-power-status-charging = Charging
 settings-power-status-discharging = Discharging
+settings-power-status-not-charging = Not charging
+settings-power-status-unknown = Unknown
 settings-power-status-full = Full
 
 ## Settings — idle inhibitor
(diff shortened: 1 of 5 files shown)


MalpenZibo (approved):
👌

PR #802, opened 2026-06-08
remove clip from module item to fix cropped icon

### LINKED ISSUE #571: [Bug] Update module icon is cropped
Like this:
Image

romanstingler:
Removing `.clip(true)` from module item containers prevents icons from being cropped when they extend beyond element bounds.

Before
image
After
image

With `vk` and `gl` renderer 
with `ICED_BACKEND=tiny-skia` is always renders fine on my machine.

sorry for the small screenshots.

closes https://github.com/MalpenZibo/ashell/issues/571

### DIFF
--- a/src/components/module_item.rs
+++ b/src/components/module_item.rs
@@ -65,8 +65,7 @@ impl<'a, Msg: 'static + Clone> From<ModuleItem<'a, Msg>> for Element<'a, Msg> {
             let mut button = position_button(
                 container(item.content)
                     .align_y(Alignment::Center)
-                    .height(Length::Fill)
-                    .clip(true),
+                    .height(Length::Fill),
             )
             .padding([2.0, space.xs])
             .height(Length::Fill)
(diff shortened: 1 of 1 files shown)


MalpenZibo:
Hmm! This makes sense, but:
- I suspect that the problem is in something like "how cosmic-text" calculates the width of some glyph
- This causes a secondary issue: if things like the window title are too long (something happens to me), the text goes on a new line. So, to resolve an issue that I never noticed on my machine (it seems to depend on some machine configuration), we're causing an issue that will happen everywhere

romanstingler:
what?? 
" if things like the window title are too long (something happens to me), the text goes on a new line"

there are dots in the middle and the text is shortened. (top latest main, bottom branch)
image

MalpenZibo:
image
This is what happens to me when I fill the space. I can reduce the maximum number of title characters, but if I set that value too high, the clip property truncates the end of the title. Without it, the text will go to the next line.

I will try to make more test about this

PR #805, opened 2026-06-09
workspaces: compare monitor names by equality, not substring

### LINKED ISSUE #325: `visibility_mode = "MonitorSpecific"` has no effect
Since #312, my workspace widget always shows all workspaces on all monitors, even with

```toml
[workspaces]
visibility_mode = "MonitorSpecific"
```

This is my config:

<details>
<summary>ashell.toml</summary>

```toml
enable_esc_key = true
log_level = "warn,ashell::services::mpris=trace,ashell::modules::media_player=trace"

[appearance]
font_name = "Meslo LG S DZ"

[clock]
format = "%F %H:%M:%S"

[modules]
center = ["MediaPlayer"]
left = ["Workspaces", "KeyboardSubmap"]
right = ["Privacy", "SystemInfo", "Tray", ["Settings", "Clock"]]

[settings]
audio_sinks_more_cmd = "pavucontrol -t 3"
audio_sources_more_cmd = "pavucontrol -t 4"
bluetooth_more_cmd = "overskride"
vpn_more_cmd = "nm-connection-editor"
wifi_more_cmd = "nm-connection-editor"

[system_info]
indicators = ["Cpu", "Memory", "Temperature"]

[system_info.cpu]
alert_threshold = 80
warn_threshold = 60

[system_info.memory]
alert_threshold = 80
warn_threshold = 60

[workspaces]
visibility_mode = "MonitorSpecific"
```

</details>

and these are my outputs, in case it's relevant:

<details>

```sh
$ hyprctl monitors
Monitor DP-1 (ID 1):
	3440x1440@59.99900 at 400x1080
	description: Beihai Century Joint Innovation Technology Co.Ltd UWQHD-100-C 0000000000000
	make: Beihai Century Joint Innovation Technology Co.,Ltd
	model: UWQHD-100-C
	physical size (mm): 800x340
	serial: 0000000000000
	active workspace: 2 (2)
	special workspace: 0 ()
	reserved: 0 34 0 0
	scale: 1.00
	transform: 0
	focused: no
	dpmsStatus: 1
	vrr: false
	solitary: 0
	solitaryBlockedBy: windowed mode,missing candidate
	activelyTearing: false
	tearingBlockedBy: next frame is not torn,user settings,missing candidate
	directScanoutTo: 0
	directScanoutBlockedBy: user settings,missing candidate
	disabled: false
	currentFormat: XRGB8888
	mirrorOf: none
	availableModes: 3440x1440@60.00Hz 3440x1440@100.00Hz 1720x1440@60.00Hz 1920x1080@119.88Hz 1920x1080@60.00Hz 1920x1080@59.94Hz 1920x1080@50.00Hz 1600x1200@60.00Hz 1680x1050@59.95Hz 1280x1024@75.03Hz 1280x1024@60.02Hz 1440x900@59.89Hz 1280x960@60.00Hz 1152x864@75.00Hz 1280x720@60.00Hz 1280x720@59.94Hz 1280x720@50.00Hz 1024x768@75.03Hz 1024x768@70.07Hz 1024x768@60.00Hz 800x600@75.00Hz 800x600@72.19Hz 800x600@60.32Hz 800x600@56.25Hz 720x576@50.00Hz 720x480@59.94Hz 640x480@75.00Hz 640x480@72.81Hz 640x480@59.94Hz 640x480@59.93Hz 
	colorManagementPreset: srgb
	sdrBrightness: 1.00
	sdrSaturation: 1.00
	sdrMinLuminance: 0.20
	sdrMaxLuminance: 80

Monitor HDMI-A-1 (ID 0):
	1920x1080@60.00000 at 0x0
	description: Ancor Communications Inc ASUS VC239 F7LMTJ005720
	make: Ancor Communications Inc
	model: ASUS VC239
	physical size (mm): 510x290
	serial: F7LMTJ005720
	active workspace: 1 (1)
	special workspace: 0 ()
	reserved: 0 34 0 0
	scale: 1.00
	transform: 0
	focused: no
	dpmsStatus: 1
	vrr: false
	solitary: 0
	solitaryBlockedBy: windowed mode,missing candidate
	activelyTearing: false
	tearingBlockedBy: next frame is not torn,user settings,missing candidate
	directScanoutTo: 0
	directScanoutBlockedBy: user settings,missing candidate
	disabled: false
	currentFormat: XRGB8888
	mirrorOf: none
	availableModes: 1920x1080@60.00Hz 1920x1080@59.94Hz 1920x1080@50.00Hz 1680x1050@59.95Hz 1280x1024@75.03Hz 1280x1024@60.02Hz 1440x900@59.89Hz 1280x960@60.00Hz 1152x864@75.00Hz 1280x720@60.00Hz 1280x720@59.94Hz 1280x720@50.00Hz 1440x576@50.00Hz 1024x768@75.03Hz 1024x768@70.07Hz 1024x768@60.00Hz 1440x480@59.94Hz 800x600@75.00Hz 800x600@72.19Hz 800x600@60.32Hz 800x600@56.25Hz 720x576@50.00Hz 720x480@59.94Hz 640x480@75.00Hz 640x480@72.81Hz 640x480@59.94Hz 640x480@59.93Hz 
	colorManagementPreset: srgb
	sdrBrightness: 1.00
	sdrSaturation: 1.00
	sdrMinLuminance: 0.20
	sdrMaxLuminance: 80

Monitor DP-2 (ID 2):
	1920x1080@60.00000 at 1920x0
	description: Ancor Communications Inc ASUS PB238 DBLMTF249642
	make: Ancor Communications Inc
	model: ASUS PB238
	physical size (mm): 510x290
	serial: DBLMTF249642
	active workspace: 3 (3)
	special workspace: 0 ()
	reserved: 0 34 0 0
	scale: 1.00
	transform: 0
	focused: yes
	dpmsStatus: 1
	vrr: false
	solitary: 0
	solitaryBlockedBy: windowed mode,missing candidate
	activelyTearing: false
	tearingBlockedBy: next frame is not torn,user settings,missing candidate
	directScanoutTo: 0
	directScanoutBlockedBy: user settings,missing candidate
	disabled: false
	currentFormat: XRGB8888
	mirrorOf: none
	availableModes: 1920x1080@60.00Hz 1680x1050@59.95Hz 1280x1024@75.03Hz 1280x1024@60.02Hz 1440x900@74.98Hz 1440x900@59.89Hz 1280x960@60.00Hz 1152x864@75.00Hz 1280x720@60.00Hz 1024x768@75.03Hz 1024x768@70.07Hz 1024x768@60.00Hz 800x600@75.00Hz 800x600@72.19Hz 800x600@60.32Hz 800x600@56.25Hz 640x480@75.00Hz 640x480@72.81Hz 640x480@59.94Hz 
	colorManagementPreset: srgb
	sdrBrightness: 1.00
	sdrSaturation: 1.00
	sdrMinLuminance: 0.20
	sdrMaxLuminance: 80
```

</details>

### LINKED ISSUE #804: MonitorSpecific visibility filter breaks when output names overlap as substrings (e.g. eDP-1 / DP-1)
### What happens

With `visibility_mode = "MonitorSpecific"` (or `"MonitorSpecificExclusive"`), the bar on a monitor whose name **contains another monitor's name as a substring** stops tracking its own focused workspace. The most common pair this hits is `eDP-1` (laptop) and `DP-1` (docked DisplayPort) — `"eDP-1".contains("DP-1")` is true, so the laptop's bar accepts every workspace owned by `DP-1` as if it were its own.

Visible symptoms on the affected bar:
- Workspaces from the other monitor are rendered alongside its own.
- The active highlight stops moving when you switch workspaces locally on that monitor.

The unaffected bar (the one whose name is the substring, e.g. `DP-1`) renders correctly because `"DP-1".contains("eDP-1")` is false.

### Repro

Hyprland with two outputs whose names are substring-overlapping. Easiest setup is any laptop with a docked DisplayPort monitor (outputs `eDP-1` + `DP-1`).

```toml
[workspaces]
visibility_mode = "MonitorSpecific"
```

Open windows on both monitors so each has at least one workspace, focus the laptop, press Super+1/2/3… — the laptop's bar highlight does not follow.

### Root cause

`src/modules/workspaces.rs:445-453` (introduced in #331, merged 2025-11-26):

```rust
WorkspaceVisibilityMode::MonitorSpecific => {
    monitor_name
        .unwrap_or_else(|| &w.monitor)
        .contains(&w.monitor)
        || !outputs.has_name(&w.monitor)
}
WorkspaceVisibilityMode::MonitorSpecificExclusive => monitor_name
    .unwrap_or_else(|| &w.monitor)
    .contains(&w.monitor),
```

#331 swapped equality for `.contains()` to fix #325, which was itself a regression from #312's fuzzy output-name matching. The substring form fixes the original case but silently breaks any pair where one canonical name is a substring of the other.

By the time those values reach this filter, both `monitor_name` (`outputs.get_monitor_name(id)`) and `w.monitor` (compositor event) are canonical output names — so equality is the correct relation. The fuzzy matching from #312 belongs in the config→output resolution path, not in the per-bar visibility filter where both sides are already resolved.

### Versions

- ashell 0.8.0 (Arch package via paru) and current `main` — both affected (the offending code is unchanged on main since #331).
- Hyprland 0.51.x (any version with the standard `eDP-1`/`DP-N` naming).

johnnybravo-xyz:
Fixes #804.

### What this changes

`WorkspaceVisibilityMode::MonitorSpecific` and `MonitorSpecificExclusive` compared `monitor_name.contains(&w.monitor)` instead of `monitor_name == w.monitor`. The substring form falsely matches any pair where one canonical name is a substring of the other (`eDP-1`/`DP-1` being the common case for laptops with a docked DisplayPort monitor).

This PR restores equality. Both sides of the comparison are canonical compositor output names by the time they reach this filter, so equality is the correct relation.

### Why not just keep `.contains()`?

#331 introduced `.contains()` to fix #325, which was a regression from #312's fuzzy output-name matching. But the fuzzy matching belongs in the config→output resolution path — once `w.monitor` and `monitor_name` are populated from compositor events, they're already canonical names. Filtering them with substring semantics ends up matching pairs the user never asked to alias.

If a future change re-introduces a config-side alias that doesn't get resolved before this filter sees it, a dedicated alias-resolution step (rather than blanket substring matching) is the right place to handle it.

### Repro for verification

Any host with two outputs whose names are substring-overlapping (`eDP-1` + `DP-1`, `HDMI-1` + `DMI-1`, etc.) and `visibility_mode = "MonitorSpecific"`. Before this patch, the bar on the *longer* name's monitor shows the other monitor's workspaces and stops tracking its own focused workspace. After this patch, each bar renders only its own monitor's workspaces.

### DIFF
--- a/src/outputs.rs
+++ b/src/outputs.rs
@@ -64,9 +64,22 @@ impl OverlaySurface {
     }
 }
 
+/// Pair of strings identifying an output. `name` is the canonical
+/// short name reported by the compositor (e.g. `eDP-1`) — used for
+/// equality checks against workspace events. `description` includes
+/// the EDID (e.g. `eDP-1 Make Model Serial`) — used for the fuzzy
+/// substring matching in `name_in_config` / `has_name` so users can
+/// alias outputs by any string that appears in the EDID (the
+/// behaviour added by #312).
+#[derive(Debug, Clone)]
+pub struct OutputKey {
+    pub name: String,
+    pub description: String,
+}
+
 #[derive(Debug, Clone)]
 pub struct Outputs {
-    entries: Vec<(String, Option<ShellInfo>, Option<OutputId>)>,
+    entries: Vec<(OutputKey, Option<ShellInfo>, Option<OutputId>)>,
     toast: Option<OverlaySurface>,
     osd: Option<OverlaySurface>,
     /// Last toast input-region request, replayed once the toast's output is
@@ -82,7 +95,7 @@ pub enum HasOutput<'a> {
 }
 
 impl Outputs {
-    pub fn iter(&self) -> std::slice::Iter<'_, (String, Option<ShellInfo>, Option<OutputId>)> {
+    pub fn iter(&self) -> std::slice::Iter<'_, (OutputKey, Option<ShellInfo>, Option<OutputId>)> {
         self.entries.iter()
     }
 
@@ -157,13 +170,23 @@ impl Outputs {
         (id, main_task)
     }
 
-    fn name_in_config(name: &str, outputs: &config::Outputs) -> bool {
+    /// Match a user-supplied output spec against an output's name +
+    /// description pair. Returns true when:
+    ///   - the spec equals the canonical name (`info.name`), OR
+    ///   - the spec appears anywhere in the description (`info.name +
+    ///     info.make + info.model`), preserving #312's fuzzy alias
+    ///     matching by EDID substring.
+    fn matches_spec(name: &str, description: &str, spec: &str) -> bool {
+        name == spec || description.contains(spec)
+    }
+
+    fn name_in_config(name: &str, description: &str, outputs: &config::Outputs) -> bool {
         match outputs {
             config::Outputs::All => true,
             config::Outputs::Active => false,
-            config::Outputs::Targets(request_outputs) => {
-                request_outputs.iter().any(|output| name.contains(output))
-            }
+            config::Outputs::Targets(request_outputs) => request_outputs
+                .iter()
+                .any(|spec| Self::matches_spec(name, description, spec)),
         }
     }
 
@@ -187,11 +210,13 @@ impl Outputs {
         })
     }
 
+    /// Returns the canonical short name (e.g. `eDP-1`) — used by the
+    /// workspace visibility filter which compares against `w.monitor`.
     pub fn get_monitor_name(&self, id: SurfaceId) -> Option<&str> {
-        self.entries.iter().find_map(|(name, info, _)| {
+        self.entries.iter().find_map(|(key, info, _)| {
             info.as_ref().and_then(|info| {
                 if info.id == id {
-                    Some(name.as_str())
+                    Some(key.name.as_str())
                 } else {
                     None
                 }
@@ -200,9 +225,11 @@ impl Outputs {
     }
 
     pub fn has_name(&self, name: &str) -> bool {
-        self.entries
-            .iter()
-            .any(|(n, info, _)| info.is_some() && n.as_str().contains(name))
+        // Match by canonical name (equality) OR description (substring,
+        // to preserve #312's EDID-alias behaviour).
+        self.entries.iter().any(|(key, info, _)| {
+            info.is_some() && Self::matches_spec(&key.name, &key.description, name)
+        })
     }
 
     #[allow(clippy::too_many_arguments)]
@@ -213,10 +240,11 @@ impl Outputs {
         position: Position,
         layer: config::Layer,
         name: &str,
+        description: &str,
         output_id: OutputId,
         scale_factor: f64,
     ) -> Task<Message> {
-        let target = Self::name_in_config(name, request_outputs);
+        let target = Self::name_in_config(name, description, request_outputs);
 
         if target {
             debug!("Found target output, creating a new layer surface");
@@ -224,10 +252,14 @@ impl Outputs {
             let (id, task) =
                 Self::create_output_layers(style, Some(output_id), position, layer, scale_factor);
 
+            // Replace an existing entry with the same canonical name
+            // (e.g. monitor was reconnected). Description-match is
+            // intentionally NOT used here — replacement is identity-
+            // based, not alias-based.
             let destroy_task = match self
                 .entries
                 .iter()
-                .position(|(key, _, _)| key.as_str() == name)
+                .position(|(key, _, _)| key.name.as_str() == name)
             {
                 Some(index) => {
                     let old_output = self.entries.swap_remove(index);
@@ -241,7 +273,10 @@ impl Outputs {
             };
 
             self.entries.push((
-                name.to_owned(),
+                OutputKey {
+                    name: name.to_owned(),
+                    description: description.to_owned(),
+                },
                 Some(ShellInfo {
                     id,
                     menu: Menu::new(),
@@ -278,7 +313,14 @@ impl Outputs {
                 name, request_outputs
             );
 
-            self.entries.push((name.to_owned(), None, Some(output_id)));
+            self.entries.push((
+                OutputKey {
+                    name: name.to_owned(),
+                    description: description.to_owned(),
+                },
+                None,
+                Some(output_id),
+            ));
 
             Task::none()
         }
@@ -327,7 +369,10 @@ impl Outputs {
                         Self::create_output_layers(style, None, position, layer, scale_factor);
 
                     self.entries.push((
-                        "Fallback".to_string(),
+                        OutputKey {
+                            name: "Fallback".to_string(),
+                            description: String::new(),
+                        },
                         Some(ShellInfo {
                             id,
                             menu: Menu::new(),
@@ -360,8 +405,10 @@ impl Outputs {
         let to_remove = self
             .entries
             .iter()
-            .filter_map(|(name, shell_info, output_id)| {
-                if !Self::name_in_config(name, request_outputs) && shell_info.is_some() {
+            .filter_map(|(key, shell_info, output_id)| {
+                if !Self::name_in_config(&key.name, &key.description, request_outputs)
+                    && shell_info.is_some()
+                {
                     *output_id
                 } else {
                     None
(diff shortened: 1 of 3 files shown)


Shadesfear:
I tested this locally on Niri. The equality-only change fixes the `eDP-1`/`DP-1` false positive, but it breaks workspace
  rendering for me: no workspaces are shown.

  The reason is that the two strings being compared are not the same shape on Niri.

  Niri IPC reports workspace outputs as canonical names:

  ```text
  DP-1
  DP-3
  eDP-1
  ```
  
  But the output string ashell gets for the layer-shell surface is the full output description:

  ```text
  eDP-1 China Star Optoelectronics Technology Co., Ltd MNE007JA1-3
  DP-1 Lenovo Group Limited LEN T27h-2L
  DP-3 Lenovo Group Limited LEN T27h-2L
  ```
 
 the name is stored as 
 
 ```rust
 let name = &format!("{} {} {}", info.name, info.make, info.model);
 ```
 
 So I think the right move is to just use info.name for matching

johnnybravo-xyz:
Thanks @Shadesfear — pushed `ce09e455` that drops the make/model concatenation in `OutputEvent::Added` so the layer-shell surface key is just `info.name`, matching how workspaces report their owning monitor on both Hyprland and Niri.

One trade-off worth flagging for @MalpenZibo: this also changes the string that `name_in_config` matches against. If a user's `outputs = ["Serial123"]` config was relying on the substring landing inside the concatenated `"eDP-1 Make Model Serial123"` (#312's fuzzy alias path), they'll need to switch to using the canonical output name. A more defensive fix would store both `name` and `description` separately and have `name_in_config` consult either — happy to follow that up in a separate PR if you'd prefer.

MalpenZibo:
I prefer not to break the current behavior. So maybe we had to add the `name` and `description` in this PR to avoid any regression

johnnybravo-xyz:
Pushed `2681ef13` per your suggestion. Each entry in `outputs.rs` now stores an `OutputKey { name, description }` pair:

- `name`        – the canonical compositor string (`eDP-1`). Used by the workspace visibility filter and `get_monitor_name`. Equality only.
- `description` – the full `name make model` concat. Used by `name_in_config` / `has_name` so configs from #312 (`outputs = ["SerialFoo"]`, alias by any EDID substring) keep working with no behaviour change.

A single `matches_spec(name, description, spec)` predicate is the source of truth – true when the spec equals the canonical name OR appears as a substring of the description. Both `name_in_config` and `has_name` route through it. `Outputs::add` signature gained a `description: &str`; the only call site (`app.rs OutputEvent::Added`) passes `info.name` + `format!("{} {} {}", info.name, info.make, info.model)`. The "Fallback" pseudo-entry gets an empty description (never participates in user config matching).

src/outputs.rs
```diff
@@ -200,9 +225,11 @@ impl Outputs {
     }
 
     pub fn has_name(&self, name: &str) -> bool {
```
MalpenZibo, line 237: Sorry. I'm a bit confused but:
The `has_name` is used by the workspace module and uses the, let's say, "precise name" so doesn't need any fuzzy match and could use the equality check against the `key.name`.

The `name_in_config` instead is what needs some fuzzy matching so cannot use the equality check but need to use the `name == spec || description.contains(spec)`.

So probably we do not use the same `Self::matches_spec` call in both cases.

I'm sorry if my previous review comment confuses you and causes this "error"

MalpenZibo (changes requested):

johnnybravo-xyz:
Thanks for clarifying that — pushed `763c059e`. `has_name` now uses strict `key.name == name` equality only. `matches_spec` stays but is called only from `name_in_config` (the fuzzy/config-alias path). Also updated the doc comment to point at why the two paths diverge.

src/outputs.rs
```diff
@@ -64,9 +64,22 @@ impl OverlaySurface {
     }
 }
 
+/// Pair of strings identifying an output. `name` is the canonical
+/// short name reported by the compositor (e.g. `eDP-1`) — used for
+/// equality checks against workspace events. `description` includes
+/// the EDID (e.g. `eDP-1 Make Model Serial`) — used for the fuzzy
+/// substring matching in `name_in_config` / `has_name` so users can
+/// alias outputs by any string that appears in the EDID (the
+/// behaviour added by #312).
```
MalpenZibo: This comment is now out of date. It still lists `has_name` as a consumer of the fuzzy description matching

MalpenZibo (commented):
A small fix and the format CI with an error, then we can merge

johnnybravo-xyz:
Fixed the stale docstring — pushed the update. `has_name` is now correctly described as strict-equality only, with the fuzzy path documented as `name_in_config`-only.

MalpenZibo:
There's still the CI problem

PR #810, opened 2026-06-11
warn and auto-select weight when font_name has no Normal(400) face

### LINKED ISSUE #768: Terminus fonts not working
For some weird reason, on Arch neither bitmap Terminus fonts from `terminus-font` package nor AUR `terminus-font-ttf` are able to be used by ashell. Running through the terminal doesn't throw any errors.
Putting `font-name="Terminus"` or `font-name="Terminus (TTF)"` in the config file doesn't do anything.
Is this even possible to implement? Maybe inputting a direct path to the font could be a solution?

romanstingler:
`ashell` silently falls back to a different font when the configured `font_name` exists in the system font database but has no face with `weight=Normal(400)` (e.g. Terminus TTF, whose Regular face reports `weight=500/Medium` in its OS/2 table). This happens because cosmic-text's font matcher re-scores all faces and prefers the one whose weight is closest to the requested `Normal(400)`, even if that means picking a font from a different family.

https://docs.rs/fontdb/0.23.0/src/fontdb/lib.rs.html#986
https://docs.rs/iced_core/0.14.0/src/iced_core/font.rs.html#74

**Changes:**
Add `fontdb` as a direct dependency (same version already used transitively by `iced_layershell`)(@MalpenZibo no idea i am really weighing if it is OK to keep it here or to put it into `iced_layershell`, but I think maybe you want to keep `iced_layershell` as slim as possible).

Add `resolve_font()` which:
- Finds all faces matching the configured font name.
- Picks the closest match to Normal style/stretch and weight 400.
- Builds an `iced::Font` with that face's actual weight and style, so cosmic-text's matcher stays on the right font.
- Warns clearly when no `Normal(400)` face is available.
- Warn when the font name isn't found at all (with a suggestion to use fc-list).
Image
image

Fixes #768
Image

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -88,6 +88,7 @@ url = "2.5.7"
 signal-hook = "0.4.3"
 signal-hook-tokio = { version = "0.4", features = ["futures-v0_3"] }
 libc = "0.2.182"
+fontdb = { version = "0.23", features = ["fontconfig"] }
 chrono-tz = "0.10.4"
 wayland-client = "0.31"
 wayland-protocols = { version = "0.32", features = ["client", "staging"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -88,6 +88,7 @@ url = "2.5.7"
 signal-hook = "0.4.3"
 signal-hook-tokio = { version = "0.4", features = ["futures-v0_3"] }
 libc = "0.2.182"
+# keep in sync with the fontdb version pulled transitively by iced_layershell
 fontdb = { version = "0.23", features = ["fontconfig"] }
 chrono-tz = "0.10.4"
 wayland-client = "0.31"
--- a/src/main.rs
+++ b/src/main.rs
@@ -6,7 +6,10 @@ use clap::Parser;
 use flexi_logger::{
     Age, Cleanup, Criterion, FileSpec, LogSpecBuilder, LogSpecification, Logger, Naming,
 };
-use iced::{Anchor, Font, KeyboardInteractivity, Layer, LayerShellSettings};
+use iced::{
+    Anchor, Font, KeyboardInteractivity, Layer, LayerShellSettings,
+    font::{Style as FontStyle, Weight as FontWeight},
+};
 use log::{debug, error, warn};
 use std::backtrace::Backtrace;
 use std::env;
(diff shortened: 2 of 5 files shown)


MalpenZibo:
Yep, it's something a little bit borderline and maybe could be placed in `iced_layershell`. I need to think about that

src/main.rs
```diff
@@ -67,6 +70,98 @@ fn get_log_spec(log_level: &str) -> LogSpecification {
     }
 }
 
+fn fontdb_weight_to_iced(weight: fontdb::Weight) -> FontWeight {
```
MalpenZibo, line 74: make sense to also add something like
```
fn fontdb_stretch_to_iced(stretch: fontdb::Stretch) -> iced::font::Stretch {
    use iced::font::Stretch;
    match stretch {
        fontdb::Stretch::UltraCondensed => Stretch::UltraCondensed,
        fontdb::Stretch::ExtraCondensed => Stretch::ExtraCondensed,
        fontdb::Stretch::Condensed => Stretch::Condensed,
        fontdb::Stretch::SemiCondensed => Stretch::SemiCondensed,
        fontdb::Stretch::Normal => Stretch::Normal,
        fontdb::Stretch::SemiExpanded => Stretch::SemiExpanded,
        fontdb::Stretch::Expanded => Stretch::Expanded,
        fontdb::Stretch::ExtraExpanded => Stretch::ExtraExpanded,
        fontdb::Stretch::UltraExpanded => Stretch::UltraExpanded,
    }
}
```

src/main.rs
```diff
@@ -67,6 +70,98 @@ fn get_log_spec(log_level: &str) -> LogSpecification {
...
+
+    Font {
+        family: iced::font::Family::Name(Box::leak(name.to_string().into_boxed_str())),
+        weight: iced_weight,
+        stretch: iced::font::Stretch::Normal,
+        style: if face.style == fontdb::Style::Italic {
+            FontStyle::Italic
+        } else {
+            FontStyle::Normal
+        },
```
MalpenZibo: We are not considering the `Oblique` case

src/main.rs
```diff
@@ -67,6 +70,98 @@ fn get_log_spec(log_level: &str) -> LogSpecification {
...
+            style_penalty + stretch_penalty + weight_penalty
+        });
+
+    if best_face.is_none() {
+        warn!("Font '{}' was not found in the system font database.", name);
+        warn!("  Use `fc-list` to list all available fonts and their exact family names.");
+        return Font::with_name(Box::leak(name.to_string().into_boxed_str()));
+    }
+
+    let face = best_face.unwrap();
```
MalpenZibo: can we rewrite this to avoid the `best_face.unwrap`?

src/main.rs
```diff
@@ -67,6 +70,98 @@ fn get_log_spec(log_level: &str) -> LogSpecification {
...
+    } else if numeric_weight <= 750 {
+        FontWeight::Bold
+    } else if numeric_weight <= 850 {
+        FontWeight::ExtraBold
+    } else {
+        FontWeight::Black
+    }
+}
+
+fn resolve_font(name: &str) -> Font {
```
MalpenZibo, line 120: Let's add a comment like
```
    // NOTE: workaround for iced-rs/iced#2847 — iced 0.14 doesn't expose face
    // selection, so cosmic-text's matcher can silently jump to another family
    // when the requested weight/stretch/style has no exact face. We pre-resolve
    // the real face here and pass its actual attributes so the matcher stays put.
    // Remove once iced exposes face selection. Loads the system font DB a second
    // time (cosmic-text already loads it internally); acceptable as a one-shot
    // startup cost paid only when `font_name` is set.
```

Cargo.toml
```diff
@@ -88,6 +88,7 @@ url = "2.5.7"
 signal-hook = "0.4.3"
 signal-hook-tokio = { version = "0.4", features = ["futures-v0_3"] }
 libc = "0.2.182"
+fontdb = { version = "0.23", features = ["fontconfig"] }
```
MalpenZibo, line 92: Let's add this comment
```
# keep in sync with the fontdb version pulled transitively by iced_layershell
```

romanstingler:
Pushed the reviewed suggestions

MalpenZibo (approved):

PR #812, opened 2026-06-12
add Name/IconAndName formats to show network name in indicator

### LINKED ISSUE #515: [MODULES] feature request: display SSID
Hi,

I currently have a CustomModule which shows the SSID of the connected WiFi network, like this:

screenshot

I'm working on improving the code for this but realized that so much of the functionality (watching NetworkManger's D-Bus API) overlaps with the code that is already in ashell. Could we have this as a built-in feature of ashell?

I am willing to contribute. If you think this is a good idea: we should also decide: should this be a part of the "Settings" module or should it be a standalone module? What should the configuration schema be?

romanstingler:
Adds 2 config options `Name` and `IconAndName` to the existing `SettingsFormat` enum.
This lets users display the connected network's name (the SSID for WiFi, the interface name for wired) in the top-bar network indicator instead of the signal-strength percentage.

```
[settings]
network_indicator_format = "IconAndName" 
# or
network_indicator_format = "Name" 
```

image

closes https://github.com/MalpenZibo/ashell/issues/515

### DIFF
--- a/src/components/format_indicator.rs
+++ b/src/components/format_indicator.rs
@@ -50,13 +50,15 @@ impl<'a, Msg: 'static + Clone> From<FormatIndicator<'a, Msg>> for Element<'a, Ms
 
         let content = match fi.format {
             SettingsFormat::Icon => fi.icon.to_text().into(),
-            SettingsFormat::Percentage | SettingsFormat::Time => fi.label_element,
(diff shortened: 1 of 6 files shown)


MalpenZibo (approved):
👍

PR #813, opened 2026-06-12
patch hyprland-rs for non-UTF-8 window titles

romanstingler:
closes https://github.com/MalpenZibo/ashell/issues/577

I pushed an upstream PR to `hyprland-rs`

https://github.com/hyprland-community/hyprland-rs/pull/395

where I added a module that detects the most likely encoding of the byte stream using
`chardetng` (the same library Firefox uses) and decodes with `encoding_rs`. 
Worst case it falls back to the current behavior.


MalpenZibo:
Did you run `cargo update` from the latest main branch? I see some strange packages downgraded in the cargo lock file

romanstingler:
I did a 
`cargo update -p hyprland`

MalpenZibo (approved):

PR #814, opened 2026-06-13
Add configuration option to show active workspace in a different color

1randomguy:
Currently the only indication of the active workspace is a slightly bigger amount of padding around the name or number.
However, especially when you have workspace labels that are not all just one character in length it becomes very hard to see which workspace is currently active. (This might be the case if when #495 is mergend and used, but can also already happen for example when using named workspaces on niri)

Therefore, as previously pushed for in #495 , I would like to add a new config option that allow the configuration of differing color(s) for active workspaces, making it easy to spot the active workspace with one glance at the bar.

### DIFF
--- a/docs/src/core/theme-system.md
+++ b/docs/src/core/theme-system.md
@@ -16,6 +16,7 @@ pub struct AshellTheme {
     pub menu: MenuAppearance,                                 // Menu-specific styling
     pub workspace_colors: Vec<AppearanceColor>,               // Per-workspace color cycling
     pub special_workspace_colors: Option<Vec<AppearanceColor>>, // Special workspace colors
+    pub active_workspace_colors: Option<Vec<AppearanceColor>>, // Active workspace colors
     pub scale_factor: f64,                                    // DPI scale factor
 }
 ```
--- a/docs/src/reference/config-reference.md
+++ b/docs/src/reference/config-reference.md
@@ -69,6 +69,7 @@ backdrop_blur = true
 [appearance]
 workspace_colors = ["#cba6f7", "#f38ba8", "#a6e3a1", "#89b4fa"]
 special_workspace_colors = ["#fab387"]
+active_workspace_colors = ["#fa9f42"]
 ```
 
 ## Updates Module
--- a/src/config.rs
+++ b/src/config.rs
@@ -903,6 +903,7 @@ pub struct Appearance {
     pub danger_color: AppearanceColor,
     pub text_color: AppearanceColor,
     pub workspace_colors: Vec<AppearanceColor>,
+    pub active_workspace_colors: Option<Vec<AppearanceColor>>,
     pub special_workspace_colors: Option<Vec<AppearanceColor>>,
 }
 
@@ -981,6 +982,7 @@ impl Default for Appearance {
                 AppearanceColor::Simple(HexColor::rgb(158, 206, 106)),
             ],
             special_workspace_colors: None,
+            active_workspace_colors: None,
         }
     }
 }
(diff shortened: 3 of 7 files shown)


1randomguy:
A small example to illustrate why I think that this is necessary:

How fast can you tell which workspace is active here?
image
image


What about now?
image
image

romanstingler:
**`website/docs/configuration/full_config.md` is not updated.**

This is the canonical reference config users copy from. It currently lists `workspace_colors` and `special_workspace_colors` at `website/docs/configuration/full_config.md:174-175`:

```toml
workspace_colors = [ "#7aa2f7", "#9ece6a" ]
# special_workspace_colors = [ "#7aa2f7", "#9ece6a" ]  # (default: None, falls back to workspace_colors)
```

Add a commented line for `active_workspace_colors` alongside them, e.g.:

```toml
workspace_colors = [ "#7aa2f7", "#9ece6a" ]
# special_workspace_colors = [ "#7aa2f7", "#9ece6a" ]  # (default: None, falls back to workspace_colors)
# active_workspace_colors = [ "#9ece6a" ]              # (default: None, falls back to workspace_colors)
```



**Clarify `palette.md`'s indexing description.**

The current sentence "the first color will be used; if workspace 2 is attached to `monitorB`, the second color will be used, and so on" implies per-monitor indexing. `active_workspace_colors` shares the same indexing as `workspace_colors`, so a one-line clarification ("shares the per-monitor indexing of `workspace_colors`") would help. The same gap already exists for `workspace_colors` under `enable_virtual_desktops`, but this PR is a natural place to mention it.

**Update dev docs.**

`docs/src/reference/config-reference.md` and `docs/src/core/theme-system.md` still document only `workspace_colors` / `special_workspace_colors` and should mention `active_workspace_colors` as well.


@MalpenZibo your opinion 

**Document (or reconsider) the active-special-workspace behavior.**

With the new branch order:

```rust
if active {
    theme.active_workspace_colors.as_ref().unwrap_or(&theme.workspace_colors)
} else if w.id <= 0 {
    theme.special_workspace_colors.as_ref().unwrap_or(&theme.workspace_colors)
} else {
    &theme.workspace_colors
}
```

an active *special* workspace (`w.id <= 0`) now uses `active_workspace_colors` if set, otherwise falls straight to `workspace_colors`,  it no longer falls back to `special_workspace_colors`. That's probably the right UX, but it's a behavior change for anyone who sets both fields. 

Either:
- mention it in `palette.md`, or
- prefer the chain `active_workspace_colors → special_workspace_colors → workspace_colors` for special workspaces so they keep their type-specific color when active.

romanstingler (changes requested):
Documentation

1randomguy:
I don't use any special workspaces ever, and afaik they aren't even a thing on niri, so I don't have a good understanding of what special workspaces might be and how they would be used.
I guess I need some input from @MalpenZibo, or anyone who actually uses special workspaces, about what UX makes the most sense here.

MalpenZibo:
I don't use special workspaces, but from your example, I can easily understand the problem.

My only concern here is that for fix this issue we are implementing new configs. What about using the existing colors and relying on the iced palette feature? I mean by using `color.weak/strong`.

1randomguy:
> My only concern here is that for fix this issue we are implementing new configs. What about using the existing colors and relying on the iced palette feature? I mean by using `color.weak/strong`.

I mean I do like the idea of it, so I made a new separate PR #856 implementing that, please take a look 😃.
 
The one here with a separate color gives more freedom, making it possible to choose higher contrast options.
But I guess for the best oob experience the iced palette one is superior, and it also eliminates the problem of active special workspaces.

MalpenZibo:
Can we close this right @1randomguy?

1randomguy:
Yes, we can close it :)

Issue #765, opened 2026-05-21
Notification action hotkeys


matuck:
I am liking the notification system.  Is there a way to bind a hotkey to open the notifications panel, or to clear all notifications?

MalpenZibo:
We have a IPC system in place but these kind of message still need to be implemented.

PR #815, opened 2026-06-13
feat(notifications): add commands

JockeM:
Just switched over to using ashells notification module, and was missing a way
to invoke actions on current/active notifications.

Decided to add some basic other commands while am at it. I tried to take some
inspiration how makoctl have designed their commands. The more
opinionated/special one is probably list since it defaults to json and required
some changes to the IpcCommand handling (Request/Response). I'm not very
familiar with unix sockets so might have missed something obvious.

Thought about doing a scoped msg like `ashell msg notifications list` but tried
to follow the current naming schema. But I think this could be a better way to
handle things when the command list grow over time.

Not overly happy with how much I tocuhed, could probably revisit and keep it smaller.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -4,7 +4,7 @@ use crate::{
     config::{self, AppearanceStyle, Config, Modules, Position},
     get_log_spec,
     i18n::{Localizer, init_localizer},
-    ipc::IpcCommand,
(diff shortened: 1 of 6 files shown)


JockeM:
Partly fixes #765

PR #819, opened 2026-06-14
Change cpu icon to nerdfont cpu icon (from nf-md-television)

nilsmsa:
A bit subjective this one. Changes the cpu icon to the nerdfont cpu icon. More consistent with the memory icon.
from this:
image

to this:
image


MalpenZibo:
We used this https://github.com/MalpenZibo/ashell/pull/809. More or less the same icon! 

Anyway thanks!! :smile:

PR #823, opened 2026-06-16
feat(utils): add send_or_log helper and replace silent output.send calls

dacrab:
Adds a `send_or_log` utility function in `src/utils/mod.rs` that logs a warning when an iced channel send fails (receiver dropped), instead of silently discarding the error with `let _ = output.send(...)`.

Affected files:
- `src/utils/mod.rs` — new `send_or_log` helper
- `src/services/audio.rs`, `bluetooth/mod.rs`, `brightness.rs`, `logind.rs`, `mpris/mod.rs`, `network/mod.rs`, `notifications/mod.rs`, `privacy.rs`, `tray/mod.rs`, `upower/mod.rs` — replace all `let _ = output.send()` patterns
- `src/config.rs` — same replacement in config subscription

Also removes unused `SinkExt` imports from all affected service files.
No behavior changes — pure quality-of-life improvement for debugging.

### DIFF
--- a/src/services/mpris/mod.rs
+++ b/src/services/mpris/mod.rs
@@ -407,12 +407,13 @@ impl MprisPlayerService {
                                     match res {
                                         Ok(bytes) => {
                                             state_data.fetched_covers.insert(url.clone());
-                                        send_or_log(
-                                            output,
-                                            ServiceEvent::Update(
-                                                Event::CoverFetched(url.clone(), bytes)
-                                            ),
-                                        ).await;
+                                            send_or_log(
+                                                output,
+                                                ServiceEvent::Update(
+                                                    Event::CoverFetched(url.clone(), bytes),
+                                                ),
+                                            )
+                                            .await;
                                         }
                                         Err(err) => {
                                             error!("Failed to fetch cover art from {url}: {err}");
--- a/src/services/mpris/mod.rs
+++ b/src/services/mpris/mod.rs
@@ -1,10 +1,11 @@
 use super::{ReadOnlyService, Service, ServiceEvent};
+use crate::utils::send_or_log;
 use dbus::MprisPlayerProxy;
 use iced::{
     Subscription,
     core::Bytes,
     futures::{
-        FutureExt, SinkExt, Stream, StreamExt,
+        FutureExt, Stream, StreamExt,
         channel::mpsc::Sender,
         future::{BoxFuture, join_all},
         select,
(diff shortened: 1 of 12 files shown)


src/services/mpris/mod.rs
```diff
@@ -404,9 +407,12 @@ impl MprisPlayerService {
                                     match res {
                                         Ok(bytes) => {
                                             state_data.fetched_covers.insert(url.clone());
-                                            let _ = output.send(ServiceEvent::Update(
+                                        send_or_log(
```
romanstingler: indentation

romanstingler:
Maybe let's discuss that the PR introduces two equivalent call styles across the codebase. `output` is already a place expression (typically `&mut Sender<...>`), so both compile, but mixing them feels strange:

Implicit reborrow (`output`): 
```src/services/audio.rs:167,189,194,199   
src/services/bluetooth/mod.rs:185,216 
src/services/brightness.rs:149,202 
src/services/mpris/mod.rs:359,410,443 
src/services/network/mod.rs:456,505,538 
src/services/notifications/mod.rs:120,141 
src/services/privacy.rs:183,212,226 
src/services/tray/mod.rs:563,597 
src/services/upower/mod.rs:724,745
```

Explicit (`&mut output`): 
```
src/config.rs:1339,1351 
src/services/logind.rs:26,35,44,49,55
```
I'd recommend standardizing on `&mut output` explicitly, since this is a free function, and being explicit at the call site documents that we're taking a mutable borrow.

romanstingler (changes requested):

dacrab:
Regarding the `&mut output` vs `output` discussion — after investigating, the two patterns exist because of Rust's ownership rules, not style:

- **Closures** (`|mut output|` in `subscribe()`): `output` is an owned `Sender`, so `&mut output` is required to pass a mutable reference
- **Methods** (`output: &mut Sender<...>` in `start_listening`, `update_data`, etc.): `output` is already `&mut Sender`, so bare `output` works via implicit reborrow

Using `&mut output` in the method context fails to compile — you can't reborrow an already-mutable binding without `*output`. So both patterns are correct for their respective contexts.

I've fixed the indentation issue in `mpris/mod.rs` and pushed it.

MalpenZibo:
I don't think this is worth landing, though. 

A failed output.send() here isn't an "error".  The bounded Sender only fails when the receiver is dropped, which happens when iced tears the subscription down (in theory, and if it happens in other cases, it's an iced bug). That's expected lifecycle behavior.

`let _ = output.send(...).await;` is the idiomatic pattern across iced's own examples.

I'm not convinced these changes add benefit to the codebase.

dacrab:
Closing per maintainer feedback — @MalpenZibo noted that silent `let _ = output.send(...).await` is the idiomatic iced pattern and the warning log doesn't add actionable value.

PR #826, opened 2026-06-16
refactor(hyprland): unify dispatch logic and switch to std RwLock

dacrab:
Simplify the Hyprland compositor backend:

- Replace separate `dispatch_hyprlang` (sync) and `dispatch_lua` (async) functions with a single `dispatch()` function parameterized by `DispatchStrategy` enum (Socket vs Lua)
- Make `is_lua_config()` sync (uses `std::process::Command` instead of `tokio::process::Command`)
- Simplify `execute_command` to detect strategy once and call `dispatch()`
- Switch from `tokio::sync::RwLock` to `std::sync::RwLock` since lock hold times are negligible
- Use let-else / if-let patterns in event handlers for ergonomic error handling

No behavior changes.

### DIFF
--- a/src/services/compositor/hyprland.rs
+++ b/src/services/compositor/hyprland.rs
@@ -11,106 +11,125 @@ use hyprland::{
     prelude::*,
 };
 use itertools::Itertools;
-use std::sync::Arc;
-use tokio::sync::{RwLock, broadcast};
+use std::sync::{Arc, RwLock};
+use tokio::sync::broadcast;
 
-/// Detect whether Hyprland is using Lua or hyprlang config.
-/// Checks `hyprctl status` for the `configProvider` field.
-/// Falls back to hyprlang (false) if detection fails.
-async fn is_lua_config() -> bool {
-    tokio::process::Command::new("hyprctl")
+enum DispatchStrategy {
+    Socket,
+    Lua,
+}
+
+fn is_lua_config() -> bool {
+    std::process::Command::new("hyprctl")
         .arg("status")
         .output()
-        .await
         .ok()
         .and_then(|o| String::from_utf8(o.stdout).ok())
         .map(|s| s.lines().any(|l| l.starts_with("configProvider: lua")))
         .unwrap_or(false)
 }
 
-/// Dispatch a command using the old hyprlang socket protocol.
-/// Works on all Hyprland versions but is broken on 0.55+ with Lua config.
-fn dispatch_hyprlang(cmd: CompositorCommand) -> Result<()> {
+async fn dispatch(cmd: CompositorCommand, strategy: DispatchStrategy) -> Result<()> {
+    fn socket_call(dt: DispatchType<'_>) -> Result<()> {
+        Ok(Dispatch::call(dt)?)
+    }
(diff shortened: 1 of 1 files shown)


src/services/compositor/hyprland.rs
```diff
@@ -11,106 +11,110 @@ use hyprland::{
...
-        .args(["eval", &lua])
-        .output()
-        .await?;
-    Ok(())
 }
 
 pub async fn execute_command(cmd: CompositorCommand) -> Result<()> {
-    if is_lua_config().await {
-        dispatch_lua(cmd).await
+    let strategy = if is_lua_config() {
```
MalpenZibo, line 127: Here, we can add a way to fetch the `is_lua_config` only when it's really required.

We could use something like 
```
  static CONFIG_PROVIDER: AtomicU8 = AtomicU8::new(0);

  fn is_lua_config() -> bool {
      match CONFIG_PROVIDER.load(Ordering::Relaxed) {
          1 => false,
          2 => true,
          _ => {
              let lua = detect_lua_config(); // lo spawn di `hyprctl status` attuale
              CONFIG_PROVIDER.store(if lua { 2 } else { 1 }, Ordering::Relaxed);
              lua
          }
      }
  }
```

And update the value only when the hyprland config changes with
```
  listener.add_config_reloaded_handler(|| {
      Box::pin(async { CONFIG_PROVIDER.store(0, Ordering::Relaxed); })
  });
```

boomboompower:
A few questions/thoughts about this PR after going through it:
- Every arm now has to branch on strategy internally, so the Socket-vs-Lua decision ends up repeated across all six commands instead of being resolved once. With the previous implementation (two flat functions), each one was a single switch over commands. Was there a specific motivation for restructuring it this way?
- Would it be possible to add a bit more logging around the if let Ok(state_guard) = internal_state.read() and if let Ok(mut state_guard) = internal_state.write() spots? Since std::sync::RwLock can get poisoned by a panic (unlike the old tokio lock), some debug logging would make any issues a bit easier to trace.

---

In general, I think there might be a couple of subtle behavior changes. `is_lua_config()` and the eval calls switched to blocking `std::process::Command`, but they're still called from inside the async execute_command (which runs through Task::perform), so dispatches now block an executor thread waiting on hyprctl instead of yielding like before. Same goes for the RwLock poisoning from panic as above.

dacrab:
Addressed feedback:\n\n1. **Lua eval now async** — `dispatch` is `async` and uses `tokio::process::Command` for Lua eval calls, so they no longer block the executor thread. The Socket path remains sync (it always was via `Dispatch::call`).\n\n2. **RwLock poisoning logging** — added `log::warn!` at all three `std::sync::RwLock` access points so a poisoned lock is visible in logs instead of being silently ignored.\n\n3. **Strategy branching** — the per-arm strategy switch is unavoidable since each command has different Lua syntax, but the dispatch logic is now unified in a single function instead of duplicated across two separate functions.

dacrab:
Closing per maintainer feedback — will avoid large refactors without prior discussion. The changes here addressed boomboompower's review but I don't want to add review burden right now.

PR #827, opened 2026-06-16
feat(network): detect IWD connection state in known_connections

dacrab:
Replaces the TODO placeholders for `state` and `working` in IWD's `known_connections` by querying each station's `State` and `ConnectedNetwork` properties to determine the actual connection state.

### DIFF
--- a/src/services/network/iwd_dbus/mod.rs
+++ b/src/services/network/iwd_dbus/mod.rs
@@ -145,6 +145,12 @@ impl super::NetworkBackend for IwdDbus<'_> {
     /// List known (provisioned) SSIDs.
     async fn known_connections(&self) -> anyhow::Result<Vec<KnownConnection>> {
         let nets = self.reachable_networks().await?;
+        let stations: HashMap<_, _> = self
+            .stations()
+            .await?
+            .into_iter()
+            .map(|s| (s.inner().path().clone(), s))
+            .collect();
         let mut networks = HashMap::<String, AccessPointData>::new();
         for (n, signal_strength) in nets {
             if n.known_network().await.is_err() {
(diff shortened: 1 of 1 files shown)


src/services/network/iwd_dbus/mod.rs
```diff
@@ -153,16 +153,48 @@ impl super::NetworkBackend for IwdDbus<'_> {
             let ssid = n.name().await?;
             let path = n.inner().path().clone().into();
             let device_path = n.device().await?.clone();
+
+            // Determine connection state by checking if this network is connected
+            let station = self
+                .stations()
```
romanstingler: This is a performance concern `self.stations()` is called inside the loop.
`stations()`  is `list_proxies!(...)` over the zbus object manager. Every iteration re-enumerates station proxies. 

move the lookup out of the loop:
```
let stations: HashMap<_, _> = self
    .stations()
    .await?
    .into_iter()
    .map(|s| (s.inner().path().clone(), s))
    .collect();

for (n, signal_strength) in nets {
    ...
    let station = stations.get(&device_path);
    ...
}
```

romanstingler (changes requested):

src/services/network/iwd_dbus/mod.rs
dacrab: Good catch — moved the stations() call out of the loop and into a HashMap lookup before iterating. Pushed in 194ba7c3.

MalpenZibo (approved):
For me it's ok! Let's wait @romanstingler

dacrab:
Rebased on latest `main` and resolved merge conflict (upstream renamed `AccessPoint` → `AccessPointData`). @romanstingler your concern about `stations()` being called in the loop was already addressed in the previous commit — it now uses a single `HashMap` lookup.

romanstingler (approved):

PR #829, opened 2026-06-16
fix(logger): scope stdout log duplication to debug builds, guard OSD handler

dacrab:
The unconditional `duplicate_to_stdout(Duplicate::All)` on the primary logger made the conditional debug-only duplication on line 97 dead code. Removed the unconditional call so stdout duplication only happens in debug builds.

Also replaces `unreachable!()` for `IpcCommand::ToggleVisibility` in the `IpcOsdCommand` handler with a logged warning, since the invariant was externally enforced and fragile.


romanstingler:
I only renamed the title to cover both fixes, 
they are small and ok to be in one PR. 
Code makes sense now that we have OSD.

romanstingler (approved):

MalpenZibo (approved):
👌

PR #835, opened 2026-06-18
add support for bond network interfaces

### LINKED ISSUE #806: add bond interfaces to SystemInfo/network module
The filter in the modules/system_info on the network filter out bond(and tun) interfaces. Would be nice to maybe just this be a config? idk how often new interface types are added lol. for now it is an easy fix:
modules/system_info.rs:336
```diff
diff --git a/src/modules/system_info.rs b/src/modules/system_info.rs
index a5358664..317525fd 100644
--- a/src/modules/system_info.rs
+++ b/src/modules/system_info.rs
@@ -334,6 +334,7 @@ fn get_system_info(
                 || name.contains("wl")
                 || name.contains("wlan")
                 || name.contains("br")
+                || name.contains("bond")
         })
         .sorted_by_key(|(name, _)| {
             if name.contains("en") {
@@ -354,6 +355,9 @@ fn get_system_info(
             if name.contains("br") {
                 return 4;
             }
+            if name.contains("bond") {
+                return 5;
+            }
 
             99
         })
```

romanstingler:
Added because I haven't seen the reporter doing the PR
closes https://github.com/MalpenZibo/ashell/issues/806

Added Bond network type.
Refactored some code because it duplicated the interface names. 
Then there was `contains` and 'wl' and after that 'wlan' so basically 'wlan' was never reachable. 
Instead of the 'contains' I replaced that with 'starts_with' as all interface names have this notion.

### DIFF
--- a/src/modules/system_info.rs
+++ b/src/modules/system_info.rs
@@ -188,6 +188,13 @@ fn get_temperature_sensor_label(
     sensor_label
 }
 
+/// Returns a sort priority for Linux network interface names we want to surface
+/// in the system info panel, or `None` for interfaces we should ignore.
+fn interface_priority(name: &str) -> Option<usize> {
+    const PREFIXES: &[&str] = &["en", "eth", "wl", "br", "bond"];
+    PREFIXES.iter().position(|p| name.starts_with(p))
+}
+
 #[allow(clippy::too_many_arguments)]
 fn get_system_info(
     system: &mut System,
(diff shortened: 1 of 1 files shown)


InnocentZero:
Hey, I'm not the reporter, but is it possible to add wireguard based tunnel VPNs? Idk how it'd be checked though, since the interface names are varied quite a bit. Netbird creates wt0 (this is fixed), tailscale creates a default of tailscale0, and stock wireguard creates a default of wg0 (I think).

romanstingler:
@InnocentZero  I could add 

```
const PREFIXES: &[&str] = &[
    "en", "eth", "wl", "br", "bond",  // physical / local virtual
    "wg", "wt", "tailscale",          // WireGuard-based tunnels
    "tun", "tap",                     // generic tunnels (OpenVPN, etc.)
];
```

but the network priority code would find your eth or wlan connection and would prioritize this one.
I would suggest creating a new issue and referencing  
"https://github.com/MalpenZibo/ashell/pull/315

I have no idea how 
@MalpenZibo 
Wants to handle that in the project

MalpenZibo:
Let's create a new issue for that. I need to think about it

MalpenZibo (approved):

PR #843, opened 2026-06-27
Add NewIcon watcher

SimoneFelici:
Hi,

This patch adds support for the `NewIcon` signal.

For example, ProtonVPN updates its tray icon by changing `IconName` and emitting `NewIcon` when connecting to a server. Previously, AShell would ignore this signal.

I also added `current_icon_from_proxy()` to reuse the existing icon lookup logic when handling icon update signals.

I also had to disable caching, since zbus only invalidates the cache when `org.freedesktop.DBus.Properties.PropertiesChanged` is called, but ProtonVPN does not do that when sending `NewIcon`, as expected.
With zbus property caching enabled, reading `IconName` after `NewIcon` returns the cached value, so the icon does not update correctly.

If there is a better way to invalidate or refresh the cached properties only when `NewIcon` is received, I’m happy to adjust the implementation.

# Before

https://github.com/user-attachments/assets/e16954ef-dea0-47a8-94de-5fb6d79206ed

Note:
The icon is grayed out because the cache is enabled. Disabling the cache makes it show the normal icon, as in the “after” video, but it doesn't update since there is no NewIcon watcher.

# After

https://github.com/user-attachments/assets/d32f638b-83ae-4bc8-b8e3-19edc3fe70b5

Thank you!


src/services/tray/mod.rs
```diff
@@ -345,20 +357,13 @@ impl StatusNotifierItem {
         let item_proxy = StatusNotifierItemProxy::builder(conn)
             .destination(dest.to_owned())?
             .path(path.to_owned())?
+            .cache_properties(zbus::proxy::CacheProperties::No)
```
MalpenZibo: We should avoid this. Introduce a performance regression.

My idea is to create a dedicated proxy with the cache disabled only for this.

``` rust
fn split_service_name(name: &str) -> (&str, &str) {
    match name.find('/') {
        Some(idx) => (&name[..idx], &name[idx..]),
        None => (name, "/StatusNotifierItem"),
    }
}

...

pub async fn new(conn: &zbus::Connection, name: String) -> anyhow::Result<Self> {
        let (dest, path) = split_service_name(&name);

        let item_proxy = StatusNotifierItemProxy::builder(conn)
            .destination(dest.to_owned())?
            .path(path.to_owned())?
            .build()
            .await?;

...

 let new_icon = item.item_proxy.receive_new_icon().await;
 if let Ok(new_icon) = new_icon {
    // NewIcon has no matching PropertiesChanged, so a cached read would be stale;
    // use an uncached proxy here while the cached one keeps the *_changed streams alive.
    let (dest, path) = split_service_name(&name);
    let uncached_proxy = StatusNotifierItemProxy::builder(conn)
        .destination(dest.to_owned())?
        .path(path.to_owned())?
        .cache_properties(zbus::proxy::CacheProperties::No)
        .build()
        .await?;

    new_icon_change.push(
        new_icon
        .cache_properties(zbus::proxy::CacheProperties::No)
        .build()
        .await?;

    new_icon_change.push(
        new_icon
        .filter_map({
            let name = name.clone();
            let uncached_proxy = uncached_proxy;

            move |_| {
                let name = name.clone();
                let uncached_proxy = uncached_proxy.clone();

                async move {
                    current_icon_from_proxy(&uncached_proxy)
                        .await
                        .map(|icon| TrayEvent::IconChanged(name.to_owned(), icon))
                    }
                }
            })
        .boxed(),
    );
  }

```

Can you check if this could work?

MalpenZibo (changes requested):

src/services/tray/mod.rs
SimoneFelici: Yes, this works, thank you!

MalpenZibo (approved):

PR #845, opened 2026-06-29
feat/generic compositor backend

MalpenZibo:


src/services/compositor/types.rs
```diff
@@ -34,38 +34,49 @@ pub struct ActiveWindowNiri {
...
 #[derive(Debug, Clone, PartialEq)]
 pub enum ActiveWindow {
     Hyprland(ActiveWindowHyprland),
     Niri(ActiveWindowNiri),
+    Generic(ActiveWindowGeneric),
 }
 
 impl ActiveWindow {
     pub fn title(&self) -> &str {
         match self {
```
clotodex, line 52: I assume there are a lot of cases where you would just want to access self.title
Maybe worth implementing a deref trait? Wdyt?

src/services/compositor/hyprland.rs
clotodex, line 1: We could think about moving the function implementation directly into the impl block
But no strong opinion on it.
This way the struct is a wrapper around what was there rather than the rewrite.

Up to you

clotodex (changes requested):
Lgtm if you want to merge
I did not test it yet
Otherwise the patch system is an interesting approach
I assume the goal is to minimize the specific surface area that is exposed per compositor? So whatever CAN be handled by the generic wayland interface should

I dont know if i would have built it the same way but it looks like it works :P and i dont have a better idea at the moment

The only thing i dont really like about the code is that it feels like a lot of types and a lot of wrapping if things. This might be a rust effect though to have the good typing experience.

Long story short, great job!! Will be amazing for people on other compositors 🎉

MalpenZibo:
I tried to have a generic Wayland implementation that could be used to compose a dedicated implementation. That said, I also have the doubt that this approach doesn't provide any real benefit but increases the complexity. I don't know how many hybrid approaches we will ever have. (With a hybrid approach, I mean a specialized implementation that uses some feature of the generic one)

MalpenZibo:
I also had this version https://github.com/MalpenZibo/ashell/pull/847, much simpler.

Maybe it's the right direction because we will probably never have to create a hybrid backed by extending the generic one

MalpenZibo:
Superseded by #847

PR #846, opened 2026-06-29
feat(compositor): add MangoWC backend

MalpenZibo:


MalpenZibo:
@Sachin-Bhat, if you're still interested, could you test it?

PR #847, opened 2026-06-29
Feat/generic backend simple

MalpenZibo:
A simple version of #845 without the possibility to compose a compositor implementation using generic compositor function + specific one


website/docs/compositor-support.md
```diff
@@ -0,0 +1,43 @@
...
+
+| Protocol | Provides |
+| --- | --- |
+| `wl_output` | Monitors |
+| `ext-workspace-v1` | Workspaces (listing and switching) |
+| `wlr-foreign-toplevel-management` | Active window (title and class) |
+
+## Feature matrix
+
+| Feature | Hyprland | Niri | Generic Wayland |
```
clotodex, line 35: Not sure if necessary but initial title is only supported by hyprland
It is a bit more of a behind the scenes thing

clotodex (approved):
Nice, quite clean and i agree more with this approach
If lateron we realize a rework would be worth it due to code duplication etc we can still rewrite it
For now this is much vleaner to maintain as well
Thank you!!

PR #848, opened 2026-06-30
fix(iwd): derive AccessPoint state and working from connection status

dacrab:
fix(iwd): derive AccessPoint state and working from connection status

Instead of hardcoding `DeviceState::Unknown` and `working: false` for every access point, query `net.connected()` via D-Bus to determine the actual state. Connected networks get `DeviceState::Activated` + `working: true`; others get `Disconnected` + `working: false`. Also extracts a small `ap_state_from_connected` helper to remove the duplicated mapping.

### Note on impact (data hygiene, not a behavioral fix)

Per review: `AccessPoint.working` is currently write-only (never read), and `AccessPoint.state` is never read for access points — the UI decides the connected network via `active_connections` matched by SSID in `src/modules/settings/network.rs`. So this PR does **not** change any user-visible behavior today; its value is correctness of the data model and removing the `// TODO:`s, so anything that reads these fields in the future is correct. The unrelated signal-agent threshold comment hunk was dropped (reverted to the original NM-aligned wording).

### DIFF
--- a/src/services/network/iwd_dbus/mod.rs
+++ b/src/services/network/iwd_dbus/mod.rs
@@ -84,6 +84,17 @@ fn map_iwd_rssi_to_percent(rssi_hundredths: i16) -> u8 {
     percent.round() as u8
 }
 
+/// Maps an IWD network's connection status to the `DeviceState` reported for
+/// its `AccessPoint`. A connected network is `Activated`; everything else is
+/// `Disconnected`.
+fn ap_state_from_connected(connected: bool) -> DeviceState {
+    if connected {
+        DeviceState::Activated
+    } else {
+        DeviceState::Disconnected
+    }
+}
+
 impl<'a> Deref for IwdDbus<'a> {
     type Target = ObjectManagerProxy<'a>;
     fn deref(&self) -> &Self::Target {
@@ -161,11 +172,7 @@ impl super::NetworkBackend for IwdDbus<'_> {
                 strength: map_iwd_rssi_to_percent(signal_strength),
                 max_bitrate: 0,
                 frequency: 0,
-                state: if connected {
-                    DeviceState::Activated
-                } else {
-                    DeviceState::Disconnected
-                },
+                state: ap_state_from_connected(connected),
                 public: n.type_().await? == "open",
                 working: connected,
             };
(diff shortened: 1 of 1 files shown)


MalpenZibo:
## Review

Thanks for this! The mapping itself is correct: `net.connman.iwd.Network.Connected` is the canonical "this network is the active one" property, and `connected()` is a cached `#[zbus(property)]`, so the extra read is essentially free after the first fetch.

One important thing to flag about **actual impact**, and one unrelated hunk I'd drop.

### This is correct, but it has no observable effect today

I traced the consumers of these two fields:

- **`AccessPoint.working` is never read** anywhere. It only appears in the struct definition and in three assignments (the two in this PR plus the hardcoded `false` in the NM backend at `src/services/network/dbus.rs:720`). It's a write-only field.
- **`AccessPoint.state` is never read for access points.** The only `.state` in the network module is `station.state()` (a different thing), and the `DeviceState::` comparisons in `dbus.rs` operate on the NM *device* state, not on `AccessPoint`.
- The UI decides which network is connected in `src/modules/settings/network.rs` via `service.active_connections`, matched by SSID (`.partition(|ac| active_connection.is_some_and(|(ssid, _)| ssid == ac.ssid))`), **not** by looking at `AccessPoint.state`/`working`.
- It also doesn't introduce update churn: events replace the list wholesale, and lookups match by `path`/`ssid` rather than full-struct equality (despite the `Eq` derive).

So the real value here is **data hygiene**: the fields now reflect reality instead of misleading constants, the `// TODO:`s are resolved, and anything that reads them in the future will be correct. That's worth doing. But the description ("query `net.connected()` to determine the actual state") and the `bug` label suggest a user-visible fix, and there isn't one right now. If there's an actual symptom (e.g. a connected network not being highlighted), the root cause is elsewhere, in `active_connections_info()`, and this PR doesn't touch it. Worth clarifying the description so it doesn't imply a behavioral fix.

### Please drop the unrelated comment change

The hunk that rewrites the signal-agent threshold comment is scope creep, and it removes accurate information:

```diff
- // Register signal level agent with thresholds aligned to NM's bar ranges
- // NM's bar ranges: >80% (5 bars) ...
- // Using NM linear mapping ...
+ // ... aligned to common bar ranges
+ // Bar ranges: >80% ...
+ // Using linear mapping ...
```

Those thresholds were deliberately aligned to NetworkManager's bar mapping so the IWD and NM backends render signal bars consistently. "common bar ranges" / "linear mapping" is vaguer and loses the rationale for those specific numbers. I'd revert this hunk since it's unrelated to the state fix.

### Minor

- The `if connected { Activated } else { Disconnected }` block is duplicated across the two sites. With only two occurrences it's fine, but a small helper (`fn ap_state(connected: bool) -> DeviceState`) would remove the duplication. Optional.

### Verdict

The state/working derivation is correct and harmless, happy to see it land. Two asks before merge:

1. **Drop the signal-agent comment hunk** (unrelated and regresses the info).
2. **Adjust the PR description** so it doesn't read as a visible-bug fix, since those fields are currently inert.

Optional: the small helper to dedupe the `connected -> DeviceState` mapping.

PR #849, opened 2026-06-30
refactor(widget): replace bare unwrap with expect in layout children access

dacrab:
Replaces 13 occurrences of `layout.children().next().unwrap()` across 4 custom iced widgets with `.expect()` calls containing descriptive messages about which widget failed. This turns potentially silent panics into actionable error messages if the layout invariant (exactly one child) is ever violated.


dacrab:
Closing — this is a minor style change that doesn't add meaningful value. Will be more selective going forward.

PR #851, opened 2026-06-30
fix(workspaces): use equality instead of substring for monitor visibility

### LINKED ISSUE #804: MonitorSpecific visibility filter breaks when output names overlap as substrings (e.g. eDP-1 / DP-1)

dacrab:
Fixes #804

## Problem

`MonitorSpecific` / `MonitorSpecificExclusive` visibility mode used `str::contains()` to decide whether a workspace belongs to the current monitor. When one output name is a substring of another (e.g. `eDP-1` contains `DP-1`), the laptop bar accepts every workspace owned by the dock's DisplayPort as if it were its own.

## Fix

Replace `.unwrap_or_else(..).contains(..)` with `.is_none_or(|name| name == w.monitor)`. Both `monitor_name` and `w.monitor` are canonical output names at this point — exact equality is the correct relation. The `is_none_or` fallback preserves the original behavior of showing everything when the monitor name is unknown.

## Testing

`make check` passes (fmt + cargo check + clippy -D warnings).

### DIFF
--- a/src/modules/workspaces.rs
+++ b/src/modules/workspaces.rs
@@ -443,14 +443,12 @@ impl Workspaces {
                         let show = match self.config.visibility_mode {
                             WorkspaceVisibilityMode::All => true,
                             WorkspaceVisibilityMode::MonitorSpecific => {
-                                monitor_name
-                                    .unwrap_or_else(|| &w.monitor)
-                                    .contains(&w.monitor)
+                                monitor_name.is_none_or(|name| name == w.monitor)
                                     || !outputs.has_name(&w.monitor)
                             }
-                            WorkspaceVisibilityMode::MonitorSpecificExclusive => monitor_name
-                                .unwrap_or_else(|| &w.monitor)
-                                .contains(&w.monitor),
+                            WorkspaceVisibilityMode::MonitorSpecificExclusive => {
+                                monitor_name.is_none_or(|name| name == w.monitor)
+                            }
                         };
 
                         if show {


romanstingler:
you are frustrating me 
duplicate of 
https://github.com/MalpenZibo/ashell/pull/805/changes

@MalpenZibo

MalpenZibo:
This start to become quite annoying... @dacrab please avoid doing random works. 
I know that write code nowadays it's quite cheap but review everything it's still a pain in the ass

PR #852, opened 2026-06-30
fix(module_item): remove clip to prevent icon cropping

### LINKED ISSUE #571: [Bug] Update module icon is cropped

dacrab:
Fixes #571

## Problem

`.clip(true)` on the module item container clips child content to the container's bounding box. Nerd Font icon glyphs can slightly overflow their text bounds (especially at certain scale factors or with tinyskia software renderer), causing the icon to appear cropped.

## Fix

Remove `.clip(true)` from both the button path (line 69) and the plain container path (line 97) in `module_item.rs`. This matches the fix on @romanstingler's branch `fix/cropped-module-icon`.

## Testing

`make check` passes (fmt + cargo check + clippy -D warnings).


romanstingler:
your bot is out of control 
@MalpenZibo fyi

PR #856, opened 2026-07-01
Auto highlight active workspace using color palette

1randomguy:
Automatically highlight the active workspace with a different color from the specified colorpalette. (Alternative to #814 )

e.g.: 
image

My biggest gripe with it is that the contrast is not super strong, but when taken together with the size difference its manageble for me, and at least a lot more legible than without it.

### DIFF
--- a/src/theme.rs
+++ b/src/theme.rs
@@ -612,11 +612,7 @@ impl AshellTheme {
                 })),
                 border: Border {
                     width: if is_urgent || is_empty { 1.0 } else { 0.0 },
-                    color: if is_urgent {
-                        urgent_color
-                    } else {
-                        bg_color
-                    },
+                    color: if is_urgent { urgent_color } else { bg_color },
                     radius: radius_lg.into(),
                 },
                 text_color: if is_urgent && is_empty {
--- a/src/theme.rs
+++ b/src/theme.rs
@@ -519,84 +519,34 @@ impl AshellTheme {
     ) -> impl Fn(&Theme, Status) -> button::Style + use<> {
         let radius_lg = self.radius.lg;
         move |theme: &Theme, status: Status| {
-            let (bg_color, fg_color) = colors.map_or_else(
-                || {
-                    (
-                        theme.extended_palette().background.weak.color,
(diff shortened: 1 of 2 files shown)


src/theme.rs
```diff
@@ -544,12 +545,68 @@ impl AshellTheme {
                     )
                 },
             );
+            let (bg_strong, fg_strong) = colors.map_or_else(
```
MalpenZibo: Here we can simplify a little bit

```rust
let primary = colors.map(|c| {
    c.map_or_else(
        || theme.extended_palette().primary,
        |c| palette::Primary::generate(
            c.get_base(),
            theme.palette().background,
            c.get_text().unwrap_or_else(|| theme.palette().text),
        ),
    )
});
let resolve = |bg: fn(&palette::Background) -> palette::Pair,
               pr: fn(&palette::Primary) -> palette::Pair| {
    match primary {
        Some(p) => { let pair = pr(&p); (pair.color, pair.text) }
        None => { let pair = bg(&theme.extended_palette().background); (pair.color, theme.palette().text) }
    }
};
let (bg_color, fg_color)   = resolve(|b| b.weak, |p| p.base);
let (bg_strong, fg_strong) = resolve(|b| b.strong, |p| p.strong);
let (bg_weak, fg_weak)     = resolve(|b| b.weak, |p| p.weak);▎
```

MalpenZibo (commented):
Let's use this convention:
- workspace NOT active -> weak
- workspace active -> base
- workspace hover -> strong

For the empty workspace (the one with only the border)
- workspace NOT active -> fill: background.weak, border: color.weak
- workspace active -> fill: background.strong, border: color.base
- workspace hover -> fill: background.strong, border: color.strong

src/theme.rs
1randomguy: oh yeah, that is much nicer 😄

1randomguy:
Alright, I'm fine with that, here is what that looks like with my theme for reference:
image

Let me know what you think :)

MalpenZibo (approved):

PR #866, opened 2026-07-02
feat(blur): opt-in compositor blur behind translucent surfaces

MalpenZibo:


website/docs/configuration/appearance/general.md
romanstingler: maybe now here the blur description is confusing with the blur section below. 
maybe reword that to "darkening" or simmilar

src/config.rs
```diff
@@ -1049,6 +1049,32 @@ pub struct Appearance {
...
+    /// Blur the wallpaper behind ashell's translucent surfaces via
+    /// `ext-background-effect-v1`. No-op where the protocol is unsupported.
+    pub blur: BlurMode,
+}
+
+/// When to ask the compositor for background blur.
+#[derive(Deserialize, Default, Clone, Copy, Debug, PartialEq, Eq)]
+#[serde(rename_all = "lowercase")]
+pub enum BlurMode {
+    /// Blur when `opacity < 1`; blurring an opaque surface would not show.
```
romanstingler, line 1061: here the wording "resolved against the most translucent surface" would be more honest

currently we have OSD and toast cards with 1.0 opacity
if we had a scenario: bar = 1.0, menu = 0.7, auto → bar/pills enabled(1.0) = off, menu enabled(0.7) = on, OSD/toast enabled(1.0) = off 
we would have no need to draw most of them. 

what if we do something like
```
// theme.rs
pub blur: BlurMode,          // was: bool
...
blur: appearance.blur,       // no min(), no "collapse later" comment

// app.rs (Solid) and module_group.rs:
if blur.enabled(opacity) && ...          // bar opacity, already in scope

// menu.rs:
if blur.enabled(menu_opacity)            // already in scope

// osd.rs and notifications.rs:
if blur.enabled(1.0)                     // "opaque today" — literal and self-documenting
```

website/docs/configuration/appearance/general.md
```diff
@@ -138,3 +138,49 @@ between `0.0` (disabled) and `1.0` (maximum blur).
...
+blur = "auto"
+
+[appearance.bar]
+opacity = 0.8
+```
+
+### Compositor setup
+
+Some compositors need blur enabled per surface before they will apply it. On
+niri, add a layer rule matching ashell's namespaces:
```
romanstingler: Hyprland's `shouldBlur` checks `decoration:blur:enabled`

example 

```
blur = {
    enabled   = true, 
    size      = 3,
    passes    = 1,
    vibrancy  = 0.1696,
},
```

src/config.rs
MalpenZibo: In reality, I have a follow-up PR about this. Originally, I would like to address here some inconsistency in how `opacity` works, but it ends up being quite a heavy refactor, so I moved the changes to a dedicated PR.

One of the changes is to move to a single opacity level for the entire app, with the possibility to re-add dedicated values in the future. For this reason, this stuff is temporary.

romanstingler:
lgtm

PR #867, opened 2026-07-02
feat(appearance): replace bar AppearanceStyle with dedicated settings

MalpenZibo:
## Breaking
- Drop style config
- drop gradient style
- move `opacity` config under `[appearance.bar]` section
- Add the following settings for styling the app bar:
  - surface - transparent or solid
  - radius 
  - margin


clotodex (approved):
Lgtm (untested)
Do you want to do a hard switchover or support the config file options as deprecated. E.g. gradient collapses to solid, but the config does not error?

MalpenZibo:
The gradient could be hard-dropped.
The style config I can create a mapper that maps that setting to the new one. I still need to understand if I prefer an hard drop or not

PR #877, opened 2026-07-09
fix(hyprland) Switch main keyboard only

impervius:
When querying the keyboard layout, we only consider the `main` keyboard:
https://github.com/MalpenZibo/ashell/blob/0eb6587827f36a34933889cf3375eaad31759295/src/services/compositor/hyprland.rs#L271-L279
Similarly, when changing the keyboard layout, we should apply the change only to the `main` keyboard.


MalpenZibo (approved):
👍

PR #880, opened 2026-07-17
fix(osd): align the audio OSD elements correctly

s-stepien:
The current OSD implementation displays the audio (mic, volume) elements a bit askew. The icon isn't centered, the progress bar shrinks/expands based on text width.

Fix this by:
(1) asking the progress bar to fill the available space - will work nice
    when show_volume_percentage is set to false
(2) setting the width and alignment of the text elements (note: icon
    element now drops the container)
(3) removing the spacing between elements that isn't really needed

With this, the display should look nice for all values of scale_factor.

Tested only with audio stuff, but should also look good with brightness OSD.

### DIFF
--- a/src/osd.rs
+++ b/src/osd.rs
@@ -1,7 +1,7 @@
 use std::time::Duration;
 
 use iced::{
-    Alignment, Border, Element, Length, Task, Theme,
+    Alignment, Border, Element, Length, Task, Theme, alignment,
     widget::{container, progress_bar, row, text},
 };
 use tokio::time::sleep;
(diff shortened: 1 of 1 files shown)


romanstingler:
@MalpenZibo 
**ORIG** 
image
image

**NEW**
image
image

`.width(font_size.sm * 4.0)` 
image

Visually, I don't care both have no equidistant spacing between icon and slider for different `show_volume_percentage = true|false` 
also I have no idea if `.width(font_size.sm * 3.0)` works for all fonts

s-stepien:
For me the original bubble did not look like that. The alignment was off by a lot. I'm using hyprland with scale set to "auto". Anyway...I've moved to noctalia in the meantime, so I do not care anymore.

PR #883, opened 2026-07-21
feat: Add keyboard back-light quick setting with slider

3bnz:
https://github.com/user-attachments/assets/d5eb5652-638c-4edb-9434-f22280cadfd7

- Adds a quick setting toggle with a slider in the menu for controlling built-in keyboard backlight.
- Supports a single keyboard - which should be fine since this does not concern bluetooth and usb peripherals.
- When toggling back-light off and on again the previous brightness value is retained
- [Backlight symbol ](https://fonts.google.com/icons?selected=Material+Symbols+Outlined:backlight_high) is not included i
n nerd-fonts so I drew the missing icons and added them to `AshellCustomIcon-Regular.otf` 

*I used [fontforge](https://fontforge.org/) to edit the font. You can see that after the edit - the size has decreased. Not sure why, but I did test that mouse and keyboard status bar icons still look as expected.


romanstingler:
I like the feature, even if for some reason my HP is still not supported (not the issue of this here).
I will be honest, I don't like the UI. 

In my opinion, we can use the same sliders as for the other components. 
Or we can discuss wide sliders (like i.e. Android)

image


https://github.com/user-attachments/assets/abbe39c5-1fae-4333-aa6b-6810743300c5
demo at https://github.com/romanstingler/ashell/tree/feature/kbd ()no real kbd brightness, due to the lack of drivers supporting it for my system)

But a button that opens the slider is nothing I would like to have.

MalpenZibo:
Hi, thanks for the support!

From my point of view, we should place a config setting to enable this option (by default, I prefer to hide this functionality. Then we should probably change the UX. A quick settings with a simple slider seems too much. Maybe we should simply place the slider under the screen brightness.

romanstingler:
yes I agree, just add it under the brightness slider.

3bnz:
@MalpenZibo thanks for the feedback!

### Quick Setting

The main reason for placing the slider inside a quick-setting was to copy gnome and macOS quick settings.
Personally I don't have a strong opinion here - so the PR has been simplified to a simple inline slider.
The toggle action is still kept in the leading button:



### Config Flag

As for the configuration flag - the slider would be already hidden for any system without built-in keyboard backlight - this _includes_ USB keboards since they normally handle backlight using keyboard shortcuts.
I added the flag if it's prefferred but feel free to revert the last commit if you change your mind ;)

3bnz:
@romanstingler
Cool demo! While out of scope for this feature - I also preffer **thick** sliders.
We could explore writing a custom widget for the whole shell - the default iced sliders are a bit crude/simple indeed.
This could allow us to handle scrollwheel as well as animate "jumps" in a single place.

romanstingler (changes requested):
I had a look, 
first I added the icons to the font just replace these files with yours in `assets/`

[ashell_custom_icon_project.zip](https://github.com/user-attachments/files/31320231/ashell_custom_icon_project.zip)

* `keyboard_backlight_slider` is undocumented in the docs

* ~~scroll step logic is duplicated verbatim from `brightness.rs`, same 5% calculation and clamp~~ (NO NEED TO de-duplicate now, there will be a cleanup session later over more code) 


The biggest thing is that each slider tick spawns a fresh tokio task that builds a new proxy and calls `set_brightness`. Those tasks run concurrently and their completion order depends on the scheduler. If task N-1 lands after task N, the backlight settles one step off from where you dropped the slider, and UPower reports that value back so the UI shows the wrong-but-real state with no correction. The brightness service solves this with a commander channel and a throttle that coalesces writes and keeps them ordered.

src/services/upower/mod.rs
```diff
@@ -820,86 +934,92 @@ fn battery_status_from_timed_system_state(state: dbus::DeviceState, time: i64) -
...
     type Command = UPowerCommand;
 
     fn command(&mut self, command: Self::Command) -> iced::Task<ServiceEvent<Self>> {
-        iced::Task::perform(
-            {
+        match command {
+            UPowerCommand::SetKbdBacklight(brightness) => {
+                let _ = self.commander.send(brightness);
+                iced::Task::none()
+            }
```
3bnz, line 953: This is the only new code in the function.
rest is just indentation change and splitting into separate match arms due to fact that `Task::none()` can now be returned.

3bnz:
### Icons

Thanks for the icons @romanstingler! My svgs where way too small though - not even visible.
I scaled them up and them again - now in https://www.glyphrstudio.com/ project.



I hope this is the propper way.

### Commmander

> so the UI shows the wrong-but-real state with no correction

I think this is only partially true - we have the `Remote<Value>` wrapper, which will show the source of truth after a timeout.

That being said - while it _"works on my machine"_ due to backlight changes being near instantaneous - it would indeed be more robust to have a proper throttled rx/tx queue.

The commander has been added to the upower service (slider only). :)

src/services/upower/mod.rs
```diff
@@ -722,32 +796,72 @@ impl UPowerService {
...
+                        power_profile,
+                        kbd_backlight,
+                    )) => {
                         let peripheral_paths = peripherals
                             .iter()
                             .map(|p| p.device.inner().path().clone())
                             .collect();
 
+                        let (tx, rx) = tokio::sync::mpsc::unbounded_channel();
+                        Self::start_commander(conn.clone(), rx);
```
MalpenZibo: Maybe we should start the commander only if we can actually send a command to the device and the feature is enabled in the config

src/services/upower/mod.rs
```diff
@@ -282,13 +287,22 @@ impl From<PowerProfile> for StaticIcon {
...
+
 #[derive(Debug, Clone)]
 pub struct UPowerService {
     pub system_battery: Option<BatteryData>,
     pub charge_limit: Option<ChargeLimit>,
     pub peripherals: Vec<Peripheral>,
     pub power_profile: PowerProfile,
+    pub kbd_backlight: Option<KbdBacklight>,
     conn: zbus::Connection,
+    commander: UnboundedSender<u32>,
```
MalpenZibo: It's strange to see an Option<KbdBacklight> that is only present if the device supports the feature and a commander that is always present to send a command to change the KdbBacklight. Probably we should place both stuff in the same Option

src/services/upower/mod.rs
3bnz: Patched.

src/services/upower/mod.rs
3bnz: My bad - also patched.

By the way I have gone on a small side-quest - looking at implementing a custom widget,
which would also remove the need for commanders.
More info: https://github.com/MalpenZibo/ashell/issues/931#issuecomment-5532990977

MalpenZibo (approved):
I made a small adjustment to the font, we're ready to merge it

3bnz:
> I made a small adjustment to the font, we're ready to merge it

Checked the icon chage - visually looks good.

PR #894, opened 2026-08-02
Fix empty tray shown when all icons blocked

r3dArch:
The tray should be hidden when there are no icons to show, but the filter only checked if the `TrayData` was empty rather than if there was at least one not blocked tray item.

### Before
image

### After
image


MalpenZibo (approved):
👍

Issue #895, opened 2026-08-03
Sluggish / Dropped Input Events


logannc:
I had trouble writing this issue because I'm both unsure how to accurately describe the symptoms and whether the warnings in the logs are actually related. If there is something I should do for more diagnostic information, that would be helpful.

It seems like interacting with the bar has started being sporadically extremely laggy - I'm guessing input events are being dropped because it seems to not react to my mouse being over it or clicks occurring (combined with the following logs). Sometimes they'll simply be delayed, sometimes it seems like they're never received at all. I have not noticed any correlation for it.

Checking `journalctl --user -fu ashell.service`, I see

```
Aug 02 21:05:45 logan-desktop-prime ashell[923640]: WARN [iced_futures::subscription::tracker] Error sending event to subscription: TrySendError { kind: Full }
Aug 02 21:05:45 logan-desktop-prime ashell[923640]: WARN [iced_futures::subscription::tracker] Error sending event to subscription: TrySendError { kind: Full }
Aug 02 21:05:45 logan-desktop-prime ashell[923640]: WARN [iced_futures::subscription::tracker] Error sending event to subscription: TrySendError { kind: Full }
Aug 02 21:05:45 logan-desktop-prime ashell[923640]: WARN [iced_futures::subscription::tracker] Error sending event to subscription: TrySendError { kind: Full }
Aug 02 21:05:55 logan-desktop-prime ashell[923640]: WARN [iced_futures::subscription::tracker] Error sending event to subscription: TrySendError { kind: Full }
Aug 02 21:05:55 logan-desktop-prime ashell[923640]: WARN [iced_futures::subscription::tracker] Error sending event to subscription: TrySendError { kind: Full }
Aug 02 21:05:55 logan-desktop-prime ashell[923640]: WARN [iced_futures::subscription::tracker] Error sending event to subscription: TrySendError { kind: Full }
Aug 02 21:08:15 logan-desktop-prime ashell[923640]: WARN [zbus::proxy] Failed to populate properties cache via GetAll: org.freedesktop.DBus.Error.ServiceUnknown: The name is not activatable. Property change streams will not produce values.
Aug 02 21:33:08 logan-desktop-prime ashell[923640]: WARN [zbus::proxy] Failed to populate properties cache via GetAll: org.freedesktop.DBus.Error.ServiceUnknown: The name is not activatable. Property change streams will not produce values.
[...]

PR #896, opened 2026-08-03
only listen for Escape while it can close a menu

romanstingler:
fix 1/X
related to #895

The Escape-key subscription was always active, even though it can only do
something when `enable_esc_key` is set and a menu is open. An idle
`listen_with` subscription is not free: iced copies every interaction event
into a fixed-size channel per subscription and warns when it overflows.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -715,18 +715,25 @@ impl App {
                 _ => Message::None,
             }),
             iced::output_events().map(Message::OutputEvent),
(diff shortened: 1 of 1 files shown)


MalpenZibo:
Ok, for the fix.

We should probably deprecate this function and move this feature to the IPC. It requires more settings on the compositor side (a new keybinding) but it should be a cleaner solution.

MalpenZibo (approved):

PR #897, opened 2026-08-03
add configurable visualizer framerate and optimize cava lifecycle

romanstingler:
fix 2/X
related to https://github.com/MalpenZibo/ashell/issues/895

The media player visualizer now supports a `visualizer_framerate` config
option (clamped to 1-144, default 30) to control CAVA frame production.
Additionally, `menu_visualizer` now only runs cava while the media menu is
open, reducing CPU usage when the menu is closed. The cava subscription
includes the framerate in its identity to ensure proper restart on config
reload, and stale bar frames are cleared when the menu opens.


MalpenZibo (approved):
🚀

PR #898, opened 2026-08-03
add indicator controls for direct playback on mediaplayer

### LINKED ISSUE #884: Shorter media controls
Now in order to control the media with the  you need to click the media module, which opens the widget, and there click on your desired action.

It could be handy to have an option for quicker actions, for example left click the module on the bar for previous track, middle click for play/pause and right click for next track (and maybe wheel scroll for player specific volume). 

I had this config on other bars and it was very handy, it allowed me be more productive, easily switching tracks on the fly without having to open a widget.

The opened widget itself isn't much more informative / useful than the module itself (what it does that the module isn't is provides the artwork, which is not that interesting and most of the time also delayed, and also the album name, and also support for multiple players). Opening the widget for a single player that is open (which I guess is the common situation) is just a hassle for the use case of just skipping some track.

romanstingler:
Adds `indicator_controls` configuration option to enable direct media
playback controls on the bar indicator without opening the media menu.

When enabled, the indicator responds to left/middle/right clicks and
scroll wheel for previous, play/pause, next, and volume (5% steps).

Controls target the currently playing player, **falling back to the last
active player when nothing is playing**. The media menu cannot be opened
from the bar when this option is enabled.

closes https://github.com/MalpenZibo/ashell/issues/884

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -752,6 +752,7 @@ pub struct MediaPlayerModuleConfig {
     pub indicator_visualizer: Option<MediaPlayerVisualizer>,
     pub menu_visualizer: bool,
     pub visualizer_framerate: u32,
+    pub indicator_controls: bool,
 }
 
 impl Default for MediaPlayerModuleConfig {
@@ -763,6 +764,7 @@ impl Default for MediaPlayerModuleConfig {
             indicator_visualizer: None,
             menu_visualizer: false,
             visualizer_framerate: Self::DEFAULT_VISUALIZER_FRAMERATE,
+            indicator_controls: false,
         }
     }
 }
--- a/src/modules/media_player.rs
+++ b/src/modules/media_player.rs
@@ -149,6 +149,11 @@ pub enum Message {
     PlayPause(String),
     Next(String),
     SetVolume(String, f64),
+    ActivePrev,
+    ActivePlayPause,
+    ActiveNext,
+    ActiveVolumeUp,
+    ActiveVolumeDown,
     Event(ServiceEvent<MprisPlayerService>),
     ConfigReloaded(MediaPlayerModuleConfig),
     Bars(Vec<f32>),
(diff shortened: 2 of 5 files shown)


MalpenZibo (commented):
Not a big fan of the left/middle/right mouse click (mainly because with a laptop it's a pain in the ass). We could probably create, in a follow-up PR, some control buttons directly in the status bar.

About this PR:
- check how the `src/modules/settings/audio.rs` handles the scroll, maybe with this implementation we are losing some events because of the D-Bus round-trip
- maybe we could use this config
```
[media_player.indicator_controls]
left = "PlayPause"   # Menu | Prev | PlayPause | Next | None
middle = "Menu"
right = "Next"
scroll = "Volume"
```
We lost the prev button (which, in my experience, maybe isn't as used as the others), but we can still open the menu. Also, play/pause on the left button seems a better alternative.

impervius:
Hey, thanks for implementing this

Regarding @MalpenZibo's comment, this is his project and he surely will decide what gets merged, but my 2 cents is that making this (and everything else in this project) configurable is the best way to make this project fit the need for as much people as possible, and to their use cases. While having the control buttons in the status bar is a cool new feature and I might try it, I feel like my idea of making it with left/right/middle click is more natural (and requires less attention, as you can click anywhere on the pill). 

This PR doesn't change the default behavior and just adds an alternative way to control the media player, and I think that's OK. The config suggested by @MalpenZibo is cool too (and should allow also further customization, some people can choose to not use the "Menu" config at all, some can choose to use it exclusively, etc). But maybe that can be done in another PR and not be required for this one?

Also, I wonder for the general case if that should be added on another PR as a generic feature in ashell: allowing users to choose custom right/left/middle click for all the elements in the status bar (for example, middle clicking the time to run a custom timer script, right click on notifications to clear them, etc.).

MalpenZibo:
The point is that adding too many configuration possibilities increases the project complexity. When I started ashell the idea was to have something easy to configure and ready to be used.

For this reason, we should carefully decide what to add

ScottNotFound:
Would it be easier to allow widgets or modules to have arbitrary commands attached to click/scroll events? That would solve this current issue and pretty much any other issue requesting specific features for click/scroll interaction. 

I have zero rust experience so I have no idea how easy or hard it would be to add these structured keys to every module.

PR #901, opened 2026-08-04
reset network dialog state when menu opens

romanstingler:
closes the pw input when we click outside or on the bar, so that a subsequent click on the bar opens the settings menu and not the pw input field.

https://github.com/user-attachments/assets/6ed67882-b3fb-4506-b3d3-926ddad50425


MalpenZibo (approved):
🚀

PR #902, opened 2026-08-04
show WiFi frequency band in tooltip and quick settings

romanstingler:
Add `frequency_band()` function to map MHz frequencies to band labels
(2.4 GHz, 5 GHz, 6 GHz, 60 GHz), extend `ActiveConnectionInfo::WiFi`
with a `band` field, and display the band label in the WiFi tooltip and
quick settings button using the new `header_end` parameter on
`quick_setting_button`. 
For now only working on NetworkManager, iwd doesn't provide frequency or band information.

Image


src/services/network/mod.rs
```diff
@@ -135,6 +145,7 @@ pub enum ActiveConnectionInfo {
     WiFi {
         name: String,
         strength: u8,
+        band: Option<&'static str>,
```
MalpenZibo: Can't we simply save the frequency u32 and let the ui call the `frequency_band` function? The `AccessPointData` already does something similar

src/modules/settings/mod.rs
```diff
@@ -930,9 +932,19 @@ impl Settings {
             }
             MenuType::WifiTooltip => {
                 if let Some(label) = self.network.connected_wifi_label() {
-                    row![StaticIcon::Wifi4.to_text(), iced::widget::text(label)]
-                        .spacing(space.xs)
-                        .into()
+                    let band = self.network.connected_wifi_band();
+                    let font_size = use_theme(|t| t.font_size);
+                    let primary = use_theme(|t| t.iced_theme.palette().primary);
```
MalpenZibo: here we could call the use_theme just once and return both the values

src/modules/settings/network.rs
```diff
@@ -566,7 +572,9 @@ impl NetworkSettings {
 
                             styled_button(
                                 Element::from(
-                                    container(
+                                    container({
+                                        let primary = use_theme(|t| t.iced_theme.palette().primary);
+                                        let font_size = use_theme(|t| t.font_size);
```
MalpenZibo: Same here, also at the beginning of the `wifi_menu` fn we already access to the theme, we could centralize all the things there

src/modules/settings/network.rs
```diff
@@ -575,10 +583,15 @@ impl NetworkSettings {
                                             })
                                             .width(Length::Shrink),
                                             text(ac.ssid.as_str()).width(Length::Fill),
+                                            text(frequency_band(ac.frequency).unwrap_or(""))
+                                                .size(font_size.xs)
+                                                .style(move |_theme: &Theme| text::Style {
+                                                    color: Some(primary),
+                                                }),
```
MalpenZibo: here for the color you could use directly _theme.palette().primary. no need to access the `use_theme` for that information

romanstingler: changed, thx for the review

src/modules/settings/network.rs
```diff
@@ -575,10 +583,15 @@ impl NetworkSettings {
                                             })
                                             .width(Length::Shrink),
                                             text(ac.ssid.as_str()).width(Length::Fill),
+                                            text(frequency_band(ac.frequency).unwrap_or(""))
```
MalpenZibo: Here, it's better to avoid adding the text instead of a `unwrap_or("")`. Otherwise, we could add a `space.xs` for an empty string

MalpenZibo (approved):
🚀

PR #905, opened 2026-08-04
Webpage: automate changelog page generation and navigation integration

romanstingler:
Add a `sync-changelog.mjs` script that generates `website/src/pages/changelog.md` from the repository `CHANGELOG.md`, rendering each release as a collapsible details card. Wire the script into the `gh-pages-deploy` workflow so the changelog page stays in sync on every deploy. Update `docusaurus.config.ts` to index pages for search, and add Changelog links to the left navbar and the "More" dropdown menu. Include styling for the changelog release cards in `custom.css`.

Image

Image

@MalpenZibo I hope you don't mind me messing with the webpage. 😄

### DIFF
--- a/.github/workflows/gh-pages-deploy.yml
+++ b/.github/workflows/gh-pages-deploy.yml
@@ -41,6 +41,9 @@ jobs:
       - name: Install dependencies
         run: pnpm i
 
+      - name: Sync changelog page
+        run: pnpm run changelog
+
       - name: Build website
         run: pnpm build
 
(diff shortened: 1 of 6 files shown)


MalpenZibo:
Make sense but two things:
- with this we have a duplicate change log file (the one in the root and the MD generated). We could hit ignore the one generated but then a local pnpm start doesn't generate that file
- the MDX conversion could lead to other possible problems. That regex replace is enough for every possible cases?

MalpenZibo (approved):

PR #906, opened 2026-08-04
chore(docs): unfold configuration sidebar categories by default

romanstingler:
we don't have so many entries that we need to fold/collapse them 
image


MalpenZibo (approved):
👍

PR #915, opened 2026-08-10
[Feat] Custom margin settings

### LINKED ISSUE #351: [Feature] Appearance - Custom margin & padding
Hi! Awesome work, looks cool!

I would love a feature that allows users to choose the padding & the margin inside/outside de bar.
Could this be a new option?

ty

r3dArch:
Closes #351 

This allows for custom margin size in pixels without breaking the old margin system, although it removes the scaling that was applied to the old margins.

### Samples with config:

margin = [10, 20]

margin = "xs"

margin = 20

margin = [50, 20]

margin = [10, 20, 30, 40]

margin = [10, "xxl", 30, 40]

edit: wrote out the margins for easier parsing.

P.S. I've really enjoyed Ashell, thank you for making it.


MalpenZibo:
Check this https://github.com/MalpenZibo/ashell/pull/867

We already added a margin settings. For now it uses the spacing scale from the DS token but we don't have yet a way to change that scale.

r3dArch:
This PR builds on that one. It still allows specifying the margin as in #867, but also allows specifying it in pixels. To make the implementation simpler I removed margin scaling since I think most users would want to work with physical pixels, not scaled pixels; if you want, I can add scaling for just the space size margins, but I think it won't matter to most users.

r3dArch:
I would still need to update the docs as in #867, so I've made this PR a draft for now. Please let me know if this is something you would want to merge and I'll update the docs and make any requested changes.

r3dArch:
I updated the documentation. Please let me know if there is anything you'd like done differently.

src/outputs.rs
```diff
@@ -144,29 +144,21 @@ impl Outputs {
     pub fn get_height(surface: BarSurface, scale_factor: f64) -> f64 {
         (HEIGHT
             - match surface {
-                BarSurface::Solid => 8.,
-                BarSurface::Transparent => 0.,
+                BarSurface::Solid => 0.,
+                BarSurface::Transparent => 8.,
```
MalpenZibo: Why is it the opposite now?

src/outputs.rs
```diff
@@ -144,29 +144,21 @@ impl Outputs {
...
     pub fn exclusive_zone(layout: BarLayout, position: Position, scale_factor: f64) -> i32 {
         let height = Self::get_height(layout.surface, scale_factor);
-        let (top, _, bottom, _) = Self::margin(layout, scale_factor);
+        let (top, _, bottom, _) = layout.margin.into();
         height as i32
             + match position {
-                Position::Top => top,
-                Position::Bottom => bottom,
+                Position::Top => bottom,
+                Position::Bottom => top,
```
MalpenZibo, line 167: same here

src/outputs.rs
r3dArch: I had changed the transparent to be 8 because I removed the padding, but I think the solid should still be 8, so I could probably remove this match and make `HEIGHT = 26` instead. I'll push some changes soon.

src/outputs.rs
r3dArch: This is because the exclusive zone pushes other windows away from the bar, so when the bar is on the bottom it is the top margin that determines the exclusive zone and vice versa. I've retested this and it seems to work as expected.

MalpenZibo (approved):
🚀

Issue #627, opened 2026-04-14
Notification timeout progressbar


MalpenZibo:
Add an optional progress bar on the notification toast to show the timeout

0xch4z:
is this being worked on? if not I'd like to give it a go :)

Issue #626, opened 2026-04-14
Notification timeout stop


MalpenZibo:
Stop a notification toast timeout when the mouse is hovering

romanstingler:
@MalpenZibo  in the case multiple messages arrive at the same time 
do you want to keep all or just the one with the mouse on top? 

https://github.com/user-attachments/assets/580faed3-8132-42c4-bcf6-1a29939ddeab

PR #920, opened 2026-08-13
prevent duplicate notification toasts from being queued

romanstingler:
Ran into two related bugs while poking at the toast notification flow. Both stem from the
freedesktop `replaces_id` mechanism: a client can send a new `Notify` call with
`replaces_id` pointing at a notification that is still on screen, and the daemon just echoes that id back.

### Bug 1: duplicate entries on replace

Nothing checked whether an incoming notification's id was already visible. A replace for
an id that is still showing got pushed onto both `self.toasts` (the toast queue) and
`self.notifications` (the full list) a second time, instead of updating the existing
entry, so the same notification was drawn twice.

Fixed by skipping the eviction/push step in `toast_action_for_update_event` when the id is
already in `toasts`, and by `retain`ing away any existing entry for that id in
`apply_update_event` before pushing the new one.

### Bug 2: the expiry timer survives the notification it belongs to

Toast expiry was a fire and forget `Task::perform(sleep(timeout))` resolving to
`ExpireToast(id)`. Nothing cancelled it, and the handler only checked "is this id still
visible and not already dismissing", so it had no idea the notification instance had
changed underneath it. Two user visible symptoms:

* A notification that arrives as normal priority (say a 5s timeout) and is then replaced
  by a `Critical` version gets auto dismissed about 5s later by the first instance's
  timer. Critical notifications are meant to be persistent: the spec says they need
  explicit user acknowledgement and no auto timeout.
* No critical urgency needed either: a normal notification replaced with a *longer*
  timeout still dies at the old, shorter deadline. A 1.5s notification replaced by a 4s
  one disappears after about 1s.

### How expiry works now

Each toast's absolute deadline lives in `toast_timers: HashMap<u32, Instant>`, and expiry
is still a `tokio::time::sleep`, but the wakeup is now checked against the stored
deadline instead of being trusted:

* No deadline for that id, so the toast is persistent or already gone: ignore the wakeup.
* Deadline not reached yet, so the toast was replaced with a later one: reschedule for
  the remaining time.
* Otherwise: drop the deadline and dismiss.

A replace therefore just overwrites or removes an entry in the map. Replacing with a
critical notification does `toast_timers.remove(&id)` and there is nothing left that can
dismiss it. A sleep is spawned when the id had no deadline (a first notification, or a
critical one being replaced by a normal one, which needs a fresh timer since no wakeup is
pending) or when the new deadline lands before the pending wakeup. Later or equal
deadlines ride the in flight wakeup's reschedule, so a client that replaces the same id
in a loop (progress notifications, the canonical `replaces_id` user) does not pile up
timers.

A replace also cancels an in flight dismissal now, so a replacement arriving while the
previous instance is sliding out is shown instead of being swallowed by the animation.
`DismissAnimationComplete` bails out when its phase entry is already gone.

Behaviour was checked against `main` and against the first version of this PR on a live
Hyprland session, driving the D-Bus interface directly and measuring the rendered toast
stack. Details in the comment below.

### DIFF
--- a/src/modules/notifications.rs
+++ b/src/modules/notifications.rs
@@ -47,20 +47,20 @@ fn invoke_and_close_task(
     connection: Option<Connection>,
     id: u32,
     action_key: Option<String>,
+    revision: Option<u64>,
 ) -> Task<Message> {
     Task::perform(
         async move {
-            if let Some(connection) = connection {
-                if let Some(action_key) = action_key
-                    && let Err(e) =
-                        NotificationDaemon::invoke_action(&connection, id, action_key).await
-                {
-                    error!("Failed to invoke notification action for id {}: {}", id, e);
-                }
-                if let Err(e) = NotificationDaemon::close_notification_by_id(&connection, id).await
-                {
-                    error!("Failed to close notification id {}: {}", id, e);
-                }
+            if let Some(connection) = connection
+                && let Err(e) = NotificationDaemon::close_notification_by_id(
+                    &connection,
+                    id,
+                    action_key,
+                    revision,
+                )
+                .await
+            {
+                error!("Failed to close notification id {}: {}", id, e);
             }
         },
         |_| Message::NotificationClosed,
@@ -70,18 +70,26 @@ fn invoke_and_close_task(
 async fn close_notification_ids(connection: Option<Connection>, notification_ids: &[u32]) {
     if let Some(connection) = connection {
         for id in notification_ids {
-            if let Err(e) = NotificationDaemon::close_notification_by_id(&connection, *id).await {
+            if let Err(e) =
+                NotificationDaemon::close_notification_by_id(&connection, *id, None, None).await
+            {
                 error!("Failed to close notification id {}: {}", id, e);
             }
         }
     }
 }
 
-fn close_notification_by_id_task(connection: Option<Connection>, id: u32) -> Task<Message> {
+fn close_notification_by_id_task(
+    connection: Option<Connection>,
+    id: u32,
+    revision: Option<u64>,
+) -> Task<Message> {
     Task::perform(
         async move {
             if let Some(connection) = connection
-                && let Err(e) = NotificationDaemon::close_notification_by_id(&connection, id).await
+                && let Err(e) =
+                    NotificationDaemon::close_notification_by_id(&connection, id, None, revision)
+                        .await
             {
                 error!("Failed to close notification id {}: {}", id, e);
             }
@@ -197,6 +205,10 @@ impl Notifications {
         self.notifications.iter().find(|n| n.id == id)
     }
 
+    fn revision_of(&self, id: u32) -> Option<u64> {
+        self.find_notification(id).map(|n| n.revision)
+    }
+
     fn find_first_action_key(&self, id: u32) -> Option<String> {
         self.find_notification(id)
             .filter(|n| !n.actions.is_empty())
@@ -276,9 +288,7 @@ impl Notifications {
                 }
 
                 // A replace cancels an in-flight dismissal so the replacement is
-                // shown instead of being swallowed by the slide-out. A dismissal
-                // already forwarded over D-Bus can still race its own trailing
-                // `Closed` past here and take the replacement down with it.
+                // shown instead of being swallowed by the slide-out.
                 self.dismiss_phases.remove(&notification.id);
 
                 // Critical notifications are persistent per the freedesktop
@@ -368,7 +378,8 @@ impl Notifications {
             Message::NotificationClicked(id) => {
                 let connection = self.connection.clone();
                 let action_key = self.find_first_action_key(id);
-                Action::Task(invoke_and_close_task(connection, id, action_key))
+                let revision = self.revision_of(id);
+                Action::Task(invoke_and_close_task(connection, id, action_key, revision))
             }
             Message::NotificationClosed => Action::None,
             Message::ClearNotifications => {
@@ -441,17 +452,19 @@ impl Notifications {
             }
             Message::CloseNotificationById(id) => {
                 let connection = self.connection.clone();
+                let revision = self.revision_of(id);
                 let had_toasts = self.remove_toast(id);
                 self.dismiss_phases.remove(&id);
 
-                let task = close_notification_by_id_task(connection, id);
+                let task = close_notification_by_id_task(connection, id, revision);
                 self.hide_toasts_if_empty_with_task(had_toasts, task)
             }
             Message::DismissToast(id) => {
                 if self.toasts.contains(&id) && !self.dismiss_phases.contains_key(&id) {
                     let connection = self.connection.clone();
                     let action_key = self.find_first_action_key(id);
-                    let invoke_task = invoke_and_close_task(connection, id, action_key);
+                    let revision = self.revision_of(id);
+                    let invoke_task = invoke_and_close_task(connection, id, action_key, revision);
                     if !self.animations_enabled {
                         let had_toasts = self.remove_toast(id);
                         let hide_action =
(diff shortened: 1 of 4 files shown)


src/modules/notifications.rs
MalpenZibo, line 396: Is not possible to maintain the `tokio::time::sleep` and add a check here instead of using the tick every 250ms?

Something like 
```
let Some(&deadline) = self.toast_timers.get(&id) else {
    return Action::None;              
};
let now = Instant::now();
if now < deadline {                   
    return Action::Task(delayed_toast_message(deadline - now, id, Message::ExpireToast));
}
self.toast_timers.remove(&id);
```

JockeM:
Ahh I didn't check for prs when creating #924

~~Would it make sense to use a single one-shot subscription for the earliest toast deadline instead of polling every 250 ms?~~

romanstingler:
@MalpenZibo yeah, I think what you sketched there is exactly right for expiry: keep `sleep`, and on a stale wakeup either reschedule for the remaining time or actually expire, no tick needed for that part. A replace then just updates the stored deadline; whichever sleep is already in flight will catch up on its next wakeup instead of spawning a new one.

Where it doesn't fully cover things is the two follow-ups:

#627 (progress bar): once there's a shrinking timeout bar, the view needs `remaining/total` on every repaint, not just at the deadline. That data doesn't exist anywhere until the sleep resolves, so I'd mirror `(deadline, total)` into the struct for the view to read, same as before. But something also has to trigger repaints while the bar is shrinking, since nothing else in state changes between now and expiry. Even with animations off (not everyone enables them), that means a small ~50ms tick purely to drive redraws. It wouldn't decide expiry, your deadline check still does that.

#626 (hover pause): pausing is just "don't let the deadline win while hovered," then push it out by the paused duration on unhover, no new mechanism, same check as your snippet. The part I'm not sure about is the question I left on that issue: when several toasts are visible and only one is hovered, do the others keep counting down, or does hovering any one of them pause all?

Should we keep your deadline check/reschedule for actual expiry, no polling there. Add a separate, small tick only while a toast with the bar visible is on screen, solely for view repaints. Want the bar to update continuously like that(see https://github.com/MalpenZibo/ashell/issues/626#issuecomment-5271782623) , or would you rather it just jump at fixed intervals so there's no tick at all?

MalpenZibo:
Both bugs are real. Let's go with the deadline check and drop the tick. Three things to notice.

1. A new sleep on every deadline, not just the first one. Otherwise critical → normal replace stores a deadline with no sleep in flight to catch it, and the toast never expires:

```rust
let task = if let Some(timeout) = timeout {
    self.toast_timers
        .insert(notification.id, Instant::now() + timeout);
    delayed_toast_message(timeout, notification.id, Message::ExpireToast)
} else {
    self.toast_timers.remove(&notification.id);
    Task::none()
};

Action::Show(task)
```
  Extra sleeps are harmless: the first one past the deadline expires and removes the entry, the others hit the `else { return Action::None }.`

2. A replace landing while the toast is in `dismiss_phases` still gets dropped. A replace should cancel the dismissal:

``` rust
// in toast_action_for_update_event, Received branch
self.dismiss_phases.remove(&notification.id);
```

which also needs a guard here, since today it removes unconditionally:

```rust
Message::DismissAnimationComplete(id) => {
    if self.dismiss_phases.remove(&id).is_none() {
        return Action::None; // cancelled by a replace
    }
    let had_toasts = self.remove_toast(id);
    self.hide_toasts_if_empty(had_toasts)
}
```

3. #627 we can think about this later, but in theory the animated components repaint themselves from inside the widget, so it's possible that we do not need a tick

On #626: pause all visible toasts while the pointer is over the toast surface, not just the hovered one.

Note: Message::ConfigReloaded does self.toasts.clear() by hand, so it now leaves toast_timers behind as well. clear_toasts() covers all three maps; we should use it.

MalpenZibo:
@romanstingler are we able to fix this before the release? If you don't have time, I can check it

romanstingler:
@MalpenZibo great to hear that the next release is near. when do you want to release it? 
Please take it over, I have security hardening and audit at the company for this and the next week and I have no mental capacity to think complex things through. I think I will need a few days downtime. 

Down here just the leftover and end of progress.

test harness which used `Notify` and grim to determine the outcome of the various scenarios. 
[scratchpad.zip](https://github.com/user-attachments/files/31477481/scratchpad.zip)



**Sleep per deadline.** The check holds because a stored deadline always has a wakeup
pending for it: every wakeup either expires and drops the deadline, or reschedules itself.
Critical to normal was the hole, nothing was pending there.

Added on top: spawn the sleep only if the id had no deadline, or if the new deadline
precedes the pending wakeup. The extra sleeps are harmless for correctness as you say, so
this is cost only, and the cost is small. A stale wakeup re-arms instead of returning, so
each chain survives and keeps waking until the toast expires. One id replaced in a loop,
10s timeout, debug builds, read the ratios:

| | replaces | CPU during the stream | CPU in the 12s after |
|---|---|---|---|
| every 100ms for 30s, unguarded | 224 | 9.41s | 0.62s |
| every 100ms for 30s, guarded | 224 | 8.39s | 0.12s |
| flat out for 20s, unguarded | 370 | 14.66s | 1.39s |
| flat out for 20s, guarded | 372 | 13.76s | 0.08s |

7 to 12% during the stream, where repaint dominates, 5x to 17x on the drain. An
optimisation, not a fix: two lines, keeps it at one chain per id. Drop it if you prefer.

**Replace cancels the dismissal.** `is_none()` bail in `DismissAnimationComplete`.
`StartCollapse` self cancels through its `Some(Sliding)` guard.

Open point on that path: `DismissToast` sets the sliding phase and sends the close in the
same update, and the `Closed` that comes back used to be absorbed by the `dismiss_phases`
guard. Now that a replace clears that entry, a `Received` landing between the click and the
close task pushing its `Closed` leaves that `Closed` unabsorbed, which drops the toast and,
since `apply_update_event` runs either way, the list entry too. One scheduling hop wide, so
it is a code comment for now. Fixing it means suppressing the whole stale event at the
dispatch site, not just the toast half, so it is not a one-liner. 

**clear_toasts.** Used in `ConfigReloaded` and the `toast_limit == 0` path.

### Tests

No test suite, so: ashell as the notification daemon on a Hyprland session, `gdbus` calling
`Notify` directly with explicit `replaces_id`, urgency hint and `expire_timeout`, and the
toast stack measured by diffing `grim` captures against a baseline. Pixel height
distinguishes zero, one and two cards. Builds: `main`, the first version of this PR, and
the current one.

| Case | `main` | first version | current |
|---|---|---|---|
| Replace stacks a second card | **223px, two cards** | 105px | 105px |
| normal(1.5s) then critical replace persists | **dismissed at 1.46s** | held | held |
| critical then normal(1.5s) replace expires | gone at 1.62s | gone at 1.96s | gone at 1.72s |
| Replace with a longer timeout (1.5s to 4s) | **gone at ~1.0s** | gone at ~4.9s | gone at ~4.97s |
| Replace with a shorter timeout (8s to 1s) | gone at 1.22s | gone at 1.47s | gone at 1.18s |
| Replace during the dismiss animation, 5 offsets | **swallowed 5/5** | **swallowed 5/5** | kept 5/5 |
| Toasts off and back on via config reload | ok | ok | ok |

* Both original bugs reproduce on `main`. The duplicate is two identical cards in the
  screenshot, both showing the replacement's text.
* Replace during dismissal was previously unverified: swallowed 5/5 on `main` and on the
  first version, kept 5/5 now.
* The longer timeout row involves no critical urgency, so bug 2 hits normal notifications
  too. Added to the description.
* Config reload behaves identically on all three builds. The stale `dismiss_phases` and
  `toast_timers` entries only matter with a dismissal in flight at reload time, which is
  not reachable from outside.

MalpenZibo:
Ok, I want to merge this one and https://github.com/MalpenZibo/ashell/pull/935 and then proceed with the release.

More or less for the first of September

PR #923, opened 2026-08-13
add [logging] section with level, target and directory in config

### LINKED ISSUE #891: configurable logging location (file or stdout)
Configuration option and/or command-line option whether to log to a file, possibly with configurable log directory location, or stdout.
Logging to stdout is useful for running ashell as a systemd service which already has logging through journald.

currently ashell always only logs to a file:

https://github.com/MalpenZibo/ashell/blob/e1a01416cb928b7e118f21aa72e6502717452c69/src/main.rs#L198-L210

romanstingler:
Right now the only logging knob is a top-level `log_level`, and logs always go to a
file in `$XDG_RUNTIME_DIR/ashell/`. Annoying if you're debugging and just want the
output in the terminal you launched ashell from, or if you want the logs somewhere
that survives a reboot.

This moves logging into its own `[logging]` section and adds two options:

```toml
[logging]
level = "warn"                 # same as the old log_level
target = "file"                # or "stdout" / "stderr"
directory = "~/.local/log"     # only for target = "file", supports ~ and $VAR
```

**Breaking**
`log_level` at the top level is gone, it's `[logging] level` now. Nothing crashes
if you leave the old key in place, it's just ignored. Once #922 lands you'll also
get a warning on `stderr` telling you the key was skipped.

The logger has to be up before the config is parsed, so there's a small
`read_logging_config` that bootstrap-parses just the `[logging]` table before
anything else. It's deliberately silent on errors (it can't log, the logger
doesn't exist yet) and falls back to the old defaults; typos still get caught by
the real `Config` parse a moment later.

Also swapped the hand-written `Config::default()` for a derive while I was in
there, I checked every field, the values are identical.

**Notes**
Changing `[logging]` needs a restart. Hot-reload only re-applies `level`; the destination is fixed at startup.
In debug builds, the duplicate-to-stdout behaviour now only kicks in for `target = "file"`, otherwise you'd see every line twice.
Docs updated in `website/docs/configuration/` (main + full_config).

closes https://github.com/MalpenZibo/ashell/issues/891

### DIFF
--- a/docs/src/contributing/common-tasks.md
+++ b/docs/src/contributing/common-tasks.md
@@ -115,7 +115,8 @@ To see debug output for a specific module:
 
 ```toml
 # In config.toml
-log_level = "warn,ashell::modules::my_module=debug"
+[logging]
+level = "warn,ashell::modules::my_module=debug"
 ```
 
 ## Testing Config Hot-Reload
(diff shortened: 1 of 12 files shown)


MalpenZibo (approved):
I fixed some documentation stuff!

Issue #889, opened 2026-07-28
Add notifications ipc commands clear/invoke


JockeM:
I wanna add support for a couple of ipc commands for notifications, thought I would add a issue as stated before I open a pr.

I looked at makoctl and took some inspiration

- notifications-clear [--id <id>] — clears every notification, or only the specified notification.
- notifications-invoke [--id <id>] [action] — invokes an action, defaulting to default. Without an ID, it targets the first notification offering that action.
- notifications-list — outputs active notifications as JSON, including id, app name, summary, body, urgency, and available actions.

```bash
  ashell msg notifications-clear
  ashell msg notifications-clear --id 42
  ashell msg notifications-invoke --id 42
  ashell msg notifications-invoke --id 42 open
  ashell msg notifications-list
```

### Related
#765

romanstingler:
yes this looks sensible. feel free to add that feature.

jazzpi:
In Dunst, I also frequently use `dunstctl close` (closes the last notification) and `dunstctl history-pop` (restores the last notification). Would be great to have these too.

PR #924, opened 2026-08-15
fix(notifications): handle replacement notifications

JockeM:
Found this when implementing the new ipc commands for notifications #889

Fix notification replacement when `replaces_id` reuses an existing ID:

- Update the existing notification instead of creating duplicates.
- Prevent old timeout and animation tasks from dismissing its replacement.
- Track each toast lifetime with a generation-aware key.

Ran some tests using this script (uses some new commands based on #889)
https://gist.github.com/JockeM/c2daf044408cdfef7fa87d6787fc523c

### DIFF
--- a/src/modules/notifications.rs
+++ b/src/modules/notifications.rs
@@ -90,11 +90,15 @@ fn close_notification_by_id_task(connection: Option<Connection>, id: u32) -> Tas
     )
 }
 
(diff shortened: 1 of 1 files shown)


JockeM:
As mentioned on #920 didn't se that pr when doing this. Feel free to close this if that one solves it better.

PR #927, opened 2026-08-16
feat(memory): amount option, showing XX.X GiB

Aadniz:
This PR introduces a new option for how to display the memory. I called it "Amount", which will just show the amount of GiB used in XX.X GiB format.

I chose 1 decimal for aesthetics. I know 2 decimals are also often preferred, but it will then take up more space.

It looks like this:

image

I was considering also setting the units based on how much is used (so if 0.5 GiB is used, it should show 500 MiB instead), but I think that's too much of a stretch.

Let me know what you think, let me know if you want anything changed.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -318,6 +318,7 @@ pub enum DiskFormat {
 pub enum MemoryFormat {
     #[default]
     Percentage,
+    Amount,
     Fraction,
 }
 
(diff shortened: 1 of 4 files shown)


romanstingler (changes requested):
missing docs 
`website/docs/configuration/modules/system_info.md:47-48 `
Same for the Memory Swap section note (line 57), it should mention that `Amount` is also shared.

as well as in 
`full_config.md` 

just my opinion, for me both indicator and system info menu could have just 1 decimal looks better IMO.
lets wait on that one for @MalpenZibo 's opinion

Aadniz:
Added docs now

PR #933, opened 2026-08-22
fix(network): handle iwd station lifecycle after rfkill unblock

Scott-Nx:
## Summary

Fix IWD Wi-Fi state recovery when AShell is restarted while Wi-Fi is disabled through RFKill.

AShell itself disables IWD Wi-Fi by setting the adapter power state, which correctly results in a kernel RFKill soft block. The RFKill behavior is expected and should remain unchanged.

The bug was in how AShell recovered from that valid soft-blocked state after restarting.

Previously, the IWD backend treated a powered-off Wi-Fi device as if no Wi-Fi device was present. After restarting AShell with Wi-Fi soft-blocked, this caused the Wi-Fi toggle to disappear.

Keeping the device visible exposed a second issue: IWD creates and removes its `Station` interface as the radio state changes. If AShell starts while Wi-Fi is blocked, there is no `Station` interface to subscribe to, so enabling Wi-Fi later leaves station-specific listeners missing and the network UI stale.

## Changes

* Separate Wi-Fi device presence from whether the radio is powered on.
* Keep existing IWD/RFKill behavior when AShell disables Wi-Fi.
* Watch IWD `Station` interface additions and removals through the D-Bus object manager.
* Rebuild station-dependent listeners when IWD station topology changes.
* Refresh network state after rebuilding listeners.
* Clean up registered signal level agents before rebuilding.
* Use unique signal agent object paths across listener generations.
* Use actual device presence instead of powered state when updating Wi-Fi presence after scans.

## Before

```text
AShell disables Wi-Fi
→ IWD adapter Powered=false
→ kernel RFKill soft block (expected)
→ restart AShell
→ Wi-Fi device considered absent
→ Wi-Fi toggle unavailable
```

After correcting device presence alone:

```text
restart while blocked
→ toggle remains available
→ enable Wi-Fi
→ RFKill unblocks
→ IWD creates Station
→ old AShell subscription has no Station listeners
→ scan / connection / signal state remains stale
```

## After

```text
AShell disables Wi-Fi
→ RFKill soft block remains expected behavior
→ restart while Wi-Fi is blocked
→ Wi-Fi remains present but disabled
→ enable Wi-Fi
→ RFKill unblocks
→ IWD creates Station
→ AShell detects Station topology change
→ station listeners are rebuilt
→ network state is refreshed
→ scanning, connection state and signal updates resume
```

## Testing

Manually tested the reported flow:

* Disable Wi-Fi through AShell/IWD.
* Confirm this results in the expected RFKill soft block.
* Restart AShell while Wi-Fi remains soft-blocked.
* Confirm Wi-Fi toggle remains visible and disabled.
* Enable Wi-Fi from AShell.
* Confirm RFKill is unblocked.
* Confirm nearby Wi-Fi networks populate again.
* Confirm connection state and Wi-Fi indicator recover without restarting AShell.

### DIFF
--- a/src/services/network/iwd_dbus/mod.rs
+++ b/src/services/network/iwd_dbus/mod.rs
@@ -30,8 +30,9 @@ use iced::futures::stream::select_all;
 use iced::futures::{Stream, StreamExt};
 
 use log::{debug, info, warn};
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::ops::Deref;
+use std::sync::atomic::{AtomicU64, Ordering};
 use tokio::process::Command;
 use zbus::fdo::ObjectManagerProxy;
 use zbus::zvariant::OwnedObjectPath;
@@ -348,6 +349,8 @@ impl From<String> for IwdStationState {
     }
 }
 
+static NEXT_SIGNAL_AGENT_ID: AtomicU64 = AtomicU64::new(0);
+
 struct SignalAgent {
     tx: tokio::sync::mpsc::UnboundedSender<(OwnedObjectPath, u8)>,
 }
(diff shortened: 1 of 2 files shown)


src/services/network/iwd_dbus/mod.rs
```diff
@@ -754,10 +803,8 @@ impl IwdDbus<'_> {
             // TODO: add a future that waits for 10s to poll certain information like the signal
             // strength change
         ])
```
MalpenZibo, line 806: maybe we should add an `stream::pending().boxed(),` to the events. Probably, in cases where all the vectors are empty, we end up in a loop.

Scott-Nx: I agree. Fixed by adding `stream::pending()`.

MalpenZibo (approved):
👍

PR #934, opened 2026-08-23
Use notification hint image data for icon

r3dArch:
This PR adds support for the image-data hint of Freedesktop notifications while still falling back to the existing lookup system. It looks a bit funny since the application name is still Firefox, so this PR may still need a couple changes.

## Before
ashell_before

## After
ashell_after

### DIFF
--- a/src/services/notifications/mod.rs
+++ b/src/services/notifications/mod.rs
@@ -29,9 +29,11 @@ impl NotificationIcon {
         app_icon: &str,
         hints: &HashMap<String, OwnedValue>,
     ) -> Option<Self> {
-        try_icon_from_hints(hints)..or_else(|| icon_candidates(app_name, app_icon, hints)
-            .find_map(resolve_candidate)
-            .map(Self::from_path))
+        try_icon_from_hints(hints)..or_else(|| {
+            icon_candidates(app_name, app_icon, hints)
+                .find_map(resolve_candidate)
+                .map(Self::from_path)
+        })
     }
 
     fn from_path(path: PathBuf) -> Self {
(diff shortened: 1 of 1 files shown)


MalpenZibo:
As it is, it seems that this PR only introduces a regression :smile: 

Could you add more details about the issue and what this PR resolve?

r3dArch:
I should have given more context, sorry. Web apps like Proton Mail (the icon in the screenshot) often provide their own icon for the user to distinguish notifications more easily. The notification is technically from Firefox, but it is just passing along a notification from the Proton Mail site. This is the same behavior as my previous notification daemon: Mako.

MalpenZibo:
ok so in theory, in this case that Proton Mail icon is more correct than the generic Firefox one. Nice!

src/services/notifications/mod.rs
```diff
@@ -29,9 +29,9 @@ impl NotificationIcon {
         app_icon: &str,
         hints: &HashMap<String, OwnedValue>,
     ) -> Option<Self> {
-        icon_candidates(app_name, app_icon, hints)
+        try_icon_from_hints(hints).or(icon_candidates(app_name, app_icon, hints)
```
MalpenZibo:
```suggestion
        try_icon_from_hints(hints)..or_else(|| icon_candidates(app_name, app_icon, hints)
```

I think that `or` evaluates its argument eagerly; `or_else` in this case, is better

MalpenZibo (changes requested):
Just a little change to improve the performance!

r3dArch:
Thanks for the tip on `or_else`. Are there any other changes you'd like made?

MalpenZibo (approved):

MalpenZibo:
Perfect! Let's merge

PR #937, opened 2026-08-25
clamp tempo weather coordinates to valid ranges

romanstingler:
Add validation for tempo module weather location coordinates, 
clamping latitude to `-90..=90` and longitude to `-180..=180` when invalid values are provided.
Logs a warning when clamping occurs.


MalpenZibo (approved):
👍

PR #939, opened 2026-08-25
document shell execution of command options

romanstingler:
Add a new section explaining how command-string configuration options are executed through `bash -c`, 
listing all affected options across modules (updates, settings, power menu, custom modules) and including 
a security warning about running untrusted commands with user privileges.


MalpenZibo (approved):
👍

PR #941, opened 2026-08-29
 perf: reduce unnecessary service and icon cache overhead

Scott-Nx:
This PR contains three small, independent resource-usage cleanups:

**Brightness**
- Listen for `READABLE` readiness on the udev monitor socket instead of `WRITABLE`.
- The socket is consumed as a read-side event source, so this matches the actual I/O direction.

**Privacy**
- Track PipeWire globals that were added as privacy streams.
- Only emit `PrivacyEvent::RemoveNode` for IDs that were previously tracked as `Stream/Input/Audio` or `Stream/Input/Video`.

**XDG icons**
- Stop retaining `SYSTEM_ICON_NAMES` for the lifetime of the process.
- Build `SYSTEM_ICON_ENTRIES` from a temporary `BTreeSet` and release it after the index is constructed.
- Store a separate normalized icon name only when normalization changes the original value, avoiding duplicate string storage for already-normalized names.

## Why

These changes remove unnecessary work and retained data without changing intended user-visible behavior.

For PipeWire, `global_remove` is emitted for all kinds of graph objects. Previously every removal was forwarded through the privacy service and iced update loop, even when the object had never been tracked as a privacy stream.

For XDG icons, the system icon-name set was only needed while building the search entries but remained resident afterward. The entry representation also no longer stores a duplicate normalized string when the original name is already normalized.

The brightness change is mainly a correctness/clarity fix. The existing `clear_ready()` already prevented a busy loop, but the service was waiting for write readiness while consuming read events.

## Measurements

Tested on a Ryzen 7 7735HS / Hyprland / Wayland system using a release build.

### Privacy

```text
global_remove callbacks:        6
RemoveNode events before:       6
RemoveNode events after:        0
````

This removes 100% of unrelated privacy updates for that workload.

### XDG icons

System icon scan found:

```text
unique icon names:              9,689
retained SYSTEM_ICON_NAMES:     ~766 KiB
estimated net retained saving:  ~378–766 KiB
```

Exact RSS change depends on allocator page retention. The additional allocation reduction from the compact icon-entry representation was not separately measured.

### Brightness

A udev readiness probe showed:

```text
current path:  1 initial WRITABLE wakeup
patched path:  0 READABLE wakeups
```

No measurable idle CPU difference was observed. This change makes readiness semantics match actual socket usage.

### Validation

* `cargo fmt`
* `cargo check`
* `cargo clippy --all-features -- -D warnings`

No intended user-visible behavior changes.

### DIFF
--- a/src/services/xdg_icons.rs
+++ b/src/services/xdg_icons.rs
@@ -17,21 +17,34 @@ use std::sync::RwLock;
 
 const MAX_SIMILAR_ICON_CANDIDATES: usize = 5;
 
+struct IconEntry {
+    name: Box<str>,
+    normalized: Option<Box<str>>,
+}
+
+impl IconEntry {
+    fn normalized(&self) -> &str {
+        self.normalized.as_deref().unwrap_or(&self.name)
+    }
+}
+
 static ICON_CACHE: LazyLock<RwLock<HashMap<String, Option<XdgIcon>>>> =
     LazyLock::new(|| RwLock::new(HashMap::new()));
-static SYSTEM_ICON_ENTRIES: LazyLock<Vec<(Cow<'static, str>, Cow<'static, str>)>> =
-    LazyLock::new(|| {
-        let names = load_system_icon_names();
-        names
-            .iter()
-            .filter_map(|name| {
-                let name_str = name.to_str()?;
-                let normalized = normalize_icon_name(name_str);
-                let normalized_cow = Cow::Owned(normalized.into_owned());
-                Some((Cow::Owned(name_str.to_string()), normalized_cow))
+static SYSTEM_ICON_ENTRIES: LazyLock<Vec<IconEntry>> = LazyLock::new(|| {
+    load_system_icon_names()
+        .iter()
+        .filter_map(|name| {
+            let name_str = name.to_str()?;
+            Some(IconEntry {
+                name: name_str.into(),
+                normalized: match normalize_icon_name(name_str) {
+                    Cow::Borrowed(_) => None,
+                    Cow::Owned(normalized) => Some(normalized.into_boxed_str()),
+                },
             })
-            .collect()
-    });
+        })
+        .collect()
+});
 static DESKTOP_ICON_INDEX: LazyLock<HashMap<String, String>> =
     LazyLock::new(build_desktop_icon_index);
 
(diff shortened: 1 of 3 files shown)


src/services/xdg_icons.rs
MalpenZibo, line 18:
```suggestion
const MAX_SIMILAR_ICON_CANDIDATES: usize = 5;

struct IconEntry {
    name: Box<str>,
    normalized: Option<Box<str>>,
}

impl IconEntry {
    fn normalized(&self) -> &str {
        self.normalized.as_deref().unwrap_or(&self.name)
    }
}
```

src/services/xdg_icons.rs
```diff
@@ -19,19 +19,15 @@ const MAX_SIMILAR_ICON_CANDIDATES: usize = 5;
 
 static ICON_CACHE: LazyLock<RwLock<HashMap<String, Option<XdgIcon>>>> =
     LazyLock::new(|| RwLock::new(HashMap::new()));
-static SYSTEM_ICON_NAMES: LazyLock<BTreeSet<OsString>> = LazyLock::new(load_system_icon_names);
 static SYSTEM_ICON_ENTRIES: LazyLock<Vec<(Cow<'static, str>, Cow<'static, str>)>> =
```
MalpenZibo:
```suggestion
static SYSTEM_ICON_ENTRIES: LazyLock<Vec<IconEntry>> = LazyLock::new(|| {
    load_system_icon_names()
        .iter()
        .filter_map(|name| {
            let name_str = name.to_str()?;
            Some(IconEntry {
                name: name_str.into(),
                normalized: match normalize_icon_name(name_str) {
                    Cow::Borrowed(_) => None,
                    Cow::Owned(normalized) => Some(normalized.into_boxed_str()),
                },
            })
        })
        .collect()
});
```

src/services/xdg_icons.rs
MalpenZibo:
```suggestion
for candidate in SYSTEM_ICON_ENTRIES.iter() {
        let candidate_normalized = candidate.normalized();
        if candidate_normalized == normalized.as_ref() {
```

MalpenZibo:
```suggestion
if candidate_normalized.contains(normalized.as_ref())
            || normalized.as_ref().contains(candidate_normalized)
            || candidate_normalized.contains(normalized_no_separators.as_ref())
{
```

MalpenZibo:
```suggestion
matches.push(Cow::Borrowed(&candidate.name));
```

MalpenZibo:
```suggestion
.find(|entry| entry.normalized() == normalized.as_ref())
```

MalpenZibo:
```suggestion
return Some(Cow::Borrowed(&exact.name));
```

MalpenZibo:
```suggestion
candidates.retain(|entry| entry.normalized().chars().nth(idx) == Some(ch));
```

MalpenZibo:
```suggestion
return Some(Cow::Borrowed(&candidates[0].name));
```

MalpenZibo:
```suggestion
candidates.first().map(|entry| Cow::Borrowed(&*entry.name))
```

MalpenZibo (commented):
Nice!! This change should reduce also the number of allocations

MalpenZibo:
check the CI

MalpenZibo (approved):
Top!!

PR #961, opened 2026-09-10
fix(menu): release bar keyboard focus when a menu closes

### LINKED ISSUE #960: Ashell steals focus by mouse over in hyprland
## Setup
ashell: 0.10.0 (95cb69e)

<details>
<summary>hyprland: 0.56.0</summary>
Hyprland 0.56.0 built from branch main at commit 7ebf13abb3c391604c60c9f627c7a403bcec8d17 clean (screencopy/toplevel: fix top level capture ( 16132)).
Date: Mon Sep 7 13:38:00 2026
Tag: v0.56.0-177-g7ebf13abb, commits: 7802

Libraries:
Hyprgraphics: built against 0.5.1, system has 0.5.1
Hyprutils: built against 0.14.2, system has 0.14.2
Hyprcursor: built against 0.1.13, system has 0.1.13
Hyprlang: built against 0.6.8, system has 0.6.8
Aquamarine: built against 0.15.0, system has 0.15.0

Version ABI string: 7ebf13abb3c391604c60c9f627c7a403bcec8d17_aq_0.15_hu_0.14_hg_0.5_hc_0.1_hlg_0.6
no flags were set
</details>

## Problem

If the mouse goes over the ashell bar, the window on the same workspace cannot be focused anymore (you can interact with the window, but keyboard focus is lost). A switch to another workspace brings the focus back to the window.
- i tried different options of hyprland follow_mouse - no changes
- i think this is related https://github.com/hyprwm/Hyprland/discussions/13116 - the fix was on the app side, not hyprland. 

P.S.: thanks for the great status bar!

romanstingler:
When `enable_esc_key = true` the bar has to listen for Escape while a
menu is open, so opening a menu switches it to `OnDemand` keyboard
interactivity. It never got switched back. Hyprland hands keyboard focus
to an `OnDemand` layer surface as soon as the pointer enters it, so once
you had opened and closed a single menu, hovering the bar took the
keyboard away from whatever window you were typing in, and only a
workspace switch gave it back.

The release used to work, before the menu close animation landed in
904cd400. Back then `close()` dropped the menu straight away. Now it
starts an animation and only really closes in `finish_close()`, so the
menu still counts as open when we check whether anything still needs the
keyboard, and the release gets skipped. Clicking a button to toggle a
menu shut went through the same check and just re-armed `OnDemand`.

Use a new `menu_needs_keyboard()` for that check instead, which ignores
menus that are already animating closed. Also release every output's bar
rather than just the one being toggled: opening a menu only arms its own
bar, so on a multi-monitor setup a bar could stay armed after its menu
had closed somewhere else.

closes #960


MalpenZibo (approved):
🚀

